Issue #11, opened 2016-09-14
Download requirement fails (with http error 503)


pcmind:
I am using VSC behind a proxy and plugin is not able to download dependencies.

It fails because [Server Url](http://download.jboss.org/jbosstools/static/vscode/java-server-1.0.0-201609122022.tar.gz) is a http, and downloadManager use https agent without setting the `options.port ` to the correct port. 
[agent-base](https://github.com/TooTallNate/node-agent-base/blob/12a9a851006bd039a0340fa876916266847a6044/patch-core.js#L50) set port to 443, when `options.port ` is omitted. Request fail because resource is unavailable at this new location.

gorkem:
just pushed 0.0.2 to marketplace. It now uses https download urls

PR #12, opened 2016-09-14
Set missing port option when request url is http

pcmind:
Solution to solve issue #11


gorkem:
I think this will break the non-proxy users, or non-https proxies. The main issue is we are using https module to download an http url. https module always uses a TLS socket regardless of the port option, which fails if it is not a TLS connection.

pcmind:
You are right. 
The other solution that I had in mind was to switch to http when requesting http url like:
`if(url.protocol === 'http:'){
			return http.get(options, callback);
		}else{
			return https.get(options, callback);
		}`
This solution work behind proxy and no proxy.

PR #44, opened 2016-09-26
Fix #28 add option to set the JDK home directory, from which to run the JLS

### LINKED ISSUE #28: Option to specify Java path ?
The extension needs Java 8 and I'm working on Java 7 projects so I cannot updated to Java 8 easily.

How about an option to specify Java directory (JAVA_HOME already points to Java 7) ?

fbricon:

### DIFF
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -1,13 +1,14 @@
 'use strict';
 
-import { Uri } from 'vscode';
+import { workspace, Uri } from 'vscode';
 import * as cp from 'child_process';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as downloadManager from './downloadManager';
 
-
-var findJavaHome = require('find-java-home');
+const pathExists = require('path-exists');
+const expandHomeDir = require('expand-home-dir')
+const findJavaHome = require('find-java-home');
 
 interface RequirementsData {
     java_home: string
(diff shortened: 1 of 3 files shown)


src/requirements.ts
```diff
@@ -35,13 +35,25 @@ export async function resolveRequirements(): Promise<RequirementsData> {
...
+        const config = workspace.getConfiguration();
+        const javaHome = config.get<string>('java.home');
+		if (typeof javaHome === 'string') {
+            //Need to inject JAVA_HOME value before calling findJavaHome 
+            process.env["JAVA_HOME"] = javaHome;
+        }
+        try {
+            findJavaHome(function (err, home) {
+                if (err)
+                    openJDKDownload(reject, "The JAVA_HOME environment variable could not be detected, or it doesn't point to a JDK.");
```
fbricon: Agreed, that's something I meant to do next

fbricon:
java.home > JAVA_HOME, so I don't think we should check whether JAVA_HOME is invalid.

If java.home is not valid but JAVA_HOME is, then the error message about invalid java.home should be enough (see previous comment)

gorkem:
It is the other way around that I am worried about because we would be starting the server after restoring the JAVA_HOME to the original invalid value which would never be checked for validity.

fbricon:
So you want to keep the java.home value in JAVA_HOME env if it's valid?
I guess we could do that.

gorkem:
Actually, I was suggesting that both java.home configuration and JAVA_HOME value (if one exists) should be verified and reported for errors, separately.

fbricon:
@gorkem I've added a warning asking to restart vscode when java.home changed. 

- I'd rather access the Action api https://github.com/Microsoft/vscode/blob/eb4b50649821d61df0e6d6823b20c94927ab2996/src/vs/workbench/electron-browser/actions.ts#L392 but I haven't figured out how to use it from our extension
- In the future, I expect restarting the whole window won't be necessary, just the JLS. But the current approach is the easiest/fastest for now

src/requirements.ts
```diff
@@ -35,13 +36,43 @@ export async function resolveRequirements(): Promise<RequirementsData> {
...
     return new Promise((resolve, reject) => {
-        findJavaHome(function (err, home) {
-            if (err)
-                openJDKDownload(reject, "The JAVA_HOME environment variable could not be detected, or it doesn't point to a JDK.");
-            else
-                resolve(home);
+        const realJavaHome = process.env["JAVA_HOME"]
+        const config = workspace.getConfiguration();
+        let javaHome = config.get<string>('java.home');
+        let source = "The JAVA_HOME environment variable";
```
gorkem: Although we do not have translations, it will be difficult to have them in the future for messages that are passed in parts.

fbricon: I don't think it'll be that difficult. But I'll worry about it when there's a need for i18n

PR #50, opened 2016-10-04
Fix #25 add support for user-defined VM args

### LINKED ISSUE #25: Expose server JVM launch parameters
We should make it possible for users to override the default JVM launch parameters for the server.
Typically, users should be able to increase the amount of memory  of the JVM or launch the server with some java agents (eg. lombok, as requested on https://marketplace.visualstudio.com/items?itemName=redhat.java)

fbricon:

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,11 @@
 					"type": ["string", "null"],
 					"default": null,
 					"description": "Specifies the folder path to the JDK 8 used to launch the Java Language Server."
+				},
+				"java.jdt.ls.vmargs": {
+					"type": ["string", "null"],
+					"default": "-noverify -Xmx1G",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -39,6 +39,11 @@
 					"type": ["string", "null"],
 					"default": null,
 					"description": "Specifies the folder path to the JDK 8 used to launch the Java Language Server."
+				},
+				"java.jls.vmargs": {
```
gorkem: Should we use java.jdt.ls.vmargs since the upstream project is now renamed to jdt.ls

fbricon: sure we can do that.

I've also been pondering whether we should also rename java.home too. But I think it's not necessary as it's more _universal_

gorkem: agree on java.home, it is more universal and basically overwrites the JAVA_HOME value for the server.

fbricon: Changed to java.jdt.ls.vmargs. Also updated readme and changelog.

PR #53, opened 2016-10-05
Fix #49 show Starting Java Language Server on extension startup

### LINKED ISSUE #49: Need to add a visual clue when starting the JLS
when the extension kicks in, the JLS provide feedback only starting from when the JVM has started.
There's a few seconds during which we have 0 indication something is happening, which can be pretty confusing. The tray should display "Starting Java Language Server..." (or better), before the Initialization notifications kick in.

fbricon:
also fixes the case where status is not updated when editor switched to non-java file
before initialization ends.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -91,23 +91,16 @@ export function activate(context: ExtensionContext) {
 	let languageClient = new LanguageClient('java','Language Support for Java', serverOptions, clientOptions);
 	languageClient.onNotification(StatusNotification.type, (report) => {
 		console.log(report.message);
-		if(window.activeTextEditor && window.activeTextEditor.document){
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -133,6 +122,8 @@ export function activate(context: ExtensionContext) {
 	};
 	workspace.registerTextDocumentContentProvider("jdt", provider)
 
+	item.text = "Starting Java Language Server...";
+	toggleItem(window.activeTextEditor, item);
```
fbricon: @gorkem wondering if this is safe to do outside a promise

gorkem: Should be.. I doubt that it is a long running/blocking call.

PR #88, opened 2016-11-02
move jdt.ls.vmargs before "-jar" in order to work

jpospychala:
hi,
jdt.ls.vmargs preference doesn't work now. I think it's because it is added  added AFTER "-jar" argument, so extra args land as JAR arguments, instead of JVM arguments.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -44,6 +44,9 @@ function runJavaServer(){
 				params.push('-Dlog.level=ALL');
 			}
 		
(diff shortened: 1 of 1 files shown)


jpospychala:
meh, not sure how this submodule got here, anyway now it's removed.

fbricon:
Thanks for the fix @jpospychala , the JLS starts much faster when  `-noverify` is added to `jdt.ls.vmargs` now :-)

jpospychala:
you welcome :-)

PR #111, opened 2016-12-05
initial API description [WIP]

gorkem:
An initial description of the java API for getting class paths. Related to gorkem/java-language-server/issues/48 .  and #5 @fbricon

### DIFF
--- a/src/java.d.ts
+++ b/src/java.d.ts
@@ -1,0 +1,58 @@
+import { Uri } from 'vscode';
+
+declare module 'java' {
+
+	/**
+	 * Represents a location that defines libraries and
+	 * source code that can be build as a unit.
+	 * Usually a project.
+	 */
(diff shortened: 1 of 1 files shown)


src/java.d.ts
```diff
@@ -0,0 +1,58 @@
...
+	 */
+   export interface ClassPathRoot{
+		/**
+		 * The uri of this ClasspathRoot
+		 *
+		 * @readonly
+		 */
+		uri : Uri;
+		/**
+		 * The list if classpath entries associated with this root.
```
fbricon: list of

fbricon:
I'm wondering if this should expose scopes. (compile, test, provided...). It should be fairly straightfoward to get the info from m2e. For gradle, I have no idea, because scopes are not standard.

PR #123, opened 2016-12-20
Enable Java configuration auto-update. Fixes #122

### LINKED ISSUE #122: Allow users to enable automatic project configuration
When modifying a pom.xml or a gradle file, java project configuration and classpath should be updated. While it's possible do trigger a manual update (ctrl+alt+u or cmd+alt+u), it'd be more convenient to do it automatically, as requested by @pmuir.

The proposed change is to display a warning when a build file has been modified, asking the user to update the java configuration. Choices are
- Never: no project configuration will be triggered and the message won't show up again 
- Now:  project configuration will be triggered and the message will show up next time
- Always: project configuration will be triggered from now on and  and the message won't show up again

fbricon:

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,17 +61,27 @@
           ],
           "default": "warning",
           "description": "Specifies the severity of the message when the classpath is incomplete for a Java file"
+        },
+        "java.configuration.updateBuildConfiguration": {
+          "type": ["string"],
+          "enum": [
+            "disabled",
+            "interactive",
+            "automatic"
+          ],
+          "default": "interactive",
+          "description": "Specifies how modifications on build files update the Java configuration"
         }
 			}
 		},
 		"commands": [{
-				"command": "java.project.configuration.update",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -61,6 +61,16 @@
...
+        },
+        "java.configuration.updateBuildConfiguration": {
+          "type": ["string"],
+          "enum": [
+            "disabled",
+            "interactive",
+            "automatic"
+          ],
+          "default": "interactive",
+          "description": "Specifies if modifications on build files should update the project configuration"
```
gorkem: Change description: _Specifies how modifications on build files update the Java configuration_

fbricon: Fixed

fbricon:
`java.project.configuration.update` was actually renamed `java.projectConfiguration.update` for consistency reasons.

And yes, it's still useful, in case automatic project is disabled, then you still want to be able to update it manually

PR #141, opened 2017-01-17
a command for applying edits

gorkem:
A client side command for applying the given TextEdits when invoked.


fbricon:
Is there any way to have a shortcut that removes all unused imports?

fbricon:
Other than that, this new changes work as expected. 
+1

PR #172, opened 2017-03-28
Add moar logging when running the jdt.ls

fbricon:


fbricon:
See instructions at https://github.com/redhat-developer/vscode-java/wiki/Troubleshooting#get-the-vscode-java-logs

PR #174, opened 2017-03-30
Support starting the Java Language Server with JDK 9. Fixes #43

### LINKED ISSUE #43: Support running the JLS with Java 9
We check the java version used to run the JLS is 1.8. We should ensure 1.9 works too.

fbricon:
A bit of clarification  : this PR provides a way to start the language server with JDK 9, but doesn't provide support for building Java 9 projects.

### DIFF
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -14,6 +14,7 @@ const JAVAC_FILENAME = 'javac' + (isWindows?'.exe':'');
 
 interface RequirementsData {
     java_home: string;
+    java_version: number;
 }
 
 interface ErrorData {
@@ -31,9 +32,9 @@ interface ErrorData {
  */
 export async function resolveRequirements(): Promise<RequirementsData> {
     let java_home = await checkJavaRuntime();
-    await checkJavaVersion(java_home);
+    let javaVersion = await checkJavaVersion(java_home);
     await checkServerInstalled();
-    return Promise.resolve({ 'java_home': java_home });
+    return Promise.resolve({ 'java_home': java_home, 'java_version': javaVersion});
 }
 
 function checkJavaRuntime(): Promise<any> {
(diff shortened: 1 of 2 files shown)


src/requirements.ts
```diff
@@ -81,11 +82,13 @@ function readJavaConfig() : string {
...
         cp.execFile(java_home + '/bin/java', ['-version'], {}, (error, stdout, stderr) => {
-            if (stderr.indexOf('1.8') < 0){
+            if (stderr.indexOf('version "9') > -1){
+                resolve(9);
+            } if (stderr.indexOf('1.8') < 0){
                 openJDKDownload(reject, 'Java 8 is required to run. Please download and install a JDK 8.');
             }
             else{
-                resolve(true);
+                resolve(8);
```
gorkem: resolve 1.8 to be consistent?

src/requirements.ts
```diff
@@ -81,11 +82,13 @@ function readJavaConfig() : string {
 function checkJavaVersion(java_home: string): Promise<any> {
     return new Promise((resolve, reject) => {
         cp.execFile(java_home + '/bin/java', ['-version'], {}, (error, stdout, stderr) => {
-            if (stderr.indexOf('1.8') < 0){
+            if (stderr.indexOf('version "9') > -1){
```
gorkem: What is the full string returned by the Java9 vm?

src/requirements.ts
fbricon: I disagree. Oracle is calling Java 8 [Java 8](http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html), so that's more consistent was the bigger versioning scheme

src/requirements.ts
fbricon: java -version
java version "9-ea"
Java(TM) SE Runtime Environment (build 9-ea+162)
Java HotSpot(TM) 64-Bit Server VM (build 9-ea+162, mixed mode)

PR #182, opened 2017-04-11
Include server packages into vsix

### LINKED ISSUE #178: Publish server included packages to marketplace
According to https://github.com/Microsoft/vscode/issues/18146 marketplace no longer has restrictions on extension sized. Change the packaging and build to take advantage of this new feature.

gorkem:
updates to include packages in vsix. Removes the
dependency to vscode profile on jdt.ls. Updates jenkins
pipeline and travis builds. Removes downloadManager
related code. fixes #178


fbricon:
Need to update https://github.com/redhat-developer/vscode-java/blob/master/README.md#troubleshooting, 1st item

CONTRIBUTING.md
```diff
@@ -33,13 +33,16 @@ This assumes that you are starting on the `vscode-java` directory
...
-	$ npm run-script build-server
+	$ npm run build-server
 	```
-You can run faster builds by calling the `build-server-mac`, `build-server-win` or `build-server-linux` script instead, depending on your OS. These commands run Maven in offline mode, so you might need to run `build-server` at least once, to fetch all the dependencies.
+You can run faster server builds during development by calling `./node_modules/.bin/gulp dev_server` script instead, this will build server binaries that are required by your host OS only. You can also use `npm run watch-server` which will build and place them on the extension for Java changes. These commands run Maven in offline mode, so you might need to run `build-server` at least once, to fetch all the dependencies.
 
-This will build and place the binaries under the `server` folder. Alternately you can download
-and unzip a pre-built server. For pre-built server information refer to eclipse.jdt.ls
-project.
+This will build and place the binaries under the `server` folder. Alternately you can download and use the latest snapshot build from [Eclipse &trade; JDT Language Server](https://github.com/eclipse/eclipse.jdt.ls) project with the following
```
fbricon: Maybe add link to http://download.eclipse.org/jdtls/snapshots/?d on [snapshot](http://download.eclipse.org/jdtls/snapshots/?d)

PR #245, opened 2017-06-24
Disable the keybinding of formatting document

### LINKED ISSUE #186: How to disable the keybinding of formatting document
Unfortunately code formatting doesn't work, so I want to turn to _clang-format_ instead.
Is there a setting item for _redhat.java_ to disable keybinding of formatting docment ?
##### Environment
- Operating System: Windows 10 Pro x64
- JDK version: 1.8.121
- Visual Studio Code version: 1.11.2
- Java extension version: 0.1.0

##### Steps To Reproduce
1. Open a java file
2. Press _Ctrl + Shift + P_ and enter _Format Document_

##### Current Result

Nothing happends and _clang-format_ extension doesn't work as well.

##### Expected Result

Keybinding of formatting docment can be disabled and  _clang-format_ extension works.

##### Additional Informations

_redhat.java_ code formatting solution is in need.

snjeza:
Fixes #186

You have to set the following

```
{
"java.format.enabled": false
}
```
in order to disable the Eclipse Java formatting.

Requires https://github.com/eclipse/eclipse.jdt.ls/pull/272

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,11 @@
           "type": "string",
           "default": null,
           "description": "Absolute path to Maven's settings.xml"
+        },
+        "java.format.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable default Java formatter"
         }
       }
     },


package.json
```diff
@@ -110,6 +110,11 @@
           "type": "string",
           "default": null,
           "description": "Absolute path to Maven's settings.xml"
+        },
+        "java.format.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable default JAVA formatter"
```
fbricon: s/JAVA/Java

fbricon:
The changes work as expected except the Formatting menus are still available in the UI. Looking at typescript, you can see the formatting provider is [disposed](https://github.com/Microsoft/vscode/blob/8d6645c2420ae59d985b6dd2e0e9f3c6f1e18818/extensions/typescript/src/typescriptMain.ts#L345) when formatting is disabled, so the contextual menus don't show up when disabled.

But the way the formatter is enabled with Java is via ServerCapabilities set during server [initialization](https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/InitHandler.java#L93). I don't see how we can update the ServerCapabilities outside the init phase. @aeschli any recommendations?

aeschli:
In version 3 of the JSL protocol, you can enable and disable capabilities dynamically. See https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#client_registerCapability.

Here's how the HTML LS does it: https://github.com/Microsoft/vscode/blob/master/extensions/html/server/src/htmlServerMain.ts#L98

The JLS node client then does the registering/deregistering of the providers for you.

fbricon:
@aeschli thanks for the pointer, I forgot about that. How come the server is not telling the client it's unregistering the formatter. I see the formatter is disposed on the server side but how does the client know?

aeschli:
When the server sends a unregisterCapability request to the client, the client will know that formatting is no longer offered by the server and will unregister its provider in VSCode.

fbricon:
@aeschli yes I get the theory, but I just don't see how the html server is calling unregisterCapability

aeschli:
Ah it's sugar coated in the NodeJS language server implementation:
Register gives you a disposable, and disposing of that calls unregister:
https://github.com/Microsoft/vscode-languageserver-node/blob/bb57bdc88fb239c42c53edce73c08aa3f4f3313f/server/src/main.ts#L627

PR #257, opened 2017-07-11
CodeLens for Interfaces

snjeza:
Fixes https://github.com/eclipse/eclipse.jdt.ls/issues/99

I have added the java.implementationsCodeLens.enabled property, false by default.
To enable the codeLens implementors, you have to add the following:
```
"java.implementationsCodeLens.enabled": true
```
The implementations will be shown for interfaces and abstract classes.

Requires https://github.com/eclipse/eclipse.jdt.ls/pull/301

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "activationEvents": [
     "onLanguage:java",
     "onCommand:java.show.references",
+    "onCommand:java.show.implementations",
     "onCommand:java.open.output",
     "onCommand:java.open.serverLog",
     "onCommand:java.projectConfiguration.update",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -106,6 +106,11 @@
           "default": true,
           "description": "Enable/disable the references code lens."
         },
+        "java.implementationsCodeLens.enabled": {
+          "type": "boolean",
+          "default": false,
```
fbricon: wondering whether we should enable it by default.
@gorkem ?

fbricon:
Can you please also update the readme, to describe the new preference key?

Issue #282, opened 2017-08-23
Support 3rd-party bundles 


yaohaizh:
Current LS hasn't provide a way to support 3rd party extension/bundles. 

We have some features can be built as extension bundles to vscode-java
Debugger: https://github.com/redhat-developer/vscode-java/issues/9
Selfhost PDE: https://github.com/eclipse/eclipse.jdt.ls/pull/305

fbricon:
Done

PR #283, opened 2017-08-23
Load plugins to vscode java extension.

yaohaizh:

@fbricon @gorkem @aeschli 

https://github.com/redhat-developer/vscode-java/issues/282


aeschli:
Given that the new 'bundles install' request is only needed a single time (on initialization), maybe it's better to just pass the list of bundles to add to the language on initialize.
That keeps the APIs as simple as possible.

The [LSP InitializeRequest](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialize-request) has a property `initializationOptions` that can be used for that.

yaohaizh:
If we go through  the LSP InitializeRequest , should we still expose the installbundle/uninstallbundle protocol for 3rd-party to load/unload themselves dynamically in time?

aeschli:
> should we still expose the installbundle/uninstallbundle protocol for 3rd-party to load/unload themselves dynamically in time?

I don't think there's a need for it, so I'd suggest to remove it. Its also hard to imagine how uninstalling at runtime will work if that bundle is already in use.

yaohaizh:
Thanks, Martin. Updated the code and this did make the code logic clear and clean.

gorkem:
I think I am comfortable with this solution and implementation looks good. @yaohaizh can you possibly add tests for it, and we can then merge?

yaohaizh:
@gorkem You mean test case on the jdt.ls side or the vscode java side? For the jdt.ls side, i have add more testing for loading extension. If for vscode-java side, i might need refactor current code for testing private methods

gorkem:
@yaohaizh client side, something that massages the `javaExtension` contribution logic.

src/plugin.ts
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+import { Extension } from 'vscode';
+import * as path from 'path';
+
+export function collectionExtensions(extensions: Extension<any>[]): string[] {
```
fbricon: please rename to `collectJavaExtensions`

fbricon:
Apart from the function name I'd like to see renamed, it will also a require documentation on contributing Java extensions, in https://github.com/redhat-developer/vscode-java/wiki, that we can reference from a paragraph in  the README.md.

yaohaizh:
Add the wiki for extension side, yet need add server side wiki but I don't have the permission yet to add eclipse/eclipse.jdt.ls wiki.

test/extension.test.ts
```diff
@@ -62,4 +65,20 @@ suite('Java Language Extension', () => {
 		assert.equal(vmArgs[2], '-Dfoo=Some "crazy" stuff');
 	});
 
+	test('should collection java extensions', function () {
```
akaroml: collection -> collect

PR #296, opened 2017-09-11
Disable reference codelenses by default. Fixes #293

### LINKED ISSUE #293: Disable code lenses by default
Code lenses are more of a distraction at the moment and we should have them off by default.

fbricon:


fbricon:
@gorkem is it really what you want? ;-)

PR #298, opened 2017-09-19
build workspace manually

ansyral:
Signed-off-by: xuzho <xuzho@microsoft.com>

eclipse/eclipse.jdt.ls#326

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,8 +147,8 @@
         "category": "Java"
       },
       {
-        "command": "java.build.workspace",
-        "title": "Build workspace manually",
+        "command": "java.recompile.workspace",
+        "title": "Force Java compilation",
         "category": "Java"
       }
     ],
--- a/package.json
+++ b/package.json
@@ -145,6 +145,11 @@
         "command": "java.open.serverLog",
         "title": "Open Java Language Server log file",
         "category": "Java"
+      },
+      {
+        "command": "java.build.workspace",
+        "title": "Build workspace manually",
+        "category": "Java"
       }
     ],
     "keybindings": [
--- a/package.json
+++ b/package.json
@@ -158,6 +158,12 @@
         "key": "ctrl+alt+u",
         "mac": "cmd+alt+u",
         "when": "editorFocus"
+      },
+      {
+        "command": "java.recompile.workspace",
+        "key": "alt+ctrl+b",
+        "mac": "alt+cmd+b",
+        "when": "editorFocus"
       }
     ]
   },
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -171,26 +171,24 @@ export function activate(context: ExtensionContext) {
 						window.withProgress({ location: ProgressLocation.Window }, p => {
 							return new Promise((resolve, reject) => {
 								p.report({ message: 'Building workspace...' });
-								languageClient.sendRequest(BuildWorkspaceRequest.type).then((s: BuildWorkspaceStatus) => {
-									switch (s) {
-										case BuildWorkspaceStatus.CANCELLED:
-											p.report({ message: 'Build cancelled' });
-											reject(s);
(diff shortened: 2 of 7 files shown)


fbricon:
Can you please : 
- update the list of commands in README.MD
- update the list of commands checked in https://github.com/redhat-developer/vscode-java/blob/master/test/extension.test.ts#L25

ansyral:
@fbricon  resolved

gorkem:
I am not sure about how we are presenting this command to the user. We are calling it `Build Workspace Manually`, however this is not really a build for the user (which is done via maven/gradle through tasks) but rather an internal build for re-calculating diagnostics.  Should we rename the command to something that will be confused with the real project build.

ansyral:
@gorkem I'm not that familiar with eclipse API, but i assume that eclipse uses `IWorkspace.build` to build all the projects in the workspace, of course it would leave out other tasks other than `compile`. Am I correct? Or how would we define `build`? here i just meant to `compile` java files to `.class` files

ansyral:
The travis build error should have been fixed by vscode. see https://github.com/Microsoft/vscode/issues/33998

ansyral:
@fbricon @gorkem could you please help review this pr? Thanks.

src/extension.ts
```diff
@@ -167,6 +167,32 @@ export function activate(context: ExtensionContext) {
...
+												break;
+											case BuildWorkspaceStatus.WITHERROR:										
+												reject(s);
+												break;
+											case BuildWorkspaceStatus.SUCCEED:
+												resolve(s);
+												break;
+										}
+									});
+								}, 1000);
```
fbricon: We don't want to delay the command by 1 sec, but rather keep the message visible after 1sec max after completion

fbricon:
```typescript
window.withProgress({ location: ProgressLocation.Window }, p => {
		p.report({ message: 'Building workspace...' });
		let start = new Date().getTime();
		languageClient.sendRequest(BuildWorkspaceRequest.type).then((s: BuildWorkspaceStatus) => {
			let elapsed = new Date().getTime() - start;
			// set a timeout so user would still see the message when build time is short
			let humanVisibleDelay = elapsed < 1000? 1000:0;
			setTimeout(function(){
				switch (s) {
					case BuildWorkspaceStatus.CANCELLED:
					case BuildWorkspaceStatus.FAILED:
					case BuildWorkspaceStatus.WITHERROR:
						reject(s);
						break;
					case BuildWorkspaceStatus.SUCCEED:
						reject(s);
						break;
				}
			}, humanVisibleDelay);
		});
	});
});
```

package.json
```diff
@@ -145,6 +145,11 @@
         "command": "java.open.serverLog",
         "title": "Open Java Language Server log file",
         "category": "Java"
+      },
+      {
+        "command": "java.build.workspace",
+        "title": "Build workspace manually",
```
fbricon: I think we should rename the command title to `Force Java compilation`, as to avoid user confusion wrt actual Maven/Gradle build

src/extension.ts
ansyral: make sense

src/extension.ts
```diff
@@ -167,6 +167,35 @@ export function activate(context: ExtensionContext) {
 						return languageClient.sendRequest(ExecuteCommandRequest.type, params);
 					});
 
+					commands.registerCommand(Commands.RECOMPILE_WORKSPACE, () => {
+						window.withProgress({ location: ProgressLocation.Window }, p => {
+							return new Promise((resolve, reject) => {
+								p.report({ message: 'ReCompiling workspace...' });
```
fbricon: 'Compiling workspace...'

fbricon:
superseded by #350

Issue #314, opened 2017-10-02
External clean build gets jdt.ls out-of-sync


fbricon:
As reported by @xuaveneg in #306, during the execution of a build task (using maven), The errors showing in the output console are:
```
[Error - 10:27:32] 2 oct. 2017 10:26:41 JavaBuilder handling CoreException
File not found: [filename].
org.eclipse.core.runtime.CoreException: File not found: [filename].
	at org.eclipse.core.internal.filesystem.Policy.error(Policy.java:45)
	at org.eclipse.core.internal.filesystem.local.LocalFile.openInputStream(LocalFile.java:406)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.read(FileSystemResourceManager.java:849)
	at org.eclipse.core.internal.resources.File.getContents(File.java:277)
	at org.eclipse.jdt.internal.core.util.Util.getResourceContentsAsByteArray(Util.java:1135)
	at org.eclipse.jdt.internal.core.builder.IncrementalImageBuilder.writeClassFileCheck(IncrementalImageBuilder.java:889)
	at org.eclipse.jdt.internal.core.builder.IncrementalImageBuilder.writeClassFileContents(IncrementalImageBuilder.java:831)
	at org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.writeClassFile(AbstractImageBuilder.java:860)
	at org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.acceptResult(AbstractImageBuilder.java:191)
	at org.eclipse.jdt.internal.compiler.Compiler.processCompiledUnits(Compiler.java:593)
	at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:462)
	at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:417)
	at org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.compile(AbstractImageBuilder.java:368)
	at org.eclipse.jdt.internal.core.builder.IncrementalImageBuilder.compile(IncrementalImageBuilder.java:330)
	at org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.compile(AbstractImageBuilder.java:305)
	at org.eclipse.jdt.internal.core.builder.IncrementalImageBuilder.build(IncrementalImageBuilder.java:135)
	at org.eclipse.jdt.internal.core.builder.JavaBuilder.buildDeltas(JavaBuilder.java:267)
[...]

PR #326, opened 2017-10-11
Make intrusive language server output less intrusive

bawachhe:
Every single time I remove lines or characters of code from a java file, the Language Server outputs a BadLocationException, which always interrupts my typing by popping up VSCode's output panel, and the only real solution seems to be to resize that panel to as little height as possible.  I've recently experienced different issues with stale filenames, which result in a FileNotFoundException.  Error this, exception that, and every time I attempt to navigate a java file, it seems, I am interrupted by another complaint from the Language Server.  This is untenable.

I tested this feature on VSCode v1.17.0 and it worked.  The output window still shows issues that the Language Server has, but it no longer interrupts me.

This issue has been very frustrating and exasperating for me and many others for some time.  While the exceptions themselves should be tracked down and fixed, they are separate issues to the user experience failure that is the output window consistently interrupting me, so please consider as such and merge/release this modification.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,7 +3,7 @@
 
 import * as path from 'path';
 import { workspace, extensions, ExtensionContext, window, StatusBarAlignment, commands, ViewColumn, Uri, CancellationToken, TextDocumentContentProvider, TextEditor, WorkspaceConfiguration, languages, IndentAction, ProgressLocation, Progress } from 'vscode';
-import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, ServerOptions, Position as LSPosition, Location as LSLocation } from 'vscode-languageclient';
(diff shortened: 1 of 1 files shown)


fbricon:
@bawachhe thanks for your contribution! Before we can proceed with merging it, can you please sign your commit? See https://github.com/probot/dco#how-it-works

fbricon:
@bawachhe your PR fails to build. See https://travis-ci.org/redhat-developer/vscode-java/jobs/286787939

```
src/extension.ts(58,51): error TS2551: Property 'never' does not exist on type 'typeof RevealOutputChannelOn'. Did you mean 'Never'?
```

bawachhe:
I fixed everything.

fbricon:
@bawachhe thanks again for your PR!

We'll keep on working towards resolving #314. Once we stabilize the server's behaviour, we'll see if showing the output on errors becomes acceptable, eventually.

PR #341, opened 2017-10-24
Add organize imports command

### LINKED ISSUE #253: Duplicate import on "Organize imports" action
Duplicate import on "Organize imports" action

##### Environment
- Operating System: Windows 10 x64
- JDK version: 1.8.0_131
- Visual Studio Code version: 1.13.1
- Java extension version: 0.7.0

step1
step2

yaohaizh:
Fixes #253

@fbricon @aeschli @gorkem

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,6 +150,11 @@
         "command": "java.open.serverLog",
         "title": "Open Java Language Server log file",
         "category": "Java"
+      },
+      {
+        "command": "java.edit.organizeImports",
+        "title": "Organize Imports",
+        "category": "Java"
       }
     ],
     "keybindings": [
@@ -159,7 +164,16 @@
         "mac": "cmd+alt+u",
         "when": "editorFocus"
       }
-    ]
+    ],
+    "menus": {
+      "editor/context": [
+        {
+          "command": "java.edit.organizeImports",
+          "when": "resourceLangId == java",
+          "group": "1_modification"
+        }
+      ]
+    }
   },
   "scripts": {
     "vscode:prepublish": "tsc -p ./",
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -64,8 +64,13 @@ export namespace Commands {
      */
     export const EXECUTE_WORKSPACE_COMMAND = 'java.execute.workspaceCommand';
 
-	 /**
-	 * Open Java Language Server Log file
-	 */
-	export const OPEN_SERVER_LOG = 'java.open.serverLog';
+    /**
+    * Open Java Language Server Log file
+    */
+    export const OPEN_SERVER_LOG = 'java.open.serverLog';
+
+    /**
+     * Organize Java file imports command from language server
+     */
+    export const EDIT_ORGANIZE_IMPORTS = 'java.edit.organizeImports';
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2,7 +2,7 @@
 'use strict';
 
 import * as path from 'path';
-import { workspace, extensions, ExtensionContext, window, StatusBarAlignment, commands, ViewColumn, Uri, CancellationToken, TextDocumentContentProvider, TextEditor, WorkspaceConfiguration, languages, IndentAction, ProgressLocation, Progress } from 'vscode';
+import { workspace, WorkspaceEdit, extensions, ExtensionContext, window, StatusBarAlignment, commands, ViewColumn, Uri, CancellationToken, TextDocumentContentProvider, TextEditor, Range, WorkspaceConfiguration, languages, IndentAction, ProgressLocation, Progress } from 'vscode';
 import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, RevealOutputChannelOn, ServerOptions, Position as LSPosition, Location as LSLocation } from 'vscode-languageclient';
 import { collectionJavaExtensions } from './plugin'
 import { prepareExecutable, awaitServerConnection } from './javaServerStarter';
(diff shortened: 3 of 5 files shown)


src/commands.ts
```diff
@@ -67,5 +67,15 @@ export namespace Commands {
 	 /**
 	 * Open Java Language Server Log file
 	 */
-	export const OPEN_SERVER_LOG = 'java.open.serverLog';
+    export const OPEN_SERVER_LOG = 'java.open.serverLog';
```
ankon: Something happened with the indentation here and below.

gorkem:
Can you also add the new command to tests at https://github.com/redhat-developer/vscode-java/blob/master/test/extension.test.ts#L25

src/commands.ts
fbricon: basically tabs were replaced with spaces. I think we need to set a project preference for indentation. But until then we should try to keep indentation coherent all over the place.
Ultimately we can do a major sweep to format all files at once, but that's the subject of another commit.

fbricon:
I can't bind a keyboard shortcut to the organize import command

package.json
```diff
@@ -150,6 +150,11 @@
         "command": "java.open.serverLog",
         "title": "Open Java Language Server log file",
         "category": "Java"
+      },
+      {
+        "command": "vscode.java.edit.organizeImports",
```
fbricon: Can we skip the vscode prefix? Having all commands under the java.* is more consistent

package.json
```diff
@@ -159,7 +164,16 @@
         "mac": "cmd+alt+u",
         "when": "editorFocus"
       }
-    ]
+    ],
+    "menus": {
+      "editor/context": [
+        {
+          "command": "vscode.java.edit.organizeImports",
```
fbricon: again, skip vscode prefix

src/commands.ts
```diff
@@ -67,5 +67,15 @@ export namespace Commands {
...
+    
+    /**
+	 * Organize Java file imports command from language server
+	 */
+    export const EDIT_ORGANIZE_IMPORTS = 'java.edit.organizeImports';
+    
+    /**
+	 * VSCode UI for organize imports commands
+	 */
+	export const VSCODE_EDIT_ORGANIZE_IMPORTS = 'vscode.java.edit.organizeImports';
```
fbricon: Does having the same command id for client and server cause any issue?

src/extension.ts
```diff
@@ -160,6 +160,22 @@ export function activate(context: ExtensionContext) {
 						}
 					});
 
+					commands.registerCommand(Commands.VSCODE_EDIT_ORGANIZE_IMPORTS, async (uri: Uri) => {
```
fbricon: need to check the uri is actually an Uri

src/extension.ts
```diff
@@ -160,6 +160,22 @@ export function activate(context: ExtensionContext) {
...
+						const edit = await <any>commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.EDIT_ORGANIZE_IMPORTS, uri.toString());
+						const workspaceEdit = new WorkspaceEdit();
+
```
fbricon: any way to reuse what we have L156?

package.json
```diff
@@ -159,7 +164,16 @@
...
+          "when": "resourceLangId == java",
```
fbricon: can we call the command from a folder in the explorer?

package.json
yaohaizh: This cannot be skip. There are two commands here actually. 
1. One is appeared on the VSCode UI side, which is starting with prefix "vscode"
2. Another one is just starting with "java...", which used for underlying talking with JDT.LS, which is neutral.

src/commands.ts
yaohaizh: I would like to keep the indentation consistent, and will accept either tab or whitespace.

As I went through the vscode-java codebase, there are mixed with tab and whitespaces. We need make them consistent anyway.

package.json
```diff
@@ -159,7 +164,23 @@
...
+        {
+          "command": "java.edit.organizeImports",
+          "when": "resourceLangId == java",
+          "group": "1_modification"
+        }
+      ],
+      "explorer/context": [
+        {
+          "command": "java.edit.organizeImports",
+          "when": "resourceLangId == java",
```
fbricon: can we enable it for folders too?

fbricon:
Any idea why calling a keyboard shortcut doesn't work? With:
```
{
    "key": "alt+cmd+o",
    "when": "editorTextFocus && !editorReadnly",
    "command": "java.edit.organizeImports"
}
```

the command receives an empty Object, instead of a URI, so nothing happens

yaohaizh:
It seems the keybinding doesn't provide any parameters as the context menu do. We need change the behavior of these commands.

yaohaizh:
For adding commands of "Organize Imports" in folder/workspace in VSCode, there is an issue here. We cannot determine current project(Could we? ). If so, the "Organize Imports" will be appeared in all VSCode instance, which is not desired behavior.

fbricon:
Ok, we can figure out how to get folders working later on the client side. Though I still think we can have it supported on the server side

aeschli:
@yaohaizh If you want to limit the command to only show up for certain files, this is how it works: https://github.com/Microsoft/vscode/blob/cabf4b2997fddf7eabff83d4d22ffa72381545b1/extensions/markdown/package.json#L156

fbricon:
@aeschli that's what @yaohaizh did. The question is how to enable it for folders of java projects?

fbricon:
Merged as https://github.com/redhat-developer/vscode-java/commit/977973499608ccd5e64115876b250c5dae062023 (with keybinding to `Shift+Alt+o`)

PR #357, opened 2017-11-04
fix compileworkspace command

ansyral:
Signed-off-by: xuzho <xuzho@microsoft.com>


fbricon:
The patch looks good but what problem is it fixing?

ansyral:
I expected the command would return compile status

PR #367, opened 2017-11-15
A command for invoking rename action

gorkem:
Adds a new command that invokes the rename action of the
editor.

requires https://github.com/eclipse/eclipse.jdt.ls/pull/430

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -15,6 +15,16 @@
 				"kind": "build",
 				"isDefault": true
 			}
+		},{
+			"label": "watch-server",
+			"type": "process",
+			"command": "npm",
+			"args": [
+				"run","watch-server"
+			],
+			"problemMatcher": [],
+			"isBackground": true
+
 		}
 	]
 }
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -77,5 +77,10 @@ export namespace Commands {
     /**
      * Organize Java file imports command from language server
      */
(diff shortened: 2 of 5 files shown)


fbricon:
It doesn't work for method renames:

nov-15-2017 11-12-03

In Eclipse, method renaming is not propagated.

gorkem:
Yes. same behavior also on Fields too. Our rename logic is not handling it correctly, regardless of this PR. I think I will handle that separately.

fbricon:
Instead of returning a rename command, it'll be easier to return a textedit with the new name instead (possibly as a `${newName}` placeholder.

gorkem:
That would not work. It is the same situation it would be same experience as extract method code action as we have it today.  You would have to invoke a separate rename operation to correct it. Also what is the point of avoiding the rename command?

fbricon:
If there are duplicates, it's impossible to know if the duplicate you're fixing needs to see its name change propagated. Eclipse is inconsistent there. for local variables, renaming is propagated, but not for methods.
So if rename is just for the current selection (like for methods), it doesn't need the full blown rename process. We could make it behave the same for variables, for consistency purposes.

It really depends on what we're expecting from a duplicate code action. Once we settle on a behaviour, it'll narrow down our options.

gorkem:
The inconsistency is on the rename process, it needs to be fixed on the rename operation. The behavior of invoking the rename action is the correct one, IMHO.

gorkem:
closing, as this functionality is no longer needed

Issue #237, opened 2017-06-09
Contribute Classpath Dependencies view


fbricon:
Similar to npm dependencies:

![](https://code.visualstudio.com/images/1_13_views.png)

we could display project dependencies in a custom view, see https://code.visualstudio.com/updates/v1_13#_custom-views

chvndb:
This feature would make this plugin a complete viable replacement for other IDE's like Eclipse.

blakematis:
Did this get released I do not see a resolution just many duplicate issues with the same information?

PR #389, opened 2017-12-14
Add dependency explorer

yaohaizh:
@fbricon @gorkem 

https://github.com/redhat-developer/vscode-java/issues/237

The gif/png is from eclipse. Will it bring legal issue? If you guys thanks we can implement the feature in the vscode-java repo. I'll find someone to provide the icons if the eclipse's icons is not appropriate for us to use.


yaohaizh:
dependency

fbricon:
Nice work @yaohaizh! That's an excellent starting point! 
So here are some initial remarks about the UX/UI (I haven't gone through the code yet):

- The view should be **Java dependencies**
- reusing the Eclipse icons is fine, since everything is EPL
- We need to add a `scope` level, as scope is added as a classpath attribute for both Maven and Gradle dependencies. Typically we should display test and compile nodes (or any other random scope that Gradle has).
- For Java 9 projects requiring modules say java.sql, only java.base is rendered.
- Classes from the default package should not be displayed in an empty package, but at the root of the tree:
screen shot 2017-12-14 at 9 33 05 am
vs
screen shot 2017-12-14 at 9 34 56 am
- opening classes from the view opens Untitled-1, then Untitled-2, then Untitled-3, then Untitled-2then Untitled-3, and so on. We should an editor with the actual class name, same as for regular navigation 
- it would be nice to be able to display the full path to the binary, and even better, being able to right click and open a system explorer/finder to that location
- in the absence of a notification system, changes to build files should probably trigger an automatic refresh of the view after say, 1sec (configurable? not sure). 

We need to figure out how to handle modulepath vs classpath

fbricon:
Going back to vscode after some time being idle, the icons have gone missing:

screen shot 2017-12-14 at 12 14 22 pm

yaohaizh:
@fbricon  For the last two points, I would put them at the advanced features after complete the basic.  

Most concern here is that what scope for? Any details for how this value used in the eclipse?

yaohaizh:
A separate extension for this. 
https://marketplace.visualstudio.com/items?itemName=yaozheng.vscode-java-copilot

PR #402, opened 2018-01-04
Add auto organize imports action to save action.

btstream:
Add an option for auto organize imports to save action.

### DIFF
--- a/package.json
+++ b/package.json
@@ -153,7 +153,7 @@
           "description": "Enable/disable default Java formatter",
           "scope": "window"
         },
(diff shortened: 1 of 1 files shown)


bonaluo:
There is two ways to use "auto organize imports".

- First

1. set "java.saveActions.organizeImports": true,
2. don't use autosave

- Second

1. set "editor.codeActionsOnSave": {"source.organizeImports": true},
2. set "files.autoSave": "onFocusChange",

The second can use autosave, but only "onFocusChange" or "onWindowChange"

PR #434, opened 2018-02-07
Add regular for loop snippet

rgcv:
I was wondering why there wasn't a snippet for regular `for` loops.

Hoping that it was just forgotten somewhere down the line and not implemented because it was deemed unnecessary or for any other reason, I'm submitting this proposal.

The prefix `fori` is inspired on ST3's respective snippet. The code itself is a mixture of "creativity" and the already implemented `foreach` snippet.


rgcv:
As I submitted the PR, I found another snippet could be added for iterating in a descending fashion, to complement the ascending order of iteration. Either that, or generalize this one to the point it's just a regular for loop inlining the words `init`, `condition` and `step` into the head of the loop.

To prevent further commits or changes, I'll hold for a reply or an acknowledgement and fix the DCO "sign-off" issue when and if need be.

snippets/java.json
```diff
@@ -22,6 +22,15 @@
...
 		"description": "Print to standard err"
 	},
+	"fori": {
+		"prefix": "fori",
+		"body": [
+			"for (${1:type} ${2:var} = ${3:lo}; $2 < ${4:hi}; $2++) {",
+			"\t$5",
+			"}"
+		],
+		"description": "For loop"
```
fbricon: "description": "Indexed For loop"

snippets/java.json
```diff
@@ -22,6 +22,15 @@
 		],
 		"description": "Print to standard err"
 	},
+	"fori": {
+		"prefix": "fori",
+		"body": [
+			"for (${1:type} ${2:var} = ${3:lo}; $2 < ${4:hi}; $2++) {",
```
fbricon: `for (${1:int} ${2:i} = ${3:0}; $2 < ${4:max}; $2++) {` provides better sensible default IMHO

fbricon:
> another snippet could be added for iterating in a descending fashion, to complement the one proposed, which iterates upwards. Either that, or generalize it so that it's just a regular for loop where the initialization, condition and step are just inlined

There are multiple possible variations of the indexed for loop. Honestly, I don't think it's necessary to add more than the common one.

snippets/java.json
rgcv: Thank you! I'm pretty terrible with names..

snippets/java.json
rgcv: Again, I did base myself somewhat on the already existing `foreach` snippet, generalizing the type and name of the variable, straying a bit from the conventional header suggested.

However, I do like the specificity and I do agree with you, since one aims to be fast and productive. Such comes from inlining the most common case (indexed integer-based ascending for loop).

I'll take the ball and run with it! Your way it is :)

rgcv:
Soon realised **all** commits required sign-off.. *sigh* sorry for missing that, was assuming only the last one had to follow that requirement.. which didn't really make sense. Force pushed the changes with a proper "Signed-off-by" line, all looks good!

fbricon:
I squashed all commits into one, as there's no need for keeping history here.
@rgcv thanks again for your contribution!

rgcv:
@fbricon Alright! Glad to contribute! Thank you as well for the receptiveness. I'm fairly new to contributing to OSS projects, so starting slow and all.. Pretty fearful guy, but that was a nice experience. Thank you so much yet again! :+1:

PR #470, opened 2018-03-16
Watch files under **/src/**

### LINKED ISSUE #400: Non-java file changes are not processed
a spring boot project,only the  .java file can be compiled,when I modify the  file which in /src/main/resources floder that the target/classes file is not  changed.How can I change?

fbricon:
Fixes #400


fbricon:
Obviously those will be ignored. But I'd prefer we avoid listening to everything, like *.class. while this is an incomplete solution, it should cover most cases. A more complete solution would be to expose a way to let users define folders to watch, but we can do that later.

PR #476, opened 2018-03-20
Experimental execute client-side command protocol extension

kdvolder:
This is the other half of https://github.com/eclipse/eclipse.jdt.ls/pull/596

It adds client-side handling for `workspace/executeCommand` request. This allows the language server to request command execution on the client-side.

It fits in with the discussion here: https://github.com/eclipse/eclipse.jdt.ls/issues/595

### DIFF
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -1,6 +1,6 @@
 'use strict';
 
-import { RequestType, NotificationType, TextDocumentIdentifier} from 'vscode-languageclient';
+import { RequestType, NotificationType, TextDocumentIdentifier, ExecuteCommandParams } from 'vscode-languageclient';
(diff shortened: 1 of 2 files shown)


src/protocol.ts
```diff
@@ -94,4 +94,8 @@ export namespace ActionableNotification {
 
 export namespace CompileWorkspaceRequest {
     export const type = new RequestType<boolean, CompileWorkspaceStatus, void, void>('java/buildWorkspace');
+}
+
+export namespace ExecuteClientCommandRequest {
+    export const type = new RequestType<ExecuteCommandParams, any, void, void>("workspace/executeClientCommand");
```
fbricon: should be single quotes

PR #482, opened 2018-03-28
add a reference parameter to openJDK download url

gorkem:
Signed-off-by: Gorkem Ercan <gorkem.ercan@gmail.com>


fbricon:
from=vscode-java ?

gorkem:
sorry m8 that is what the analytics man said :)

fbricon:
if we don't have a choice then +1

PR #492, opened 2018-04-04
[Discussion] [Live Share] Restricting language services on file scheme

lostintangent:
In preparation for [Visual Studio Live Share](http://aka.ms/vsls) adding support for "guests" to receive remote language services for Java, this PR simply updates the current `DocumentSelector` to be limited to `file` and `untitled` (unsaved) files. This way, when someone has this extension installed, and joins a Live Share session (where files use the `vsls:` scheme), their language services will be entirely derived from the remote/host side, which provides a more accurate and project-wide experience (guests in Live Share don't have local file access to the project they're collaborating with).

If someone joins a Java project using Live Share, and doesn't have this extension installed, then they will automatically receive language services from the host (which is awesome! 🎉), so this PR is simply an optimization for the case where collaborating developers both have the Java extension installed. Additionally, this wouldn't impact the "local" Java development experience.

*Note: As an example, the TypeScript/JavaScript language services that come in-box with VS Code [already have this scheme restriction](https://github.com/Microsoft/vscode/blob/master/extensions/typescript-language-features/src/utils/fileSchemes.ts#L12), and so this PR replicates that behavior.*

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -86,7 +86,6 @@ export function activate(context: ExtensionContext) {
 				let languageClient = new LanguageClient('java', 'Language Support for Java', serverOptions, clientOptions);
 				languageClient.registerProposedFeatures();
 				languageClient.onReady().then(() => {
-					toggleItem(window.activeTextEditor, item);
 					languageClient.onNotification(StatusNotification.type, (report) => {
 						switch (report.type) {
 							case 'Started':
@@ -242,7 +241,8 @@ export function activate(context: ExtensionContext) {
 				// client can be deactivated on extension deactivation
 				context.subscriptions.push(disposable);
 				context.subscriptions.push(onConfigurationChange());
-
+				toggleItem(window.activeTextEditor, item);
+				
 				function applyWorkspaceEdit(obj) {
 					let edit = languageClient.protocol2CodeConverter.asWorkspaceEdit(obj);
 					if (edit) {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -242,7 +242,7 @@ export function activate(context: ExtensionContext) {
 				context.subscriptions.push(disposable);
 				context.subscriptions.push(onConfigurationChange());
 				toggleItem(window.activeTextEditor, item);
-				
+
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -38,7 +38,10 @@ export function activate(context: ExtensionContext) {
 				// Options to control the language client
 				let clientOptions: LanguageClientOptions = {
 					// Register the server for java
-					documentSelector: ['java'],
+					documentSelector: [
+						{ scheme: 'file', language: 'java' },
+						{ scheme: 'untitled', language: 'java' }
```
fbricon: You also need 
```
{ scheme: 'jdt', language: 'java' },
```
else we lose all the java support (hover, navigation...) when opening sources (try ctrl+click on String)

fbricon:
@lostintangent your "Visual Studio Live Share" link is dead, do you have a better one?

src/extension.ts
lostintangent: Good catch! Just fixed this.

lostintangent:
@fbricon Apologies! I forgot to add the `http://` prefix to the link in my PR description (that's fixed now). https://code.visualstudio.com/visual-studio-live-share

src/extension.ts
```diff
@@ -82,6 +85,7 @@ export function activate(context: ExtensionContext) {
 				let languageClient = new LanguageClient('java', 'Language Support for Java', serverOptions, clientOptions);
 				languageClient.registerProposedFeatures();
 				languageClient.onReady().then(() => {
+					toggleItem(window.activeTextEditor, item);
```
fbricon: I'm not too comfortable moving the toggleItem call here. It's currently displayed as soon as the extension starts, regardless of the server state. Now, if there's an error, it won't be displayed, so you won't be able to open the server output log from there.

lostintangent: Makes sense. I'll revert this change, as it isn't critical to the main intent of this PR.

fbricon:
@lostintangent thanks for the patch!

PR #500, opened 2018-04-09
Second argument is replaced upon completing the first argument

### LINKED ISSUE #462: Second argument is replaced upon completing the first argument in a method call
**Moved from https://github.com/Microsoft/vscode-java-pack/issues/12:**

- VSCode Version: Latest Insiders
- OS Version: Mac OS

Steps to Reproduce:
1. Insert cursor before the first argument (that accesses some property) to the method.
2. Start typing new argument that also accesses property. 
3. On suggestion appearance, press Tab to enter suggestion value
It replaces the old first argument instance with the property from the new argument. 

vscode

Sample code:
```java
public class Test {
    public int testInt;

    public boolean method(int x, int y, int z) {
        return true;
    }
}

public class Main {
        Test t = new Test();
        t.method(t.testInt, t.testInt);
    }
}
```

snjeza:
Fixes #462 
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/619

### DIFF
--- a/package.json
+++ b/package.json
@@ -182,6 +182,12 @@
           "description": "Enable/disable the 'auto build'",
           "scope": "window"
         },
+        "java.completion.overwrite": {
+          "type": "boolean",
+          "default": true,
+          "description": "When set to true, code completion overwrites the current text. When set to false, code is simply added instead.",
+          "scope": "window"
+        },
         "java.completion.favoriteStaticMembers": {
           "type": "array",
           "description": "Defines a list of static members or types with static members. Content assist will propose those static members even if the import is missing.",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -182,6 +182,12 @@
           "description": "Enable/disable the 'auto build'",
           "scope": "window"
         },
+        "java.completion.insert": {
+          "type": "boolean",
+          "default": true,
+          "description": "Completion inserts/overwrites",
```
gorkem: I think we need to explain on description what does true mean?

snjeza:
@gorkem I have updated the PR.

fbricon:
Honestly, from a user standpoint, I think it'd be clearer to have a `java.completion.overwrite` pref, that said 

> When set to `true`, code completion overwrites the current text. When set to `false`, code is simply  added instead.

snjeza:
@gorkem @fbricon I have updated the PR.

fbricon:
completion keeps overwriting the text, regardless of the preference

package.json
```diff
@@ -182,6 +182,12 @@
           "description": "Enable/disable the 'auto build'",
           "scope": "window"
         },
+        "java.completion.overwrite ": {
```
fbricon: beware of the trailing space, it messes up configuration

snjeza:
@gorkem @fbricon I have updated the PR.

PR #528, opened 2018-05-03
Add command to create new Java formatter settings

### LINKED ISSUE #521: Add command to create new Java formatter settings
Following up on https://github.com/redhat-developer/vscode-java/pull/517#issuecomment-384768400:
> The idea is to give users something to start from. So the command would be like "Add Java formatter settings", maybe ask for a file location, a complete formatter xml file would be added there and the java.formatter.url would be updated

snjeza:
Fixes #521 

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ The following commands are available:
 - `Java:Open Java Language Server log file`: opens the Java Language Server log file, useful for troubleshooting problems.
 - `Java:Force Java compilation` (`Shift+Alt+B`): manually triggers compilation of the workspace.
 - `Java:Organize imports` (`Shift+Alt+O`): Organize imports in the currently opened Java file.
+- `Java:Open Java formatter settings`: Open the Eclipse formatter settings. Creates a new settings file if none exists.
 
 Supported VS Code settings
 ==========================
--- a/package.json
+++ b/package.json
@@ -266,6 +266,11 @@
         "command": "java.edit.organizeImports",
         "title": "Organize Imports",
         "category": "Java"
+      },
+      {
+        "command": "java.open.formatter.settings",
+        "title": "Open Java formatter settings",
+        "category": "Java"
       }
     ],
     "keybindings": [
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -78,4 +78,8 @@ export namespace Commands {
      * Organize Java file imports command from language server
      */
     export const EDIT_ORGANIZE_IMPORTS = 'java.edit.organizeImports';
+    /**
+     * Open Java formatter settings
+     */
+    export const OPEN_FORMATTER = 'java.open.formatter.settings';
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2,7 +2,8 @@
 
 import * as path from 'path';
 import * as os from 'os';
-import { workspace, extensions, ExtensionContext, window, StatusBarAlignment, commands, ViewColumn, Uri, CancellationToken, TextDocumentContentProvider, TextEditor, WorkspaceConfiguration, languages, IndentAction, ProgressLocation, Progress } from 'vscode';
+import * as fs from 'fs';
+import { workspace, extensions, ExtensionContext, window, StatusBarAlignment, commands, ViewColumn, Uri, CancellationToken, TextDocumentContentProvider, TextEditor, WorkspaceConfiguration, languages, IndentAction, ProgressLocation, Progress, Position, Range, InputBoxOptions } from 'vscode';
 import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, RevealOutputChannelOn, ServerOptions, Position as LSPosition, Location as LSLocation } from 'vscode-languageclient';
 import { collectionJavaExtensions } from './plugin';
 import { prepareExecutable, awaitServerConnection } from './javaServerStarter';
@@ -238,6 +239,9 @@ export function activate(context: ExtensionContext) {
 				// Register commands here to make it available even when the language client fails
 				commands.registerCommand(Commands.OPEN_SERVER_LOG, () => openServerLogFile(workspacePath));
 
+				let extensionPath = context.extensionPath;
+				commands.registerCommand(Commands.OPEN_FORMATTER, async () => openFormatter(extensionPath));
+
 				// Push the disposable to the context's subscriptions so that the
 				// client can be deactivated on extension deactivation
 				context.subscriptions.push(disposable);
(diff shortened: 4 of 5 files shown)


fbricon:
I think we should pivot a little bit: make that command "Open Jafa formatter settings". if url is set and file exists, open it, else add a file, same logic as current PR

src/extension.ts
```diff
@@ -237,6 +237,21 @@ export function activate(context: ExtensionContext) {
...
 				commands.registerCommand(Commands.OPEN_SERVER_LOG, () => openServerLogFile(workspacePath));
 
+				commands.registerCommand(Commands.SET_FORMATTER, async () => {
+					let formatterUrl: string = getJavaConfiguration().get('format.settings.url');
+					const options: InputBoxOptions = {
+						value: formatterUrl,
+						prompt: 'please enter URL or Path:',
+						ignoreFocusOut: true
+					};
+					await window.showInputBox(options).then(f => {
```
fbricon: you need to create the actual settings from a template

src/extension.ts
```diff
@@ -237,6 +237,21 @@ export function activate(context: ExtensionContext) {
...
+						if (f) {
+							getJavaConfiguration().update('format.settings.url', f, true);
+							commands.executeCommand('workbench.action.openGlobalSettings');
```
fbricon: you want to open the actual settings

snjeza:
@fbricon I have updated the PR. It works in the following way:

If the java.format.settings.url property isn't set, VS Code will set it to 'https://raw.githubusercontent.com/google/styleguide/gh-pages/eclipse-java-google-style.xml' (GoogleStyle formatter).
If the property is set and points to an existing file or an URL, VS Code will open that file or URL.
Otherwise, VS Code will enable a user to enter a new property.

README.md
```diff
@@ -57,6 +57,7 @@ The following commands are available:
 - `Java:Open Java Language Server log file`: opens the Java Language Server log file, useful for troubleshooting problems.
 - `Java:Force Java compilation` (`Shift+Alt+B`): manually triggers compilation of the workspace.
 - `Java:Organize imports` (`Shift+Alt+O`): Organize imports in the currently opened Java file.
+- `Java:Add/Open Java formatter settings`: Add or open Java formatter settings.
```
fbricon: Add/Open => Open
Open the Java formatter settings. Creates a new settings file if none exists.

package.json
```diff
@@ -264,6 +264,11 @@
         "command": "java.edit.organizeImports",
         "title": "Organize Imports",
         "category": "Java"
+      },
+      {
+        "command": "java.add.formatter",
```
fbricon: java.open.formatter.settings

package.json
```diff
@@ -264,6 +264,11 @@
...
+        "title": "Add/Open Java formatter settings",
```
fbricon: Open Java formatter settings

src/commands.ts
```diff
@@ -78,4 +78,8 @@ export namespace Commands {
      * Organize Java file imports command from language server
      */
     export const EDIT_ORGANIZE_IMPORTS = 'java.edit.organizeImports';
+    /**
+     * Add Java formatter settings
+     */
+    export const ADD_OPEN_FORMATTER = 'java.add.formatter';
```
fbricon: OPEN_FORMATTER  = java.open.formatter.settings

src/commands.ts
```diff
@@ -78,4 +78,8 @@ export namespace Commands {
      * Organize Java file imports command from language server
      */
     export const EDIT_ORGANIZE_IMPORTS = 'java.edit.organizeImports';
+    /**
+     * Add Java formatter settings
```
fbricon: Open Java formatter settings

src/extension.ts
```diff
@@ -424,4 +427,55 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+			  if (!doc) {
+				  addFormatter(formatterUrl);
+			  }
+			  return window.showTextDocument(doc, window.activeTextEditor ?
+				  window.activeTextEditor.viewColumn : undefined)
+				  .then(editor => !!editor);
+		  }, () => false)
+		  .then(didOpen => {
+			  if (!didOpen) {
+          window.showWarningMessage('Could not open Formatter Settings file');
```
fbricon: indentation is off here

src/extension.ts
```diff
@@ -424,4 +427,55 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+			  } else {
+			    return didOpen;
+        }
+      });
+    }
+  }
+  addFormatter('https://raw.githubusercontent.com/google/styleguide/gh-pages/eclipse-java-google-style.xml');
+}
+
+async function addFormatter(formatterUrl) {
```
fbricon: I'd prefer if we added an actual settings to the project, since, most likely users will want to tweak the values. Pointing to a read-only file is not helping.

src/extension.ts
```diff
@@ -424,4 +427,55 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+  const options: InputBoxOptions = {
+    value: formatterUrl,
+    prompt: 'please enter URL or Path:',
+    ignoreFocusOut: true
+  };
+  await window.showInputBox(options).then(f => {
+    if (f) {
+      getJavaConfiguration().update('format.settings.url', f, true);
+      commands.executeCommand('workbench.action.openGlobalSettings');
```
fbricon: I'd prefer if the workspace settings were modified, not the global ones, as it's easier to share in an SCM.
Unless a standalone file is opened, in which case global settings is the only solution

snjeza:

src/extension.ts
```diff
@@ -424,4 +428,76 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+	}
+	let global = workspace.workspaceFolders === undefined;
+	let fileName = 'eclipse-formatter.xml';
+	let file;
+	if (!global) {
+		file = path.join(workspace.workspaceFolders[0].uri.path, fileName);
+	} else {
+		file = path.join(workspacePath, '.metadata', fileName);
+	}
+	if (!fs.existsSync(file)) {
```
fbricon: we should not create the formatter file without the user's approval. Should be more like "no formatter exists, do you want to create a new file?"

src/extension.ts
```diff
@@ -424,4 +428,76 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+			}
+		}
+	}
+	let global = workspace.workspaceFolders === undefined;
+	let fileName = 'eclipse-formatter.xml';
+	let file;
+	if (!global) {
+		file = path.join(workspace.workspaceFolders[0].uri.path, fileName);
+	} else {
+		file = path.join(workspacePath, '.metadata', fileName);
```
fbricon: temp workspace file is not easily accessible and ... temporary. I'd rather use `<user.home>\.vscode[-flavor]\redhat.java` instead

snjeza:

src/extension.ts
```diff
@@ -424,4 +428,80 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+				return openDocument(document);
+			}
+		}
+	}
+	let global = workspace.workspaceFolders === undefined;
+	let fileName = 'eclipse-formatter.xml';
+	let file;
+	if (!global) {
+		file = path.join(workspace.workspaceFolders[0].uri.path, fileName);
+	} else {
```
fbricon: the problem with extension path is it'll change after each release. That's why I proposed to use the extension name (without the version number)

snjeza:

src/extension.ts
```diff
@@ -424,4 +428,84 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+		}
+		file = path.join(root, fileName);
+	}
+	if (!fs.existsSync(file)) {
+		const selection = await window.showQuickPick(['Yes', 'No'], { 'placeHolder': 'no formatter exists, do you want to create a new file?' });
+		if (selection === 'Yes') {
+			fs.createReadStream(defaultFormatter).pipe(fs.createWriteStream(file));
+		}
+	}
+	getJavaConfiguration().update('format.settings.url', file, global);
```
fbricon: we want to keep the file path relative, if possible

fbricon:
There's something weird with the workflow. When no url is set and it asks if you want to create the file, there's no prompt for the location

src/extension.ts
fbricon: Yeah that's bad https://www.screencast.com/t/r7plcZ7UB

src/extension.ts
```diff
@@ -424,4 +428,84 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+			if (fs.existsSync(document)) {
+				return openDocument(document);
+			}
+		}
+	}
+	let global = workspace.workspaceFolders === undefined;
+	let fileName = 'eclipse-formatter.xml';
+	let file;
+	if (!global) {
+		file = path.join(workspace.workspaceFolders[0].uri.path, fileName);
```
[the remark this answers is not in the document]
snjeza: In this case, the formatter path is relative under the first workspace folder.

src/extension.ts

src/extension.ts
```diff
@@ -424,4 +428,84 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+			fs.mkdirSync(root);
+		}
+		file = path.join(root, fileName);
+	}
+	if (!fs.existsSync(file)) {
+		const selection = await window.showQuickPick(['Yes', 'No'], { 'placeHolder': 'no formatter exists, do you want to create a new file?' });
+		if (selection === 'Yes') {
+			fs.createReadStream(defaultFormatter).pipe(fs.createWriteStream(file));
+		}
+	}
```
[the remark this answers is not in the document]
fbricon: Not talking about the template. My point is the user could have several formatter settings ( 'eclipse-formatter.xml'  and others)  in his project, so he should be able to choose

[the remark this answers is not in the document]
snjeza: He can use the java.format.settings.url property.

fbricon:
I'm still not satisfied with the workflow.

What I expect is:
if formatter path/url exists => open it. That works for now
else open text box to enter new path (default value would be simply eclipse-formatter.xml)
THEN
   if path/file doesn't exist, show selection box (...do you want to create it? Yes/No), create file 
 accordingly
   then update settings, 
   then open file if it exists

snjeza:

src/extension.ts
```diff
@@ -425,4 +429,126 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+				} else {
+					openDocument(extensionPath, formatterUrl, defaultFormatter, relativePath);
+				}
+			}
+		}
+	});
+}
+
+async function createFormatter(extensionPath, file, defaultFormatter, relativePath) {
+	const selection = await window.showQuickPick(['Yes', 'No'], { 'placeHolder': 'no formatter exists, do you want to create the it?' });
```
fbricon: s/'no formatter exists'/file + ' does not exist'

snjeza:

snjeza:

src/extension.ts
```diff
@@ -425,4 +429,127 @@ function openServerLogFile(workspacePath): Thenable<boolean> {
...
+						if (!fs.existsSync(root)) {
+							fs.mkdirSync(root);
+						}
+						f = path.join(root, fileName);
+					}
+				}
+				getJavaConfiguration().update('format.settings.url', (relativePath !== null ? relativePath : f), global);
+				if (!fs.existsSync(f)) {
+					let name = relativePath !== null ? relativePath : f;
+					let msg = '\'' + name + '\' has been created. Do you want to create it?';
```
fbricon: does not exists. Do you want to create it?
Yes - No

fbricon:
if settings file doesn't exist, it's ignored when displaying the input box:
may-18-2018 11-43-46

Value in settings should always be shown, only fall back to default name if nothing is set

snjeza:

PR #529, opened 2018-05-07
Update vscode-language-client and vscode

gorkem:
upates to new versions

Signed-off-by: Gorkem Ercan <gorkem.ercan@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "bugs": "https://github.com/redhat-developer/vscode-java/issues",
   "preview": true,
   "engines": {
-    "vscode": "^1.15.0"
+    "vscode": "^1.22.0"
   },
   "repository": {
     "type": "git",
@@ -18,7 +18,9 @@
   },
   "categories": [
     "Programming Languages",
-    "Linters"
+    "Linters",
+    "Formatters",
+    "Snippets"
   ],
   "keywords": [
     "multi-root ready"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,15 +10,17 @@
...
+    "vscode": "^1.22.0"
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/redhat-developer/vscode-java"
   },
   "categories": [
-    "Programming Languages",
-    "Linters"
+    "Languages",
```
fbricon: "Programming Languages" was better.
screen shot 2018-05-10 at 9 35 30 am

package.json
```diff
@@ -321,7 +323,7 @@
   },
   "devDependencies": {
     "typescript": "^2.3.4",
-    "vscode": "^1.1.4",
+    "vscode": "1.1.17",
```
fbricon: why not ^1.1.17 or at least ~1.1.17 ?

package.json
```diff
@@ -10,15 +10,17 @@
   "bugs": "https://github.com/redhat-developer/vscode-java/issues",
   "preview": true,
   "engines": {
-    "vscode": "^1.15.0"
+    "vscode": "^1.22.0"
```
yaohaizh: If update the engine, I would suggest to update 1.23.0 since 1.23.0 provides two UI enhancement: view containers and webview API. These two features are great usage for advanced features.

package.json
gorkem: For some reason my vscode was complaining about`Programming Languages`, fixed

package.json
gorkem: fixed

package.json
gorkem: `^1.22.0` does not exclude `1.23.0` this extension does not specifically require those APIs to be installed.

PR #538, opened 2018-05-25
move class & interface snippets to server side

### LINKED ISSUE #505: Generate package declaration with class code snippet
When generating a code snippet for class, the class name is correct, but the package declaration is missing.
image

The language server already has the knowledge because user can fix this using quick fix suggestion. So please also generate the package declaration along with the class body.

Reported by @biezhi

SummerSun:
requires https://github.com/eclipse/eclipse.jdt.ls/pull/671/

Fixes #505

### DIFF
--- a/snippets/java.json
+++ b/snippets/java.json
@@ -49,31 +49,6 @@
 		],
 		"description": "Public constructor"
 	},
-	"Class Definition": {
-		"prefix": "class",
-		"body": [
-			"/**",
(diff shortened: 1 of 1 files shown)


fbricon:
Thanks for your contribution @SummerSun. Can you please sign-off your commit so we can merge it? See https://github.com/probot/dco#how-it-works

PR #540, opened 2018-05-25
disable organize imports to be shown in the output channel

### LINKED ISSUE #539: Organize imports shows in the output panel
untitled

jdneo:
fix #539

### DIFF
--- a/package.json
+++ b/package.json
@@ -300,7 +300,7 @@
       "editor/context": [
         {
           "command": "java.edit.organizeImports",
-          "when": "resourceLangId == java",
+          "when": "editorLangId == java",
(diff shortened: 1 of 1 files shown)


fbricon:
Thanks for your contribution @jdneo . Can you please sign-ff your commit so we can merge it? See https://github.com/probot/dco#how-it-works

fbricon:
So the patch works as advertised, just waiting for the signed-off commit before merging.

@yaohaizh didn't we enable the "organize imports" menu from the explorer view (it doesn't work currently)? I remember we couldn't do it for folders, but java files can have it enabled. Am I remembering wrong?

PR #552, opened 2018-06-05
add step to publish to static when...

nickboldt:
add step to publish to static when publishing to Marketplace; remove unneeded code (?) too

Change-Id: Iae731a81524bf8e416c0af67170b7378977efd47
Signed-off-by: nickboldt <nboldt@redhat.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,11 +18,10 @@ node('rhel7'){
 	sh "./mvnw clean verify -B -U -fae -e -Pserver-distro -DdisableP2Mirrors=true"
 
 	def files = findFiles(glob: '**/org.eclipse.jdt.ls.product/distro/**.tar.gz')
-	stash name: 'server_distro',includes :files[0].path
+	stash name: 'server_distro', includes :files[0].path
 }
 
 node('rhel7'){
-
 	stage 'Checkout vscode-java code'
 	deleteDir()
 	git url: 'https://github.com/redhat-developer/vscode-java.git'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -59,11 +58,18 @@ node('rhel7'){
...
+		unstash 'vsix'
 		withCredentials([[$class: 'StringBinding', credentialsId: 'vscode_java_marketplace', variable: 'TOKEN']]) {
 			def vsix = findFiles(glob: '**.vsix')
 			sh 'vsce publish -p ${TOKEN} --packagePath' + " ${vsix[0].path}"
 		}
 		archive includes:"**.vsix"
-	}//if publishMarketPlace
+	}// if publishToMarketPlace
+
+	if(publishToStable.equals('true')){
```
fbricon: We don't need a new publishToStable  parameter. This block should be executed if package was successufully pushed to the marketplace

nickboldt: yes, that was just for my own testing. Here's the cleaned up / squashed PR. https://github.com/redhat-developer/vscode-java/pull/552/commits/a1da5e3da8a2b349b3231e38baca3a93936c00ac

nickboldt:
Note that UPLOAD_LOCATION has already been added as a param on https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Adapters/view/VSCode/job/vscode-java_release/configure so once merged this should Just Work (tm).

fbricon:
Thanks for the patch @nickboldt!

Issue #557, opened 2018-06-10
Code fixes force tabs for indentation


MeikTranel:
When using a java code fixes, the extension forces tabs into a 4-Space indentation workspace

##### Environment
- Operating System: Windows 10 1803 (RS5) Insider Preview
- JDK version: 1.8.172
- Visual Studio Code version: 0.25.0-insider
- Java extension version: 0.26.0

##### Steps To Reproduce
1. Set workspace to space indentation
2. let the extension implement an interface

##### Current Result
Injects tabs
##### Expected Result
adheres to user->workspace-settings

fbricon:
Ok so that's interesting. Code snippets (like ctor) are indented with tabs, but vscode fixes the indentation on-the-fly, according to its preferences.
However CompletionItems are indented according to the project's `.settings/org.eclipse.jdt.core.prefs`, so implementing an interface's methods would have the server use the eclipse indentation settings, but vscode doesn't correct it when applying the changes, contrary to snippets. @aeschli is this intentional?

Anyways, @MeikTranel,  one way to mitigate the problem is to add this to your project's  `.settings/org.eclipse.jdt.core.prefs`:

> org.eclipse.jdt.core.formatter.tabulation.char=space
org.eclipse.jdt.core.formatter.tabulation.size=4

MeikTranel:
@fbricon What you said is exactly what sparked my confusion with the title. 
First i thought it was regular code snippets but then i found out it was actually only the fixes rioting.

Aren't code fixes a language server thing?

yaohaizh:
https://github.com/Microsoft/vscode/issues/52272

yaohaizh:
I logged an issue for vscode on this issue. 

On the language side, @fbricon  Which setting should be honor?  If we honer eclipse setting, there is nothing left to do.

fbricon:
Fixed with #617, thanks to @jdneo. Nice work!

HanKruiger:
This issue persists when the fix is triggered in a different file from where the fix is applied.

For example when in a class `Foo` I try to call a non-existing method on an object of class `Bar`, and trigger the fix to implement the new method in `Bar`, the new method uses tabs for indentation, whereas my setting prescribes spaces. (With the same settings, it _does_ correctly indent with spaces for implementing interfaces in the same file, as in @MeikTranel's example.)

I hope I made myself clear, and would happily elaborate if anything was unclear.

Cheers!

PR #617, opened 2018-08-27
Fix for indention issue

jdneo:
A workaround for issue #557.

For now, just send a range format request after `workspace.applyEdit()`

Some GIFs for illustration:

### Add unimplemented methods
unimplement

### Extract to local variable
local

### Extract to constant
const

### Extract to method
method

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,7 +3,7 @@
 import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
-import { workspace, extensions, ExtensionContext, window, StatusBarAlignment, commands, ViewColumn, Uri, CancellationToken, TextDocumentContentProvider, TextEditor, WorkspaceConfiguration, languages, IndentAction, ProgressLocation, Progress, Position, Range, InputBoxOptions } from 'vscode';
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -248,10 +248,30 @@ export function activate(context: ExtensionContext) {
...
+							if (currentEditor.document.uri.fsPath !== edit.entries()[0][0].fsPath) {
+								return;
+							}
+							let lastPosition = currentEditor.selection.active;
+							let changes = edit.entries()[0][1];
+							let startLineNum = changes[0].range.start.line;
+							let startCharacterNum = changes[0].range.start.character;
+							let endLineNum = startLineNum;
+							for (let newText of changes) {
+								endLineNum += newText.newText.split('\n').length - 1;
```
fbricon: What if the line delimiter is `\r\n`? The solution is probably platform dependent

fbricon:
The PR seems to work as expected so far. Except for the potential line-delimiter issue, my main concern is about the use of seemingly random (it's not) index values while manipulating objects. 
For future maintainers' sake, please add some more comments to the code as it's not obvious what is happening.

src/extension.ts
jdneo: Ok, let me change it to `split(/\r?\n/)`

jdneo:
@fbricon Updated the code to support the discontinuous condition. 

Sometimes the server side may response multiple ranges which are discontinuous. We need to range format each of them separately.

fbricon:
Testing that fix, I found that formatting with tabs always use spaces instead. See https://github.com/eclipse/eclipse.jdt.ls/issues/775

PR #628, opened 2018-09-05
Allow max 5 seconds for server to shutdown.

yaohaizh:
Currently, the server will immediately be killed by client when call stop, this will prevent server from saving status. Mediate this issue by following https://github.com/Microsoft/vscode/issues/35196

> The disposable assumes that everything can be executed sync which is not the case in the LSP. We later added async deactivate to VS Code's API which needs to be used for the LSP client. I might want to consider removing the return type form start in a version that allows breakage to make this more clear.
> 
> So for you that means that you shouldn't push the disposable into the array but implement a deactivate method for the extension and return the promise that is returned from the stop method. Then VS Code will wait max 5 seconds for your server to shutdown.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -13,6 +13,7 @@ import { StatusNotification, ClassFileContentsRequest, ProjectConfigurationUpdat
 
 let oldConfig;
 let lastStatus;
+let languageClient : LanguageClient;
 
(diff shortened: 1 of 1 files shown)


fbricon:
Seems to work as expected, merging.
Thanks @yaohaizh!

Issue #627, opened 2018-09-05
The language support always unavailable after change the method type


bobjiang1988:
##### Environment
- Operating System: Ubuntu 18.04
- JDK version: 1.8
- Visual Studio Code version: 1.26.1
- Java extension version: 0.30.0

##### Steps To Reproduce
1. I have a method like that
```
public void test() {
}
```

2. Change the method type to Integer
```
public Integer test() {
}
```

3. Then I got a syntax error, and here is the log
```
null
org.eclipse.jface.text.BadLocationException
	at org.eclipse.jface.text.TreeLineTracker.fail(TreeLineTracker.java:1054)
	at org.eclipse.jface.text.TreeLineTracker.lineByOffset(TreeLineTracker.java:268)
	at org.eclipse.jface.text.TreeLineTracker.getLineNumberOfOffset(TreeLineTracker.java:1122)
	at org.eclipse.jface.text.AbstractLineTracker.getLineNumberOfOffset(AbstractLineTracker.java:154)
	at org.eclipse.jface.text.AbstractDocument.getLineOfOffset(AbstractDocument.java:871)
	at org.eclipse.core.internal.filebuffers.SynchronizableDocument.getLineOfOffset(SynchronizableDocument.java:334)
	at org.eclipse.jdt.ls.core.internal.handlers.JsonRpcHelpers.toLine(JsonRpcHelpers.java:60)
	at org.eclipse.jdt.ls.core.internal.JDTUtils.toRange(JDTUtils.java:468)
	at org.eclipse.jdt.ls.core.internal.JDTUtils.toLocation(JDTUtils.java:382)
	at org.eclipse.jdt.ls.core.internal.JDTUtils.toLocation(JDTUtils.java:344)
	at org.eclipse.jdt.ls.core.internal.handlers.DocumentSymbolHandler.collectChildren(DocumentSymbolHandler.java:75)
	at org.eclipse.jdt.ls.core.internal.handlers.DocumentSymbolHandler.collectChildren(DocumentSymbolHandler.java:68)
	at org.eclipse.jdt.ls.core.internal.handlers.DocumentSymbolHandler.getOutline(DocumentSymbolHandler.java:52)
	at org.eclipse.jdt.ls.core.internal.handlers.DocumentSymbolHandler.documentSymbol(DocumentSymbolHandler.java:44)
	at org.eclipse.jdt.ls.core.internal.handlers.JDTLanguageServer.lambda$13(JDTLanguageServer.java:545)
[...]

PR #637, opened 2018-09-12
Adopt the send notification to avoid deadlock.

yaohaizh:
We have deadlock issue when installed both vscode-lombok & vscode-java-test: 
https://github.com/Microsoft/vscode-java-test/issues/260
And potential related issue: https://github.com/redhat-developer/vscode-java/issues/627

Adopt this https://github.com/eclipse/eclipse.jdt.ls/pull/779/files to client to avoid deadlock.

@fbricon

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -9,7 +9,7 @@ import { collectionJavaExtensions } from './plugin';
 import { prepareExecutable, awaitServerConnection } from './javaServerStarter';
 import * as requirements from './requirements';
 import { Commands } from './commands';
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -157,6 +157,11 @@ export function activate(context: ExtensionContext) {
 					languageClient.onRequest(ExecuteClientCommandRequest.type, (params) => {
 						return commands.executeCommand(params.command, ...params.arguments);
 					});
+
+					languageClient.onRequest(SendNotificationRequest.type, (params) => {
```
fbricon, line 161: @yaohaizh @testforstephen as @angelozerr noticed, shouldn't that be `languageClient .onNotification`?

PR #694, opened 2018-10-29
Fix indentation rules by copying the [ones used for js/ts](https://github.com/Microsoft/vscode/blob/41ddd41b71ea2bd5c6e49125beeb3e951f3d9c7c/extensions/typescript-language-features/src/features/languageConfiguration.ts#L17-L18).

### LINKED ISSUE #692: moveLinesUpAction indentation broken after if statement
Version: 1.29.0-insider
Commit: 5d68db08cda5accef740c69541178911b865d069
Date: 2018-10-26T05:13:31.757Z
Electron: 2.0.12
Chrome: 61.0.3163.100
Node.js: 8.9.3
V8: 6.1.534.41
Architecture: x64

Java Language Extension version: 0.33.0

OS: Linux Mint 19

When I run the command moveLinesUpAction/DownAction, indentation can be affected when deemed proper. In certain cases indentation is erroneously affected. The erroneous behaviour is not present when Java Language Support Extension is disabled or removed.

Observe the following code:

```
if (blah)
    someCode;

line1;
line2;
```

This code is properly indented and syntactically correct.
1. Put the caret at line2 and run moveLinesUpAction
2. Observe that in addition to swapping lines 1 and 2, line2 will erroneously have its indentation affected by adding an additional tab in front of it.

The previously error is rectified by adding brackets to the already syntactically correct if statement.

```
if (blah) {
    someCode;
}

line1;
line2;
```

In this new example, running moveLinesUpAction/DownAction  will not affect the indentation of line1/line2.

fbricon:
Fixes #692


fbricon:
mmm that sucks. Seems like this works with vscode insiders but not the last official release. @snjeza can you double check that please?

fbricon:
Looks like the same issue exists in typescript in the last vscode release. Works with insiders (I'm using the same patterns). So I'd say, we ship it anyway. It'll be available to users on regular vscode in early November.

snjeza:
I can also reproduce the issue with code-insiders-1.29.0. I have tested on Windows and Linux.

Version: 1.29.0-insider (system setup)
Commit: 2b06dd1ca31e407abed24f65e1804ae86180401d
Date: 2018-10-30T11:12:35.070Z
Electron: 2.0.12
Chrome: 61.0.3163.100
Node.js: 8.9.3
V8: 6.1.534.41
Architecture: x64

The issue can't be reproduced if we exclude the Java extension.

fbricon:
@snjeza FYI I opened https://github.com/Microsoft/vscode/issues/62198 upstream

fbricon:
@snjeza I'm taking another shot at this old issue, it seems to work now. Please review.

fbricon:
@snjeza let me know if you can't review this PR, I'll ask someone else

snjeza:
I still can reproduce the problem.

https://github.com/user-attachments/assets/3ccfca01-0de7-402c-b601-525af14b8841

fbricon:
damn, works for me
move-lines

fbricon:
wait your gif shows the problem is fixed on Java (yeah!), not typescript (we don't care). On the same vscode version, it also works for me on typescript.

snjeza:
@fbricon could you try the following code:

```
public class E1 {
    void test() {
        if (true)
        System.out.println("hello");

        line1;
        line2;
    }
}
```
and call moveLinesUpAction on line2

fbricon:
OK so if the line under the if is not indented, moving line 2 up de-increments it, which is not happening with this PR.

So we'd be swapping the bug from it's happening if line under if is incremented, to it's happening if line under it is not incremented. 
BUT, in the meantime we gain better (de)increment on next line after brace-less ifs:


https://github.com/user-attachments/assets/849d6162-db7f-4971-b0bd-68a519f18728

So still better than the existing IMO

Issue #233, opened 2017-06-08
Add ability to attach missing sources


fbricon:
If a class file has no known source, a disassembled source is displayed instead. 
There should be a way to easily attach missing sources to the jar containing the file in question.

Originally suggested by @jtsiros on https://marketplace.visualstudio.com/items?itemName=redhat.java#review-details

fbricon:
@yaohaizh @testforstephen I think this could be implemented via a menu/command with stock vscode-java

akaroml:
Propose a workflow:
1. Show "Open Folder" dialog to let users pick the location of the additional source
2. Persist the location into user settings - "java.additionalSourceLocations: []"
3. Language server follows the setting and works with those additional source locations

akaroml:
@testforstephen can you take a look?

testforstephen:
The ux is adding a context menu `Update Source Attachment` on the disassembled source editor (JdtContentProvider). When the user clicks the menu, pop up a `Open File` dialog to select the source jar/zip file. Then binding the source attachment to the binary path, and re-rendering the class file contents in VS Code.

fbricon:
Fixed with #699, thanks to @testforstephen 

<img width="821" alt="screen shot 2018-11-13 at 7 05 33 pm" src="https://user-images.githubusercontent.com/148698/48451159-1f79ca80-e777-11e8-8a9c-44a75a9b05b4.png">

mikedoy:
When I attach source, where is this information stored? I'm not seeing it in my settings.json.

PR #699, opened 2018-10-31
Add context menu to update source attachment for the binary classfile

testforstephen:

It fixes the bug https://github.com/redhat-developer/vscode-java/issues/233 (Add ability to attach missing source). This PR will enable the context menu `Update Source Attachment` on the jdt content provider. See the peer PR on eclipse.jdt.ls https://github.com/eclipse/eclipse.jdt.ls/pull/837

### DIFF
--- a/package.json
+++ b/package.json
@@ -346,7 +346,7 @@
         },
         {
           "command": "java.project.updateSourceAttachment",
-          "when": "false == true"
+          "when": "false"
         }
       ]
     }
--- a/package.json
+++ b/package.json
@@ -294,7 +294,7 @@
       },
       {
         "command": "java.project.updateSourceAttachment",
-        "title": "Update Source Attachment",
+        "title": "Attach Source",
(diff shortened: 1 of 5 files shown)


testforstephen:
This PR will only support selecting a jar/zip file as new source, and use the default OS encoding to parse attachment. I feel like folder and encoding support is a low priority, and don't support selecting a folder and changing the attachment encoding format, how do you think?

fbricon:
I have a separate POC to set source encoding, I need to share. For now we'll keep it simple

package.json
```diff
@@ -333,6 +343,10 @@
         {
           "command": "java.edit.organizeImports",
           "when": "!editorReadonly && editorLangId == java"
+        },
+        {
+          "command": "java.project.updateSourceAttachment",
+          "when": "false == true"
```
fbricon: really?

PR #720, opened 2018-11-18
Signed-off-by:  Ayush Agrawal <ayushnita.ece@gmail.com>

AyushAgrawal104:
Add Snippet for Scanner in Java as it is one of most required one by new learners.

### DIFF
--- a/snippets/java.json
+++ b/snippets/java.json
@@ -181,5 +181,12 @@
 			"${1:Object} ${2:foo} = new ${1:Object}();"
 		],
 		"description": "Create new Object"
+	},
+	"Scanner": {
+		"prefix": "scanner",
+		"body": [
+			"Scanner ${1:var} = new Scanner(System.in);"
+		],
+		"description": "Scanner for reading input from System.in "
 	}
 }


yaohaizh:
I don't think is the scenario that common enough to add shared snippet. You can add personal snippet for such kind of purpose.

AyushAgrawal104:
I am also looking for the parametrized constructor and getter-setter any Idea about that

fbricon:
@AyushAgrawal104 as @yaohaizh this is a rather specific use case, that belongs to a [user defined snippet file](https://code.visualstudio.com/docs/editor/userdefinedsnippets). Note that you can also add project specific snippets, by creating a file with the `.code-snippets` extension under your project .vscode directory. See [this](https://github.com/fbricon/spring-boot-vue/blob/master/.vscode/project.code-snippets) for example.
But on top of that, the scanner snippet doesn't generate the required `import java.util.Scanner;` statement, thus inserting incomplete code. AFAIK, you can't generate multiline textedits from snippets, so this is not a good fit anyway.

For getters/setters, right now you can either generate them via code completion:
jul-24-2017 10-31-52

or through code actions on unused private fields. Unfortunately the code actions bits still need [some](597) [love](290), but we're gonna fix that eventually.

You need to be more specific about the parameterized constructor snippet.

Issue #618, opened 2018-08-27
Any way to get rid of .project .classpath .settings/ files?


Eskibear:
Every time I open a Java project, no matter Maven or Gradle, the extension always creates many `.project` `.settings/` `.classpath` files in my project folder. And yes I know it's Eclipse's behavior to create files in project root folder as mentioned in https://github.com/redhat-developer/vscode-java/issues/466#issuecomment-373391208  But it's really very annoying that the extension creates these files silently without any confirmation, especially with the SCM stuffs, I have to manually add them into `.gitignore` file. And I also have to add them into "files.exclude" in VSCode's settings to hide them from the explorer.

I've noticed some projects are also created in the "workspaceStorage" folder, so **is it possible to create a corresponding project in the "workspaceStorage" folder and link my files to it?** I followed [this post](https://stackoverflow.com/questions/3422261/store-eclipse-project-files-outside-the-project-directory) in Eclipse, and find it adds `<linkedResources>` entries in `.project`. I suppose it might be somehow doable.

Moreover, **it would be also nice for me if it notifies the changes and provides shortcuts to hide/ignore them**, because I don't know and don't care about `.settings/` stuffs. E.g. it might pop up a message box saying something like "Some files are created by the extension, do you want to add them into .gitignore?"

fbricon:
The problem is we're delegating the import part to m2e and buildship, for Maven and Gradle projects, respectively, which do generate the project metadata files directly under the project folders, so changing that behaviour will be extremely hard, if not impossible.

Adding them to .gitignore is dependent on users using Git for their, the Git integration on each client, so we'd need to be smart on how to send that sort of notification

Eskibear:
>  The problem is we're delegating the import part to m2e ...... so changing that behaviour will be extremely hard, if not impossible.

@fbricon Thanks for the clarification, and I realize it might involve the whole project management part. In fact as a user I don't care that much about where you put the metadata files, just don't let me know the existence of these files. It does confuse me when I'm using VSCode to write Java and it generates a bunch of eclipse-looked files there. And one more point is, am I supposed to modify these files manually? I suppose that's not suggested, and usually some strange errors occur wrongly when I make some mistakes in modifying those files. 

So I think if the extension manages these files and keep them invisible to me, it should help. It avoids confusion and potential errors caused by wrongly modifying/deleting metadata files. For me, a) hiding them from the explorer and b) add them into .gitignore both help a lot. 

[...]

PR #746, opened 2018-12-15
Recognize new maven, gradle, eclipse project(s) after first init

snjeza:
See https://github.com/eclipse/eclipse.jdt.ls/issues/144
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/901


jdneo:
Just an idea come up in my mind: how about prompt users to import if we detect the workspace contains valid project?

Take an example of the C# extension, when user opens a C# project folder, it will pop up a dialog to ask user if he want's to add required assets for build/debug:
Screen Shot 2020-03-13 at 9 49 35 AM

We can do the same thing, the Java Language Server can run in syntax mode by default, if it detects there is any build configuration files, then asks user to import it as a Java project. We can benefit from this experience by:

- Give users en expectation that importing might take some time to finish
- Won't generate `.project`, `.classpath` file by default, if the user just wants to view the code (#618)

snjeza:
> Just an idea come up in my mind: how about prompt users to import if we detect the workspace contains valid project?

See https://github.com/redhat-developer/vscode-java/pull/746/commits/f2aa7af29e74a42264fa98710935f310cdaa3640#diff-04c6e90faac2675aa89e2176d2eec7d8R132

fbricon:
It doesn't work: If  I have a multimodule Maven project, add a new module under the parent project, import doesn't find anything

fbricon:
The succession of merge commits is just a huge mess.
Please rebase against master to have a simple commit.

jdneo:
@fbricon The commits are rebased, could you please try again?

I tried both multi-module Maven/Gradle, both can detect new sub-module and generate `.classpath` and `.project`. The [test case](https://github.com/eclipse/eclipse.jdt.ls/blob/9f4bffbf81756d1c155d8633ed5b282890b92e0e/org.eclipse.jdt.ls.tests/src/org/eclipse/jdt/ls/core/internal/handlers/ImportNewProjectsTest.java) also imitates such scenario.

jdneo:
@fbricon If still not working, is it possible to share the project to let me diagnostic?

fbricon:
I opened an empty folder, then in vscode's terminal, I did `gradle init`, then tried to import the generated project:

Screen Shot 2020-05-18 at 6 55 54 PM

package.json
```diff
@@ -553,6 +553,11 @@
         "title": "Update project configuration",
         "category": "Java"
       },
+      {
+        "command": "java.project.import",
+        "title": "Import Java projects in workspacce",
```
fbricon, line 558: workspace

fbricon:
So this time, when calling the command, the extension activated, importing the project as part of the server startup, but ended up failing:  
Screen Shot 2020-05-19 at 3 00 12 PM

fbricon:
I keep calling the import new project command, now the server is started, but still see the "No delegateCommandHandler for java.project.import" error

fbricon:
Sorry,  looks like a PEBCAK. I didn't rebuild the proper server branch

PR #749, opened 2018-12-19
Enable Source Action 'Override/Implement Methods' with prompt

testforstephen:

It uses quick pick to list all possibles overridable methods from super types. And select the unimplemented methods by default.

override_implement

### DIFF
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -1,6 +1,6 @@
 'use strict';
 
-import { RequestType, NotificationType, TextDocumentIdentifier, ExecuteCommandParams } from 'vscode-languageclient';
+import { RequestType, NotificationType, TextDocumentIdentifier, ExecuteCommandParams, CodeActionParams, WorkspaceEdit } from 'vscode-languageclient';
 import { Command } from 'vscode';
 
 /**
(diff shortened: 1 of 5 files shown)


src/sourceAction.ts
```diff
@@ -0,0 +1,54 @@
...
+                label: `${method.name}(${method.parameters.join(',')})`,
+                description: `${method.declaringClassType}: ${method.declaringClass}`,
+                picked: method.unimplemented,
+                originalMethod: method,
+            };
+        });
+
+        const selectedItems = await window.showQuickPick(quickPickItems, {
+            canPickMany: true,
+            placeHolder: 'Select methods to override or implement.'
```
fbricon: ideally it should be `Select methods to override or implement in <Type>`, as when you're in an anonymous class, within a top level class, it's pretty confusing to know which type will receive the implemented methods.

src/protocol.ts
```diff
@@ -120,4 +120,26 @@ export interface SourceAttachmentAttribute {
...
+    key: string;
+    name: string;
+    parameters: string[];
+    unimplemented: boolean;
+    declaringClass: string;
+    declaringClassType: string;
+}
+
+export namespace OverridableMethodsRequest {
+    export const type = new RequestType<CodeActionParams, OverridableMethod[], void, void>('java/overridableMethods');
```
fbricon: Probably need a OverridableMethodsResponse object, that'd contain:
- type: String
- methods: OverridableMethod[]

that type can be displayed in the quickpick placeholder

testforstephen:
@fbricon @yaohaizh updated.

fbricon:
If you try to override/implement methods on an anonymous class, say Runnable:
screen shot 2019-02-11 at 2 46 58 pm


- dialog mentions top level class (should be Anonymous Runnable)
- no method from Runnable is selectable
- all methods are added to top level class, instead of the Anonymous one

testforstephen:
In VSCode 1.30.1 (2018 November), CodeActionParams passed the correct cursor line in the `CodeActionParams#range`, so the source action is able to resolve the right covered node. The case works.

After i updated to VSCode 1.31.0 (2019 January), i reproduced the scenario you mentioned. That's because the code action didn't pass the cursor line, `CodeActionParams#range` always return line 0. The source action cannot get the right covered node any more.

Not sure whether it's VSCode bug.

testforstephen:
Create an issue https://github.com/Microsoft/vscode/issues/68487

fbricon:
Since the upstream issue has been fixed, we'll push this feature in the next release.

PR #763, opened 2019-01-12
Changing "JDK 8" to just "JDK" in README

GabrielBB:
I Just found out JDK 11 is fully compatible so i used "java.home" with my JDK 11 folder and it works with no problems. That's why i'm doing this fix in the README


fbricon:
Thanks @GabrielBB

Issue #768, opened 2019-01-17
Autocompletion inactive after inserting a snippet in some cases


fbricon:
If you type `main` in a class, to insert the public static void main snippet, then typing `sys` next doesn't trigger autocompletion. This is caused by the main snippet using $1 as placeholder, instead of $0, which, according to the [spec](https://microsoft.github.io/language-server-protocol/specification#textDocument_completion), defines the final tab stop,

PR #769, opened 2019-01-17
Use -zsh a stop tab for client-side snippets

fbricon:
Partially addresses #768 (need to ensure jdt.ls also generates proper placholders)

### DIFF
--- a/snippets/java.json
+++ b/snippets/java.json
@@ -3,7 +3,7 @@
 		"prefix": "main",
 		"body": [
 			"public static void main(String[] args) {",
-			"\t$1",
+			"\t$0",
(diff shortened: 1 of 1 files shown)


snippets/java.json
```diff
@@ -178,7 +178,7 @@
 	"newObject": {
 		"prefix": "new",
 		"body": [
-			"${1:Object} ${2:foo} = new ${1:Object}();"
+			"${0:Object} ${1:foo} = new ${0:Object}();"
```
fbricon: this one I'm not too convinced. Would be happy to revert if someone slightly objects

PR #776, opened 2019-01-25
Hide java project setting files by default.

yaohaizh:
@fbricon 
Fix issue https://github.com/redhat-developer/vscode-java/issues/618 which gives us lots of negative feedback and requested by the users. 

Since this PR https://github.com/Microsoft/vscode/pull/66451 is closed by VSCode side, provide the vscode java setting to mitigate this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,10 +89,10 @@
           "description": "Specifies the severity of the message when the classpath is incomplete for a Java file",
           "scope": "window"
         },
-        "java.configuration.excludeProjectSettingsFiles": {
+        "java.configuration.checkProjectSettingsExclusions": {
           "type": "boolean",
           "default": true,
-          "description": "Exclude the extension generated project setting files from the file explorer. When set false, it will not revert the changes made on the setting.json file(s).",
+          "description": "Checks if the extension-generated project settings files (.project, .classpath, .factorypath, .settings/) should be excluded from the file explorer.",
           "scope": "window"
         },
         "java.configuration.updateBuildConfiguration": {
--- a/package.json
+++ b/package.json
@@ -89,10 +89,10 @@
           "description": "Specifies the severity of the message when the classpath is incomplete for a Java file",
           "scope": "window"
         },
-        "java.configuration.excludeProjectSettingFiles": {
+        "java.configuration.excludeProjectSettingsFiles": {
           "type": "boolean",
           "default": true,
-          "description": "Exclude the extension generated project setting files from file explorer.",
+          "description": "Exclude the extension generated project setting files from the file explorer. When set false, it will not revert the changes made on the setting.json file(s).",
           "scope": "window"
         },
         "java.configuration.updateBuildConfiguration": {
--- a/package.json
+++ b/package.json
@@ -89,6 +89,12 @@
           "description": "Specifies the severity of the message when the classpath is incomplete for a Java file",
           "scope": "window"
         },
+        "java.configuration.excludeProjectSettingFiles": {
+          "type": "boolean",
+          "default": true,
+          "description": "Exclude the extension generated project setting files from file explorer.",
+          "scope": "window"
+        },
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -89,6 +89,12 @@
           "description": "Specifies the severity of the message when the classpath is incomplete for a Java file",
           "scope": "window"
         },
+        "java.configuration.excludeProjectSettingFiles": {
+          "type": "boolean",
+          "default": true,
+          "description": "Exclude the extension generated project setting files from file explorer.",
```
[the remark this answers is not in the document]
fbricon: need to explain setting it to false won't remove the exclusions

fbricon:
```json
	"files.exclude": {
        "**/.git": true,
        "**/.svn": true,
        "**/.hg": true,
        "**/CVS": true,
        "**/.DS_Store": true,
        "**/.classpath": true,
        "**/.project": true,
        "**/.settings": true
    }
```
is added to the local settings.json, even if the java exclusions are already defined at the user settings level. That's a big no no

yaohaizh:
@fbricon  Addressed the comments.

src/setting.ts
```diff
@@ -0,0 +1,69 @@
+'use strict';
```
fbricon: settings.ts

fbricon:
you also need to hide `.factorypath`

fbricon:
if your settings contains
```
    "files.exclude": {
        "**/.classpath": true,
        "**/.settings": true
    }
```

agreeing to hide the project settings will change the settings to 
```
    "files.exclude": {
        "**/.project": true
    }
```

yaohaizh:
For setting generated, if user choose the workspace or global, the settings is split. Should we always keep them the same place. I would let the user to make the decision here.

src/settings.ts
```diff
@@ -0,0 +1,69 @@
...
+	const excudedConfig = getJavaConfiguration().get('configuration.excludeProjectSettingsFiles');
+	if (excudedConfig) {
+		const config = workspace.getConfiguration('files', workspaceUri);
+		const excludedValue: Object = config.get('exclude');
+		const needExcludeFiles: Object = {};
+
+		let needUpdate = false;
+		for (const hiddenFiles of DEFAULT_HIDDEN_FILES) {
+			if (!excludedValue.hasOwnProperty(hiddenFiles)) {
+				needExcludeFiles[hiddenFiles] = true;
```
fbricon: no, this will always store a subset of the files, overriding existing settings.
We do want to merge excludedValue + DEFAULT_HIDDEN_FILES in config.get('exclude')

src/settings.ts
```diff
@@ -0,0 +1,69 @@
...
+const DEFAULT_HIDDEN_FILES: string[] = ['**/.classpath', '**/.project', '**/.settings', '**/.factorypath'];
+
+let oldConfig: WorkspaceConfiguration = getJavaConfiguration();
+
+export function onConfigurationChange() {
+	return workspace.onDidChangeConfiguration(params => {
+		if (!params.affectsConfiguration('java')) {
+			return;
+		}
+		let newConfig = getJavaConfiguration();
```
fbricon: we want to check that if configuration.excludeProjectSettingsFiles is true, but it was false before, then excludeProjectSettingsFiles should be called.

Be aware that oldConfig is only changed if the "java configuration" changed, so it makes things a bit trickier.

yaohaizh:
> How about we change the choices to
> 
> * Exclude globally
> * Exclude in workspace
> * Never
> 
> If files exclusions exists in the workspace preferences, then we don't propose to "Exclude globally"

Would keep the term simple and consistent with other items.

fbricon:
I'm fine with having better labels, but they need to be explicit

yaohaizh:
@fbricon Updated PR.

Issue #788, opened 2019-02-07
Hieroglyphs in javadoc hover


spkoder:
Issue Type: <b>Bug</b>

Hello. I have a problem. I included another module as dependency to my maven java project. In this module I documented (javadoc) methods by russian symbols. Hints for this methods in this project are displayed as hieroglyphs. I think that extension decodes characters from sources of module using ANSI encoding (Win-1251), although the source itself is stored in UTF-8. Opening the same project in Idea, this problem is not observed. I tried to prescribe different settings, nothing helps.

Extension version: 0.38.0
VS Code version: Code 1.31.0 (7c66f58312b48ed8ca4e387ebd9ffe9605332caa, 2019-02-05T22:35:56.624Z)
OS version: Windows_NT x64 6.3.9600

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD FX(tm)-8350 Eight-Core Processor            (8 x 3991)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: enabled<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|7.96GB (2.98GB free)|
|Process Argv|C:\Users\hunter\Documents\!\Java\TestUseOtherModule|
|Screen Reader|no|
|VM|0%|

</details>

spkoder:
[TestLogic.zip](https://github.com/redhat-developer/vscode-java/files/2842651/TestLogic.zip)
[TestUseOtherModule.zip](https://github.com/redhat-developer/vscode-java/files/2842652/TestUseOtherModule.zip)
Source of module:
![default](https://user-images.githubusercontent.com/15929530/52442832-c7c0e480-2b56-11e9-8b0f-3c307af45aae.png)
Trying to see the hint for method of module in other project:
![2](https://user-images.githubusercontent.com/15929530/52442833-c7c0e480-2b56-11e9-817e-eaf53099eeb7.png)
When I press F12 (Go to definition):
![3](https://user-images.githubusercontent.com/15929530/52442834-c7c0e480-2b56-11e9-9b7c-37b50fb7e2fc.png)
[...]

PR #790, opened 2019-02-07
Start server with -Dfile.encoding=[vscode file encoding (UTF-8)] by default

fbricon:
Candidate fix for #788

### DIFF
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -4,6 +4,7 @@ import * as net from 'net';
 import * as glob from 'glob';
 import { StreamInfo, Executable, ExecutableOptions } from 'vscode-languageclient';
 import { RequirementsData } from './requirements';
+import { getJavaEncoding } from './settings';
 
 declare var v8debug;
 const DEBUG = (typeof v8debug === 'object') || startedInDebugMode();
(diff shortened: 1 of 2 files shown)


fbricon:
While that patch would fix #788 (and other similar reports in the past), by forcing the server to run with  UTF-8 instead of the platform's default encoding, I'm actually not sure about the potential drawbacks of that approach. Users can still set the -Dfile.encoding flag in jdt.ls.vmargs, but I don't know how many people would see a regression with the new default behavior.

tsmaeder:
In think the default should be the default from the machine where the source code lives. The language server cannot determine that, since it can be used in many different settings.

tsmaeder:
@fbricon what do you get when you ask for the default encoding from a Java program? Second, there are already provisions for overriding the default encoding in Eclipse. Maybe we should just surface them?

fbricon:
Funny thing is when I display Charset.defaultCharset() when the server start, I get US-ASCII. But when I run `locale` in a terminal I have:
```
LANG="en_CA.UTF-8"
LC_COLLATE="en_CA.UTF-8"
LC_CTYPE="en_CA.UTF-8"
LC_MESSAGES="en_CA.UTF-8"
LC_MONETARY="en_CA.UTF-8"
LC_NUMERIC="en_CA.UTF-8"
LC_TIME="en_CA.UTF-8"
LC_ALL=
```  
I have no idea why or what sets my jdt.ls to use US-ASCII. If I run a Java program from vscode, it returns UTF-8.
I like the idea of reusing vscode's default encoding.
As for exposing the server's encoding from props, I need to look into it (I also got an unfinished WIP on source encoding, but got stuck on the client side because I couldn't use a modifiable list).

fbricon:
Updated PR to send vscode's default encoding settings for Java (or global files.encoding -utf8 by default- if nothing is set)

Issue #597, opened 2018-07-26
Bulk generation would be useful


yaohaizh:
From the thumbups emoji  of this closed issue: 

https://github.com/redhat-developer/vscode-java/issues/100

> bulk generation like in jetbrains or eclipse would be pretty useful also.
> 
> And would be nice to have generators not only for getters and setters, but also for constructors, toString, hashCode, equals and overrides

fbricon:
See https://github.com/eclipse/eclipse.jdt.ls/issues/163

yaohaizh:
- `Add override/implement methods` code generation

fbricon:
Bulk generation of getters/setters is now available. The associated source action kind is "source.generate.accessors". 
Keybinding is [partially broken at the vscode level though](https://github.com/Microsoft/vscode/issues/64824).

But if/when it works, adding something like this to one's keybinding.json should do the trick
```
{
        "key": "shift+alt+g",
        "command": "editor.action.sourceAction",
        "args": {
          "kind": "source.generate",
        }
}
```

testforstephen:
Add `Generate Constructor` code generation.

yaohaizh:
Close this issue. The left one will be tracked on #972

xiuluo211314:
![image](https://user-images.githubusercontent.com/17724127/87239297-2c652a80-c440-11ea-86f1-0538d130217a.png)
open keybinding.json file, and add code like @fbricon  said， back to the editor , you can get it!

PR #814, opened 2019-02-27
Enable generate hashCode and equals source action

testforstephen:

Feature #597.

image

Prompt to select fields to include:
image

### DIFF
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -148,3 +148,31 @@ export interface AddOverridableMethodParams {
 export namespace AddOverridableMethodsRequest {
     export const type = new RequestType<AddOverridableMethodParams, WorkspaceEdit, void, void>('java/addOverridableMethods');
 }
+
+export interface VariableField {
+    bindingKey: string;
+    name: string;
+    type: string;
+}
+
+export interface CheckHashCodeEqualsResponse {
+    type: string;
+    fields: VariableField[];
+    foundHashCode: boolean;
+    foundEquals: boolean;
+    settings: any;
+}
+
+export namespace CheckHashCodeEqualsStatusRequest {
+    export const type = new RequestType<CodeActionParams, CheckHashCodeEqualsResponse, void, void>('java/checkHashCodeEqualsStatus');
+}
+
+export interface GenerateHashCodeEqualsParams {
+    context: CodeActionParams;
+    fields: VariableField[];
+    settings: any;
+}
+
+export namespace GenerateHashCodeEqualsRequest {
+    export const type = new RequestType<GenerateHashCodeEqualsParams, WorkspaceEdit, void, void>('java/generateHashCodeEquals');
+}
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -158,9 +158,7 @@ export interface VariableField {
 export interface CheckHashCodeEqualsResponse {
     type: string;
     fields: VariableField[];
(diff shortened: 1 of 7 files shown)


testforstephen:
Peer PR on eclipse.jdt.ls side eclipse/eclipse.jdt.ls#940

src/sourceAction.ts
```diff
@@ -52,3 +57,69 @@ export function registerCommands(languageClient: LanguageClient) {
...
+            if (result.foundHashCode) {
+                methods.push('\'int hashCode()\'');
+            }
+
+            if (result.foundEquals) {
+                methods.push('\'boolean equals(Object)\'');
+            }
+
+            const ans = await window.showInformationMessage(`Methods ${methods.join(' and ')} already exist in the Class '${result.type}'. `
+                + 'Do you want to regenerate the implementation?', 'YES', 'NO');
```
fbricon: 'Regenerate', 'Cancel'

src/sourceAction.ts
```diff
@@ -52,3 +57,69 @@ export function registerCommands(languageClient: LanguageClient) {
...
+            const methods = [];
+            if (result.foundHashCode) {
+                methods.push('\'int hashCode()\'');
+            }
+
+            if (result.foundEquals) {
+                methods.push('\'boolean equals(Object)\'');
+            }
+
+            const ans = await window.showInformationMessage(`Methods ${methods.join(' and ')} already exist in the Class '${result.type}'. `
```
fbricon: if methods.length == 1 => 'exists' else 'exist'

src/sourceAction.ts
```diff
@@ -52,3 +57,69 @@ export function registerCommands(languageClient: LanguageClient) {
...
+                + 'Do you want to regenerate the implementation?', 'YES', 'NO');
+            if (ans !== 'YES') {
+                return;
+            }
+
+            result.settings.regenerate = true;
+        }
+
+        if (result.settings && result.settings.useJ7HashEquals !== undefined) {
+            const template = await window.showQuickPick([
```
fbricon: I don't think we need to add a popup here. Turn it into a preference, using default template by default

src/protocol.ts
```diff
@@ -147,4 +147,32 @@ export interface AddOverridableMethodParams {
...
+export interface VariableField {
+    bindingKey: string;
+    name: string;
+    type: string;
+}
+
+export interface CheckHashCodeEqualsResponse {
+    type: string;
+    fields: VariableField[];
+    foundHashCode: boolean;
```
fbricon: replace 2 booleans with `existingMethods: string[]`
would contain `hashCode` and `equals`

fbricon:
Here are the new preferences controlling the code generation:

* `java.codeGeneration.hashCodeEquals.useJava7Objects`: Use Objects.hash and Objects.equals when generating the hashCode and equals methods. This setting only applies to Java 7 and higher. Defaults to `false`.
* `java.codeGeneration.hashCodeEquals.useInstanceof`: Use 'instanceof' to compare types when generating the hashCode and equals methods. Defaults to `false`.
* `java.codeGeneration.hashCodeEquals.useBlocks`: Use blocks in 'if' statements when generating the hashCode and equals methods. Defaults to `false`.
* `java.codeGeneration.hashCodeEquals.generateComments`: Generate method comments when generating the hashCode and equals methods. Defaults to `false`.

Issue #794, opened 2019-02-12
No Reload Adoption: Update `javaExtensions` contributions when extensions are added/removed


sandy081:
Hi, I am a developer from VS Code and in the last release `1.31`, we added a new feature - no reload is required on installing an extension - https://code.visualstudio.com/updates/v1_31#_no-reload-on-install

But there are some exceptions to it as mentioned in the doc. If the extension is contributing to external contribution points, then the extension requires reload after installing. Hence, we added a new listener API for the owners of the external contribution point to listen to extension addition/removal and can update its state - https://code.visualstudio.com/updates/v1_31#_extensions-change-event.

Our plan is to give extensions one milestone/month for adopting to the above API and make the other extensions consuming their contribution points ready for no reload. It means that from next release, above exception will be removed and all extensions are added to VS Code window immediately after installation and no reload is shown.

We scanned the extensions in the marketplace and found out that this extension uses an external contribution point `javaExtensions` and requires reload on install. I found out from code that this extension owns this contribution point and couple of popular extensions (vscjava.vscode-java-test, vscjava.vscode-java-debug) contributes to it which forces them to reload on install. So please adopt to the above API and make this and other contribution extensions ready for no reload.

Thanks

yaohaizh:
@sandy081 Thanks for bringing up this issue. We will adopt the API shortly.

yaohaizh:
@sandy081 
Something need further clarify: 

> Hence, we added a new listener API for the owners of the external contribution point to listen to extension addition/removal and can update its state - https://code.visualstudio.com/updates/v1_31#_extensions-change-event.

What's the state mean here? Could we change some internal specific state via vscode API inform VSCode to reload? Or it is required for extension to provide no reload behavior?

sandy081:
@yaohaizh You do not need to say any to VS Code. By state I mean following:

I think when your extension is activated, you will be accessing all extensions using `vscode.extensions.all` and their contributions to read `javaExtensions` contribution. You need to do this again when there is a change in `vscode.extensions.all` to accept new `javaExtensions` or remove existing `javaExtensions`. 

Let me know if it makes sense?

yaohaizh:
@sandy081 Thanks for the answer. 

@fbricon For the first step, we need check the javaExtensions and provide reload hints if the change require reload. 

In order to provide a fully no reload behavior, I logged an issue here: https://github.com/eclipse/eclipse.jdt.ls/issues/932

fbricon:
[...]

PR #822, opened 2019-03-04
Reload prompt when extension bundle changed.

yaohaizh:
@fbricon 
Fix for:  https://github.com/redhat-developer/vscode-java/issues/794

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -1,10 +1,13 @@
 'use strict';
 
-import { Extension } from 'vscode';
+import * as vscode from 'vscode';
 import * as path from 'path';
+import { Commands } from './commands';
 
-export function collectionJavaExtensions(extensions: Extension<any>[]): string[] {
-	let result = [];
+let existingExtensions: Array<string>;
+
+export function collectJavaExtensions(extensions: vscode.Extension<any>[]): string[] {
+	existingExtensions = [];
 	if (extensions && extensions.length) {
 		for (let extension of extensions) {
 			let contributesSection = extension.packageJSON['contributes'];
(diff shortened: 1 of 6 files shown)


yaohaizh:
@fbricon Could you help review this PR? Since the VSCode already removed the special handling for javaContributions.  We need handle this PR for the just released VSCode February update.

src/plugin.ts
```diff
@@ -1,22 +1,46 @@
...
+import * as vscode from 'vscode';
 import * as path from 'path';
+import { Commands } from './commands';
 
-export function collectionJavaExtensions(extensions: Extension<any>[]): string[] {
-	let result = [];
+let existingExtensions: Array<string>;
+
+export function collectJavaExtensions(extensions: vscode.Extension<any>[]): string[] {
+	existingExtensions = [];
```
fbricon: I'd rather prefer you keep using a local variable, like result before

src/plugin.ts
```diff
@@ -1,22 +1,46 @@
...
-	return result;
+	return existingExtensions;
+}
+
+export function onExtensionChange(extensions: vscode.Extension<any>[]) {
+	if (!existingExtensions) {
+		return;
+	}
+	const oldExtensions = existingExtensions.slice();
+	const newExtensions = collectJavaExtensions(extensions);
```
fbricon: Check if extensions size changed first. Then if the size is the same you can check if a new extension was added. Checking if an extension was removed becomes unnecessary

src/plugin.ts
```diff
@@ -1,22 +1,46 @@
...
+	for (const newExtension of newExtensions) {
+		if (oldExtensions.indexOf(newExtension) < 0) {
+			let msg = 'Java Language Server has new extension installed or updated, please restart VS Code to enable it.';
```
fbricon: move all that out of the loop since you need the same logic if extension size changed

src/plugin.ts
yaohaizh: @fbricon Fixed.

src/plugin.ts
yaohaizh: Not quite get the comment here. The code only check the newly added extension. The extension list could be changed with or without extension size changed.

yaohaizh:
@fbricon Updated the PR.

src/plugin.ts
fbricon: yes. But you need to check the size changed 1st. And then check for added extensions.

If you removed or added an extension, simply checking the size will tell you you need to reload.
If you don't check the size and remove an extension, your current check won't detect it.

This might not be a problem in VS Code per se, but the vscode-java extension is being loaded in Theia/Che, which behaves differently.

yaohaizh: Make sense. Updated the PR.

yaohaizh:
@fbricon  Updated the PR.

fbricon:
Tweaked the extension detection change logic and changed the prompt to:
Screen Shot 2019-03-14 at 12 37 13 PM

Thanks @yaohaizh

PR #839, opened 2019-03-14
Add status to ExtensionAPI

### LINKED ISSUE #830: Unable to distinguish Started/Error from extension.api after activation
In both "Started" and "Error" cases, the promises resolve the same object. If a 3rd-party extension calls
```ts
const api = await javaExtension.activate(); 
```
the results are the same. 

See below:
https://github.com/redhat-developer/vscode-java/blob/ca29cb2b22d2c2c5b7e44ae697c8d385607da6d7/src/extension.ts#L114-L135

Can we add a property in ExtensionAPI that indicating the status? E.g. something like
```ts
export interface ExtensionAPI {
    readonly apiVersion: string;
    readonly javaRequirement: RequirementsData;
    readonly status: "Started" | "Error"
}
```

fbricon:
Fixes #830

### DIFF
--- a/src/extension.api.ts
+++ b/src/extension.api.ts
@@ -2,5 +2,6 @@ import { RequirementsData } from './requirements';
 
 export interface ExtensionAPI {
     readonly apiVersion: string;
-    readonly javaRequirement: RequirementsData;
+	readonly javaRequirement: RequirementsData;
+	readonly status: "Started" | "Error";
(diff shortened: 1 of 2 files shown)


fbricon:
@Eskibear does it work for you?

Eskibear:
Thanks, it looks good to me.

Issue #316, opened 2017-10-05
Support new region markers


fbricon:
See https://code.visualstudio.com/updates/v1_17#_folding-regions

Java should probably support one (or all?) of those patterns:
*  `//#region` and `//#endregion` 
* `//region` and `//endregion`
* Netbeans format : `// <editor-fold desc="Description">` and `// </editor-fold>`

fbricon:
With #319 : 
![oct-05-2017 18-29-19](https://user-images.githubusercontent.com/148698/31255677-006e3eb4-a9fc-11e7-8c53-c8313311cc52.gif)

adds partial NetBeans support (NetBeans xml specific attributes are ignored)
![oct-05-2017 18-32-44](https://user-images.githubusercontent.com/148698/31255678-007350fc-a9fc-11e7-94c5-4e0faff4a887.gif)

PR #844, opened 2019-03-19
Add folding range setting.

yaohaizh:

### DIFF
--- a/package.json
+++ b/package.json
@@ -245,7 +245,7 @@
         "java.foldingRange.enabled": {
           "type": "boolean",
           "default": true,
-          "description": "Enable/disable java folding range support. If disabled, it will use the default indentation folding range provided by VSCode.",
+          "description": "Enable/disable smart folding range support. If disabled, it will use the default indentation-based folding range provided by VS Code.",
           "scope": "window"
         },
         "java.progressReports.enabled": {
--- a/package.json
+++ b/package.json
@@ -242,6 +242,12 @@
           ],
           "scope": "window"
         },
+        "java.foldingRange.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable java folding range support. If disabled, it will use the default indentation folding range provided by VSCode.",
+          "scope": "window"
+        },
         "java.progressReports.enabled": {
           "type": "boolean",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -242,6 +242,12 @@
           ],
           "scope": "window"
         },
+        "java.foldingRange.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable java folding range support. If disabled, it will use the default indentation folding range provided by VSCode.",
```
fbricon: Enable/disable smart folding range support. If disabled, it will use the default indentation-based folding range provided by VS Code.

yaohaizh: @fbricon Updated

fbricon:
Before I merge it, I'd like to see https://github.com/eclipse/eclipse.jdt.ls/issues/694#issuecomment-474987887 being addressed, as we're currently losing the custom region folding support (See #316) with the new implementation.

fbricon:
Something's strange: if you add a snippet of code (new method body), then that method is not foldable. You need to save/close/reopen the file to see the new collapsible blocks

fbricon:
So folding the try block doesn't look good (it hides the catch() bit, but leaves its content) and the switch blocks no longer allow for fined grained folding:
after

as opposed to the behavior with the current client-side folding:
before

yaohaizh:
Fixed the test. 

For the try/catch, the same behavior as the if/else statement. Eclipse don't provide such grained folding yet. The intelliJ provides the similar behavior as the Java folding range from JDT.

If we provide both, we need drill down both indentation-based rules & AST based rules in some cases, which make things quite complex, a P3 issue?

fbricon:
Intellij provides the same folding behavior for try/catch/finally/switch as in the current indentation-based folding in vscode-java. The only difference is individual cases in a switch block don't collapse.

Folding the try block is really bad with the new PR. I was thinking parsing regions like you do with comments, but using ITerminalSymbols.[TokenNametry, TokenNamecatch, TokenNameswitch] instead, might be doable. So if we could fix it now, I'd be happier.

yaohaizh:
https://github.com/eclipse/eclipse.jdt.ls/pull/968
@fbricon  Updated the PR.

fbricon:
@yaohaizh that's beautiful ;-) Thanks!!

PR #873, opened 2019-04-09
Add Code Action: Generate toString()...

testforstephen:

User setting to configure generate toString template:
image

QuickPick to select the fields to include in the toString method.
image


testforstephen:
Peer PR on jdt.ls side. https://github.com/eclipse/eclipse.jdt.ls/pull/988

dmarrazzo:
Amazing! Thank you very much. Can you give some information on the template building rules?
For example, I'd like to add the object package name.

testforstephen:
@dmarrazzo Here are the supported template [variables](https://help.eclipse.org/2019-03/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Freference%2Fref-tostring-templates.htm). 


${object.className} | inserts the class name as a simple String
-- | --
${object.getClassName} | inserts a call to this.getClass.getName()
${object.superToString} | inserts a call to super.toString()
${object.hashCode} | inserts a call to this.hashCode()
${object.identityHashCode} | inserts a call to System.identityHashCode(this)
${member.name} | inserts the first member's name
${member.name()} | inserts the first member's name followed by parenthesis in case of methods
${member.value} | inserts the first member's value
${otherMembers} | inserts the remaining members. For each member, the template fragment between the first and the last ${member.*} variable is evaluated and appended to the result. The characters between the last ${member.*} and ${otherMembers} define the separator that is inserted between members (${otherMembers} must stand after the last ${member.*} variable).

PR #880, opened 2019-04-12
Fix build on Windows

### LINKED ISSUE #793: Windows 10 x64/macOS, open spring-boot/wildfly, it will always trigger full build after initialized after imported

##### Environment

- Version: 1.31.0 (user setup)
- Commit: 7c66f58312b48ed8ca4e387ebd9ffe9605332caa
- Date: 2019-02-05T22:35:56.624Z
- Electron: 3.1.2
- Chrome: 66.0.3359.181
- Node.js: 10.2.0
- V8: 6.6.346.32
- OS: Windows_NT x64 10.0.17763

##### Steps To Reproduce
1. Clone https://github.com/spring-projects/spring-boot or https://github.com/wildfly/wildfly
2. Open the repo in vscode
3. Close the repo after the first-time initialization is done(full built is done for the first time)
4. Reopen the repo in vscode

##### Current Result
A full build will be triggered after some times

##### Expected Result
No full build should be triggered since there is no change made on the contents

##### Additional Informations
The issue only exists on the Windows platform, but not macOS, Linux
Might also exists for  small/medium size projects, but it takes less time to do the building.

snjeza:
Fixes #793 


src/javaServerStarter.ts
```diff
@@ -66,12 +67,6 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
 	if (vmargs.indexOf(encodingKey) < 0) {
 		params.push(encodingKey + getJavaEncoding());
 	}
-	if (os.platform() == 'win32') {
```
fbricon: why is this removed?

fbricon:
Restarting spring-boot workspace, I see the shutdown doesn't complete, and the workspace is still not properly saved:
```
!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 15:32:09.126
!MESSAGE >> shutdown

!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 15:32:09.127
!MESSAGE Saving workspace
!SESSION 2019-04-12 15:32:14.400 -----------------------------------------------
eclipse.buildId=unknown
java.version=1.8.0_181
java.vendor=Oracle Corporation
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=en_CA
Command-line arguments:  -data /Users/fbricon/Library/Application Support/Code - Insiders/User/workspaceStorage/4b108579b42a00cd8343838a6a610090/redhat.java/jdt_ws

!ENTRY org.eclipse.core.resources 2 10035 2019-04-12 15:32:15.906
!MESSAGE The workspace exited with unsaved changes in the previous session; refreshing workspace to recover changes.
```

src/javaServerStarter.ts
snjeza: VS Code doesn't kill Java LS server anymore. See https://github.com/Microsoft/vscode/issues/35196 and https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/main.ts#L418

snjeza:
> Restarting spring-boot workspace, I see the shutdown doesn't complete, and the workspace is still not properly saved:

Could you test with https://raw.githubusercontent.com/snjeza/vscode-test/master/java-0.43.0.vsix ?

fbricon:
This is on another workspace, but this looks much better, shutdown takes several seconds to complete, but does complete:
```

!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 16:25:34.321
!MESSAGE >> shutdown

!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 16:25:34.321
!MESSAGE Saving workspace

!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 16:25:37.599
!MESSAGE Workspace saved. Took 3277ms

!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 16:25:37.600
!MESSAGE Shutdown received... waking up main thread

!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 16:25:37.605
!MESSAGE class org.eclipse.jdt.ls.core.internal.JavaLanguageServerPlugin is stopping:
!SESSION 2019-04-12 16:25:39.447 -----------------------------------------------
eclipse.buildId=unknown
java.version=1.8.0_181
java.vendor=Oracle Corporation
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=en_CA
Command-line arguments:  -data /Users/fbricon/Library/Application Support/Code/User/workspaceStorage/4b108579b42a00cd8343838a6a610090/redhat.java/jdt_ws

!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 16:25:45.133
!MESSAGE class org.eclipse.jdt.ls.core.internal.JavaLanguageServerPlugin is started

!ENTRY org.eclipse.jdt.ls.core 1 0 2019-04-12 16:25:45.261
!MESSAGE Main thread is waiting
```

src/javaServerStarter.ts
```diff
@@ -15,6 +15,13 @@ export function prepareExecutable(requirements: RequirementsData, workspacePath,
 	const options: ExecutableOptions = Object.create(null);
 	options.env = process.env;
 	options.stdio = 'pipe';
+	if (os.platform() === 'win32') {
+		const vmargs = javaConfig.get('jdt.ls.vmargs', '');
+		const watchParentProcess = '-DwatchParentProcess=';
+		if (vmargs.indexOf(watchParentProcess) >= 0) {
```
fbricon: watchParentProcess is always present, as per L78

snjeza: This code is called before the prepareParams method (L78).

yaohaizh:
@fbricon  @snjeza  Could this change https://github.com/Microsoft/vscode-languageserver-node/issues/485 fix the parent watching process issue?

snjeza:
I have updated the PR.

src/javaServerStarter.ts
rgrunber: @snjeza , why do we need to set `detatched` to `true` only in Windows ? Does https://github.com/microsoft/vscode-languageserver-node/blob/e7b7e3134791c4208165aa57fb9182895df8ec6c/client/src/node/main.ts#L201 just work on Linux/MacOS ?

snjeza: Yes, it just works. I think it is the default on Linux/Mac.

jdneo:
@snjeza I tested this change on Windows and observed that after reloading the window, the message `!MESSAGE The workspace exited with unsaved changes in the previous session; refreshing workspace to recover changes.` **will not** be seen anymore.

One question: I think the key point is `options.detached = true;`, but according to current change, it will only be set when vmargs contains `-DwatchParentProcess=`, which is not contained in the default value of the vmargs. What's the purpose for that check?

snjeza:
> One question: I think the key point is options.detached = true;, but according to current change, it will only be set when vmargs contains -DwatchParentProcess=, which is not contained in the default value of the vmargs. What's the purpose for that check?

We could set it as the default. It will not run tasklist (on Windows) or kill (on Linux/Mac)t after https://github.com/eclipse/eclipse.jdt.ls/pull/2115
This way, Java LS will be shut down on Windows properly.
@fbricon @rgrunber @testforstephen @CsCherrYY

rgrunber:
@jdneo How does this change solve the issue on Windows, with only defaults if `-DwatchParentProcess=false` is never passed as a JVM argument. It only gets set to `false` in `prepareParams`, which happens after the change in this PR. Did you manually set `-DwatchParentProcess=false` in `java.jdt.ls.vmargs` ?

snjeza:
@rgrunber you should set `-DwatchParentProcess=30`, 30 seconds delay

rgrunber:
If this helps the situation, then I'm fine with merging. I just don't see how it would benefit the issue without users also overriding the default `java.jdt.ls.vmargs` in their settings. But if a manual change to a setting is required then most users likely won't benefit.

snjeza:
> If this helps the situation, then I'm fine with merging. I just don't see how it would benefit the issue without users also overriding the default java.jdt.ls.vmargs in their settings. But if a manual change to a setting is required then most users likely won't benefit.

See https://github.com/redhat-developer/vscode-java/pull/880#issuecomment-1260968971

jdneo:
When I tested this change, I manually changed the vmargs. `options.detached = true;` seems to work. I'm also confused the check for the `-DwatchParentProcess`.

@snjeza Could you help me understand the relationship between `-DwatchParentProcess` and `options.detached = true;`?

jdneo (approved):
I tested on the Windows machine.

Opening eclipse.jdt.ls project:

- Without the change, takes around 60s
- With the change, it takes around 30s.

PR #907, opened 2019-04-29
Enable advanced "Generate getters and setters..." source action

testforstephen:

Enable the advanced "Generate getters and setters..." source action if the class has more than one field. It will prompt a quick pick box to select the target fields for generating the accessor methods.
image

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -126,4 +126,8 @@ export namespace Commands {
      * Generate toString().
      */
     export const GENERATE_TOSTRING_PROMPT = 'java.action.generateToStringPrompt';
+    /**
+     * Generate Getters and Setters.
+     */
+    export const GENERATE_ACCESSORS_PROMPT = 'java.action.generateAccessorsPrompt';
 }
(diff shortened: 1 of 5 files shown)


fbricon:
So if you have 2 fields without accessors, the quickfix to generate them has an ellipsis, yet no wizard is shown as you'd expect from the ellipsis.

When multiple fields are available for the source action, they should all be selected by default

testforstephen:
Create a bug https://github.com/redhat-developer/vscode-java/issues/908 to track the issue for the quick fix.  Would like to fix it in a new PR.

About the suggestion to select all fields by default, i may have a different opinion. I would prefer to leave it unselected by default, this is the same behavior as the Eclipse and Intellij. Because the quick pick widget will also list the static fields, which the user may not want to generate the accessors in most of cases.

fbricon:
Merged as 6686996. I just prefixed the `static` keyword when needed, instead of enumerating it:
Screen Shot 2019-04-30 at 12 36 05 PM

PR #917, opened 2019-05-08
Optionally disable loading gradle from gradle wrapper

### LINKED ISSUE #875: Optionally disable loading gradle from gradle wrapper (request)
It would be nice to be able to either optionally disable loading gradle from the gradle wrapper and always using system gradle (if higher then 2.6) so in cases where the gradle project is using a wrapper with a version < 2.6 the java plugin can still function.

##### Environment
- Operating System: Mac OSX High Sierra
- JDK version: 1.8
- Visual Studio Code version: 1.32.3
- Java extension version: 0.42.1

##### Steps To Reproduce
1. Have gradle > 2.6 on path
2. Have a gradle project that uses the wrapper with gradle < 2.6 (ex: 1.12)
3. Redhat plugin complains about not being able to use the plugin with gradle < 2.6

##### Current Result
Can't use downstream plugins (checkstyle/run tests)

##### Wanted Result
Redhat plugin can optionally use system gradle always to at least allow some functionality on gradle 1.12 projects that use the wrapper

##### Additional Informations
I found this out while initially developing a project where it did not have the gradle wrapper with my gradle 5 system gradle version. I was able to run/debug tests and use checkstyle with no problems. I then added the wrapper, to find that afterwards i could not do either of these things because my gradle version (in the wrapper) was too low.

snjeza:
Fixes #875 
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/1026

I have added two new properties: `java.import.gradle.wrapper.enabled` and `java.import.gradle.version`

Importing a gradle project works in the following way:

- if there is gradlew and java.gradle.wrapper.enabled=true (default), the gradle wrapper distribution will be used
- in the case that gradlew doesn't exist or java.import.gradle.wrapper.enabled=false,  Java LS checks for the existence of the java.import.gradle.version property. If it exists,  Java LS will use the required version
- if there's no java.import.gradle.version property, Java LS checks for existence of the GRADLE_HOME env variable or the GRADLE_HOME system property. If it exists,  Java LS will use the GRADLE_HOME local installation,
-  else Java LS will use default buildship version (5.0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,10 @@ The following settings are supported:
 * `java.codeGeneration.toString.listArrayContents`: List contents of arrays instead of using native toString(). Defaults to `true`.
 * `java.codeGeneration.toString.limitElements`: Limit number of items in arrays/collections/maps to list, if 0 then list all. Defaults to `0`.
 
+*New in 0.45.0:*
+* `java.import.gradle.wrapper.enabled`: Enable/disable the Gradle wrapper.
+* `java.import.gradle.version`: Gradle version, used if the gradle wrapper is missing or disabled.
+
 Troubleshooting
 ===============
 1. Check the status of the language tools on the lower right corner (marked with A on image below).
--- a/package.json
+++ b/package.json
@@ -114,16 +114,28 @@
           "description": "Traces the communication between VS Code and the Java language server.",
           "scope": "window"
         },
+        "java.import.maven.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable the Maven importer.",
+          "scope": "window"
+        },
         "java.import.gradle.enabled": {
           "type": "boolean",
           "default": true,
           "description": "Enable/disable the Gradle importer.",
           "scope": "window"
         },
-        "java.import.maven.enabled": {
+        "java.import.gradle.wrapper.enabled": {
           "type": "boolean",
           "default": true,
-          "description": "Enable/disable the Maven importer.",
+          "description": "Enable/disable the Gradle wrapper.",
+          "scope": "window"
+        },
+        "java.import.gradle.version": {
+          "type": "string",
+          "default": null,
+          "description": "Gradle version, used if the gradle wrapper is missing or disabled.",
           "scope": "window"
         },
         "java.maven.downloadSources": {


package.json
```diff
@@ -114,16 +114,22 @@
           "description": "Traces the communication between VS Code and the Java language server.",
           "scope": "window"
         },
-        "java.import.gradle.enabled": {
```
fbricon: I want to keep java.import.gradle.enabled

package.json
```diff
@@ -114,16 +114,22 @@
...
+        "java.import.maven.enabled": {
           "type": "boolean",
           "default": true,
-          "description": "Enable/disable the Gradle importer.",
+          "description": "Enable/disable the Maven importer.",
           "scope": "window"
         },
-        "java.import.maven.enabled": {
+        "java.gradle.wrapper.enabled": {
```
fbricon: java.import.gradle.wrapper.enabled

package.json
```diff
@@ -114,16 +114,22 @@
...
           "type": "boolean",
           "default": true,
-          "description": "Enable/disable the Maven importer.",
+          "description": "Enable/disable the Gradle wrapper.",
+          "scope": "window"
+        },
+        "java.gradle.version": {
```
fbricon: java.import.gradle.version

fbricon:
I think it makes more sense to check
- use wrapper if enabled and present
- else use gradle version if it is set and valid
- else fall back to gradle home
- else fall back to buildship default version

snjeza:

package.json
```diff
@@ -114,16 +114,28 @@
...
           "type": "boolean",
           "default": true,
-          "description": "Enable/disable the Maven importer.",
+          "description": "Enable/disable the Gradle wrapper.",
+          "scope": "window"
+        },
+        "java.import.gradle.version": {
+          "type": "string",
+          "default": null,
+          "description": "Gradle version.",
```
fbricon: Gradle version, used if the gradle wrapper is missing or disabled.

README.md
```diff
@@ -109,6 +109,10 @@ The following settings are supported:
 * `java.codeGeneration.toString.listArrayContents`: List contents of arrays instead of using native toString(). Defaults to `true`.
 * `java.codeGeneration.toString.limitElements`: Limit number of items in arrays/collections/maps to list, if 0 then list all. Defaults to `0`.
 
+*New in 0.45.0:*
+* `java.import.gradle.wrapper.enabled`: Enable/disable the Gradle wrapper.
+* `java.import.gradle.version`: Gradle version.
```
fbricon: Gradle version, used if the gradle wrapper is missing or disabled.

snjeza:

PR #921, opened 2019-05-14
Enable Source Action: Generate Constructors

testforstephen:

generateConstructors


fbricon:
Merged as https://github.com/redhat-developer/vscode-java/commit/e93092ca61cd3e96228b684ad29d7464a47465d7, after fixing the labels and the command name on the server side.

PR #930, opened 2019-05-17
Enable code action: Generate Delegate Methods...

testforstephen:

delegateMethods


testforstephen:
Peer server side PR: https://github.com/eclipse/eclipse.jdt.ls/pull/1046

PR #942, opened 2019-05-30
Enable proposed API.

yaohaizh:
Enable proposed API since the call hierarchy & selectionRange is in insider version yet Java LS is mostly feature ready for these features. Would like to bring these features to the customers ASAP.


yaohaizh:
@fbricon Updated PR

fbricon:
Patch reverted due to the Marketplace preventing extensions using that flag from publishing. See c945683

PR #987, opened 2019-07-17
Log file to record client errors

akaroml:
Signed-off-by: Rome Li <rome.li@microsoft.com>
Fix #985

PR #988, opened 2019-07-17
Add 'Open Java extension log file' command

fbricon:
Supersedes #987


akaroml:
Here's the improvement I mentioned in #987. The new command shows `client.log` will work fine in most of the cases, except when the rotation happens to deal with a larger log volume. When the 100k size limit is hit, extra log files will be created, so there could be `client#.log` files. I was thinking about how those files could be presented in the most helpful way. A more complicated situation is when rotation happens in an older session (so we have `client7.log` and `client8.log`) and a new session tries to start from `client.log`...

An alternative is to use the daily rotation instead. In this way, the rotation only happens along with the date change, and we would just present the latest one. I'll try and possibly have a PR for that.

PR #1015, opened 2019-08-02
Log language client failures

akaroml:
Signed-off-by: Rome Li <rome.li@microsoft.com>

This is to record language client errors in the client log file.

PR #1021, opened 2019-08-05
Open client/server logs SxS

### LINKED ISSUE #1016: Show client & server logs side by side
Now we have both client and server logs. It would be great to have an entry to show both of them side by side for easier troubleshooting.

The same entry could also be the follow-up action whenever a critical issue occurs.

akaroml:
Fix #1016

### DIFF
--- a/package.json
+++ b/package.json
@@ -399,6 +399,11 @@
         "category": "Java"
       },
       {
+        "command": "java.open.logs",
+        "title": "Open all log files",
+        "category": "Java"
+      },
+      {
         "command": "java.workspace.compile",
         "title": "Force Java compilation",
         "category": "Java"
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -80,6 +80,11 @@ export namespace Commands {
     export const OPEN_CLIENT_LOG = 'java.open.clientLog';
 
     /**
+    * Open Java log files side by side
+    */
+   export const OPEN_LOGS = 'java.open.logs';
+
+    /**
      * Open Java formatter settings
      */
     export const OPEN_FORMATTER = 'java.open.formatter.settings';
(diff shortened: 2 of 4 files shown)


akaroml:
Since #1015 is merged, let me add the button to show logs when crash.

fbricon:
I noticed that if you crash the jvm on purpose (add `-crash.me` to jdt.ls.vmargs), then the error is not logged in either log file, but is visible in the output view of the language server:
```
Unrecognized option: -crash.me
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
[Error - 3:48:57 PM] Connection to server got closed. Server will not be restarted.

```

akaroml:
> I noticed that if you crash the jvm on purpose (add `-crash.me` to jdt.ls.vmargs), then the error is not logged in either log file, but is visible in the output view of the language server:
> 
> ```
> Unrecognized option: -crash.me
> Error: Could not create the Java Virtual Machine.
> Error: A fatal exception has occurred. Program will exit.
> [Error - 3:48:57 PM] Connection to server got closed. Server will not be restarted.
> ```

I think the language client is still using `console.log` in some cases, see:
https://github.com/microsoft/vscode-languageserver-node/blob/master/client/src/client.ts#L155

At the same time, it does not report such errors to the error handler.

I'll take a further look tomorrow but I guess that's the best we can do for now.

akaroml:
Check this out:
https://github.com/microsoft/vscode-languageserver-node/blob/db16b7f53b318e8ff44aa95ac8faa5e2486b63aa/client/src/main.ts#L405

The root cause is a bit different than I thought but the idea is similar - language client is not reporting everything back. In this case, the client is trying to start a child process and pipe stdout and stderr to the output channel, which is provided by vscode by default.

We do have the option to pass a custom output channel through `LanguageClientOptions.outputChannel`, but the limitation is that we could not isolate errors from verbose info.

#1022 was opened for it. Please help assign that to me.

akaroml:
#1024 was created to collect the logs in the output channel.

akaroml:
ping...

fbricon:
Thanks @akaroml !

PR #1024, opened 2019-08-07
Log info from output channel

### LINKED ISSUE #1022: Collect language client logs from output channel
https://github.com/microsoft/vscode-languageserver-node/blob/db16b7f53b318e8ff44aa95ac8faa5e2486b63aa/client/src/main.ts#L405

The language client is using the default output channel to send diagnostic info. This type of info is also helpful for troubleshooting. This issue is to also capture that info and store in the client log file.

akaroml:
Fix #1022

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,7 +3,7 @@
 import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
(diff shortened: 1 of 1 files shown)


akaroml:
ping...

fbricon:
Thanks @akaroml ! Keep'em coming!

PR #1027, opened 2019-08-08
Enable move instance method and static member

testforstephen:
requires https://github.com/eclipse/eclipse.jdt.ls/pull/1145


fbricon:
Thanks @testforstephen !

PR #1041, opened 2019-08-23
Enable to move type to new file

testforstephen:

It requires https://github.com/eclipse/eclipse.jdt.ls/pull/1151


fbricon:
Thanks @testforstephen !

PR #1059, opened 2019-09-14
Exclude certain packages from autocomplete/autoimport

### LINKED ISSUE #710: Exclude certain packages from autocomplete/autoimport
Autoimport does not work when you have packages that have the same class name. 

For example, when typing `List<String> a = new List<String>()`, there will be errors because it doesn't know whether to resolve against `java.util.List` or `java.awt.List`. 

If I know in my project I never use the `java.awt` package, can we add a setting to ignore that from the autocomplete/autoimport feature?

snjeza:
Fixes #710
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/1176

### DIFF
--- a/package.json
+++ b/package.json
@@ -263,6 +263,15 @@
           ],
           "scope": "window"
         },
+        "java.completion.filteredTypes": {
+          "type": "array",
+          "description": "Defines the type filters. All types whose fully qualified name matches the selected filter strings will be ignored in content assist or quick fix proposals and when organizing imports. For example 'java.awt.*' will hide all types from the awt packages.",
+          "default": [
+            "java.awt.*",
+            "com.sun.*"
+          ],
+          "scope": "window"
+        },
         "java.completion.importOrder": {
           "type": "array",
           "description": "Defines the sorting order of import statements. A package or type name prefix (e.g. 'org.eclipse') is a valid entry. An import is always added to the most specific group.",
--- a/README.md
+++ b/README.md
@@ -110,13 +110,14 @@ The following settings are supported:
 * `java.codeGeneration.toString.skipNullValues`: Skip null values when generating the toString method. Defaults to `false`.
 * `java.codeGeneration.toString.listArrayContents`: List contents of arrays instead of using native toString(). Defaults to `true`.
 * `java.codeGeneration.toString.limitElements`: Limit number of items in arrays/collections/maps to list, if 0 then list all. Defaults to `0`.
-
-*New in 0.47.0*
-
 * `java.import.gradle.arguments`: Arguments to pass to Gradle.
 * `java.import.gradle.jvmArguments`: JVM arguments to pass to Gradle.
 * `java.import.gradle.home`: setting for GRADLE_HOME.
 * `java.selectionRange.enabled`: Enable/disable Smart Selection support for Java. Disabling this option will not affect the VS Code built-in word-based and bracket-based smart selection.
+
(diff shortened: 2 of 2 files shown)


snjeza:

package.json
```diff
@@ -263,6 +263,15 @@
           ],
           "scope": "window"
         },
+        "java.completion.filteredTypes": {
+          "type": "array",
+          "description": "Defines the type filters. All types whose fully qualified name matches the selected filter strings will be ignored in content assist or quick fix proposals and when organizing imports. For example 'java.awt.*' will hide all types from the awt packages.",
+          "default": [
+            "java.awt.*",
```
maxandersen: what is the default list added by eclipse jdt ?

package.json
```diff
@@ -263,6 +263,15 @@
           ],
           "scope": "window"
         },
+        "java.completion.filteredTypes": {
```
maxandersen: it would be great if this list of import type filters would be possible to read from the project in a way we avoid to duplicate the list of filters between vscode, eclipse, che and more ide's. Could we add a file it would read ? like a `.javafiltertypes` all three IDE's would honor ? Maybe there exist something already for other IDE we could reuse?

package.json
snjeza: empty list

package.json
snjeza: VS Code (and che) use <project_home>/.vscode/settings.json ( Workspace settings).

maxandersen: yes, but that is not present in other IDE's - does intellij have a mechanism we could reuse or ?

package.json
maxandersen: I don't think that is true - maybe in Eclipse JDT raw, but when you run default Eclipse IDE download I believe they preconfigure the list, right ?

snjeza: I have checked Eclipse I20190830-0550 (eclipse 4.13) and Eclipse 2019-06 (4.12)

maxandersen: ack - you are right. Just something I probably just added as a reflex in the past ;)

maxandersen:
These settings are global for JDT - will they behave globally for jdt.ls too ? What we really  want in the end is the ability to have global filters (like java.awt.*) but then per project say - in this project I don't care for (x.y.z.Resource* imports/content assist)

Is that somehow doable ? basically dynamically set the type filters dependent on in which project the "operation" occurs ?

snjeza:
> Is that somehow doable ? basically dynamically set the type filters dependent on in which project the "operation" occurs ?

The org.eclipse.jdt.ui.typefilter.enabled property can't be set at the project level in Eclipse. See org.eclipse.jdt.core.manipulation.TypeNameMatchCollector.getPreference(String)

package.json
fbricon: right now will start with the low hanging fruit of exposing the list of filtered types directly in the server settings.

maxandersen:
> The org.eclipse.jdt.ui.typefilter.enabled property can't be set at the project level in Eclipse. See org.eclipse.jdt.core.manipulation.TypeNameMatchCollector.getPreference(String)

no, but could we "cheat" by setting it before doing import/content assist action and afterwards roll it back ? :) cheeky but might work.

snjeza:
> no, but could we "cheat" by setting it before doing import/content assist action and afterwards roll it back ? :) cheeky but might work.

See https://github.com/redhat-developer/vscode-java/pull/1059#discussion_r324438594

PR #1099, opened 2019-10-16
Enable autoimport on paste for non-empty files

### LINKED ISSUE #1098: Auto-import was not triggered automatically in some cases
#### Environment
- Operating System: Win/Mac/Linux
- JDK version: 13/12/8
- Visual Studio Code version:1.39.2
- Java extension version:0.51.0

#### Steps To Reproduce
1. Make sure java.actionsOnPaste.organizeImports in settings is true.
2. New Foo.java with codes as below: 
```
import java.io.File;
 
public class Foo {
 
        File f =new File("A.txt");
    }
}
```

#### Result:
1. If copy the main method to another empty java file
Auto-import was not triggered automatically.
image

2. If copy the all class to another empty java file
Auto-import was triggered automatically.
image

fbricon:
Fixes #1098

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -155,9 +155,7 @@ export function registerOverridePasteCommand(languageClient: LanguageClient, con
 
         const clipboardText: string = await env.clipboard.readText();
         const editor: TextEditor = window.activeTextEditor;
-
         const documentText: string = editor.document.getText();
-
         const isCursorOnly = editor.selection.isEmpty;
         let action: Thenable<boolean>;
         if (isCursorOnly) {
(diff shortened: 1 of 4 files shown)


src/settings.ts
```diff
@@ -174,9 +171,8 @@ export function registerOverridePasteCommand(languageClient: LanguageClient, con
         }
 
         action.then((wasApplied) => {
-            const hasText: boolean = documentText !== null && /\S/.test(documentText);
             const fileURI = editor.document.uri.toString();
-            if (wasApplied && !hasText && fileURI.endsWith(".java")) { // only organizing imports if document is blank
+            if (wasApplied && fileURI.endsWith(".java")) { // only organizing imports if document is blank
```
testforstephen: Remove the unnecessary comment `// only organizing imports if document is blank`

fbricon: done

testforstephen:
I found paste will trigger advanced organizedImports which will pop up the quick picks. I'm thinking whether it's more friendly to invoke the silent organizeImport version (Similar as java.saveActions.organizeImport). The silent version is invoked by delegate command `java.edit.organizeImports`.

fbricon:
We can do silent action when pasting in a non empty file, and interactive organize import when pasting in an empty one, but I'm afraid the discrepancy might lead to some confusion

fbricon:
I'm not seeing java.edit.organizeImports anywhere

testforstephen:
This is the old command, the eclipse.jdt.ls has it. Maybe the client removed the key constant. See https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/JDTDelegateCommandHandler.java#L31

testforstephen:
If we have knowledge to get the copy source, then we can leverage the source to resolve the ambiguous imports. If the source is unknown, then use the silent organizeImports. This can be a future experiment (not urgent, just nice to have).

fbricon:
I think that for now, we might as well stick with the import chooser, since user will have to use it eventually, to fix ambiguous imports.

fbricon:
and yeah, automatically copying imports from source file is something we could add later

testforstephen:
Ideally, the user does not need to perceive the presence of this feature. And if they find this feature, it is a surprise.

fbricon:
@testforstephen try the updated version

PR #1104, opened 2019-10-16
Global jenv support for vscode-java

timolehto:
This is achieved by updating `find-java-home` to which I made patch which should allow it to properly discover the actual sdk paths from the jenv shims. Included in this update there is also some enhanced windows registry based sdk resolving logic.

Also do note, I've not tested this fully in use with vscode, I just played around with the `find-java-home` and it should work. I tired quickly to build this and install it, but then vscode failed to load it always with or without my fixes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2194,9 +2194,9 @@
       }
     },
     "find-java-home": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/find-java-home/-/find-java-home-1.0.0.tgz",
-      "integrity": "sha512-eqC9OCUT3stF1zfDJ/g0Uz9XFPi5X+2ZAv84Y+itA2x5scCe6GCF585re8YjjoJeIOipjNMRMM7s2E3ige+6+w==",
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/find-java-home/-/find-java-home-1.1.0.tgz",
+      "integrity": "sha512-bSTCKNZ193UM/+ZZoNDzICAEHcVywovkhsWCkZALjCvRXQ+zXTe/XATrrP4CpxkaP6YFhQJOpyRpH0P2U/woDA==",
       "requires": {
         "which": "~1.0.5",
         "winreg": "~1.2.2"
--- a/package.json
+++ b/package.json
@@ -554,7 +554,7 @@
   },
   "dependencies": {
     "vscode-languageclient": "5.3.0-next.9",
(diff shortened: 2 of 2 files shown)


fbricon:
Ok so this partially works. The global jenv shim is loaded properly but if I open a folder containing a local .java-version file, the local value is ignored. My guess is the upstream `find-java-home` library needs an option to pass a base folder, used to infer the proper Java home, based on the local jenv settings

timolehto:
Right, I see. When this `find-java-home` is executed in the context of this plugin it probably isn't run in the project folder or if it is then it's somehow isolated. The way `find-java-home` works is it normally just calls `which.js` on the thing it's looking for (e.g. `java`) and what I did was just added a feature to execute a system call `jenv which java` instead if it looks like which returns a jenv shim. However, this only takes `$(pwd)/.java-version` into consideration.

If we could `cd` into project folder before calling `findJavaHome` it should work. I have no idea how feasible would it be to wrap `findJavaHome` into a pushd/popd calls using [shelljs](https://documentup.com/shelljs/shelljs#pushdoptions-dir---n--n)? What you think, @fbricon?

timolehto:
Alternative solution is that the project root would need to passed to `findJavaHome` which, if the author accepts, could be made to support such an argument and the `jenv` resolution could use it.

fbricon:
> Alternative solution is that the project root would need to passed to `findJavaHome` which, if the author accepts, could be made to support such an argument and the `jenv` resolution could use it.

yeah that was my idea

timolehto:
@fbricon as anyways the only viable solution to this is to do directory switching. Would you be willing to consider a PR that does the directory switching on this side? This would have some advantages. Like any other projects that might not be interested in `jenv` support wouldn't need to include extra dependencies (`shelljs`) in their project. If I include this logic into `find-java-home` then everyone depending on it will also get `shelljs`.

brunoborges:
Any chance to get this PR merged?

fbricon:
I'm fine updating the `find-java-home` dependency to 1.1.0 for now, but as I said before this only partially solves the jenv support. We should update this PR's title to "Provide global jenv support"

fbricon:
> @fbricon as anyways the only viable solution to this is to do directory switching. Would you be willing to consider a PR that does the directory switching on this side? 

Sure, let's see how that looks

fbricon:
@timolehto thanks for this PR. Hope you find a way to provide support for local jenv settings :-)

timolehto:
Yeah, I will try to find the time @fbricon !

PR #1151, opened 2019-11-19
Added getDocumentSymbols call to extension API

omusil24:
Hello everyone,

at Qore Technologies we are working on a VSCode extension and in it we need support for finding symbols in Java files. Adding a whole language server in it and supporting it would be way too much work, especially since your extension already has everything we need. There just wasn't a way to call the language server you are using from outside the extension.

This PR adds function `getDocumentSymbols` to the extension API, which can then be called by other extensions. This function performs `textDocument/documentSymbol` call to the language server and returns a promise with the result.

If you have any questions just ask, or if this PR has incorrect style etc. then I can fix that, no problem.

Signed-off-by: Ondřej Musil <ondrej.musil@qoretechnologies.com>

### DIFF
--- a/src/documentSymbols.ts
+++ b/src/documentSymbols.ts
@@ -1,0 +1,12 @@
+'use strict';
+
+import { LanguageClient, DocumentSymbolParams } from "vscode-languageclient";
+import { DocumentSymbolRequest, DocumentSymbolsResponse } from "./protocol";
+
+export type getDocumentSymbolsCommand = (params: DocumentSymbolParams) => Promise<DocumentSymbolsResponse>;
+
+export function getDocumentSymbolsProvider(languageClient: LanguageClient): getDocumentSymbolsCommand {
+    return async (params: DocumentSymbolParams): Promise<DocumentSymbolsResponse> => {
+        return languageClient.sendRequest(DocumentSymbolRequest.type, params);
+    };
+}
--- a/src/documentSymbols.ts
+++ b/src/documentSymbols.ts
@@ -2,11 +2,14 @@
 
 import {
     CancellationToken,
+    DocumentSymbol,
     DocumentSymbolParams,
     DocumentSymbolRequest,
-    LanguageClient
+    LanguageClient,
+    SymbolInformation
 } from "vscode-languageclient";
-import { DocumentSymbolsResponse } from "./protocol";
+
+type DocumentSymbolsResponse = DocumentSymbol[] | SymbolInformation[] | null;
 
 export type getDocumentSymbolsCommand = (params: DocumentSymbolParams, token?: CancellationToken) => Promise<DocumentSymbolsResponse>;
 
--- a/src/documentSymbols.ts
+++ b/src/documentSymbols.ts
@@ -1,12 +1,20 @@
 'use strict';
 
-import { LanguageClient, DocumentSymbolParams } from "vscode-languageclient";
-import { DocumentSymbolRequest, DocumentSymbolsResponse } from "./protocol";
+import {
+    CancellationToken,
+    DocumentSymbolParams,
+    DocumentSymbolRequest,
+    LanguageClient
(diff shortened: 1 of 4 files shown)


src/protocol.ts
```diff
@@ -356,3 +370,9 @@ export interface LinkLocation extends Location {
 export namespace FindLinks {
     export const type = new RequestType<FindLinksParams, LinkLocation[], void, void>('java/findLinks');
 }
+
+export type DocumentSymbolsResponse = DocumentSymbol[] | SymbolInformation[] | null;
+
+export namespace DocumentSymbolRequest {
```
testforstephen: The standard 'vscode-languageclient' contains the defintion DocumentSymbolRequest. No need to redefine it here.

src/documentSymbols.ts
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+import { LanguageClient, DocumentSymbolParams } from "vscode-languageclient";
+import { DocumentSymbolRequest, DocumentSymbolsResponse } from "./protocol";
```
testforstephen: Please use DocumentSymbolRequest in vscode-languageclient

src/documentSymbols.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type getDocumentSymbolsCommand = (params: DocumentSymbolParams) => Promise<DocumentSymbolsResponse>;
```
testforstephen: Please expose the cancellation token paramter of the original languageClient.sendRequest api.

```java
(params: DocumentSymbolParams, token?: CancellationToken) => Promise<DocumentSymbol[] | SymbolInformation[] | null>
```

omusil24:
@testforstephen I fixed the issues that you pointed out but suddenly the calls started to fail with the following response error:
> ResponseError: Message could not be parsed.

> code: -32700

> Object {message: "java.lang.IllegalStateException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column 74 path $.params"}

```
"Error: Message could not be parsed.
	at handleResponse (/home/omusil/.vscode/extensions/redhat.java-0.54.0/node_modules/vscode-languageserver-protocol/node_modules/vscode-jsonrpc/lib/main.js:448:48)
	at processMessageQueue (/home/omusil/.vscode/extensions/redhat.java-0.54.0/node_modules/vscode-languageserver-protocol/node_modules/vscode-jsonrpc/lib/main.js:275:17)
	at Immediate.setImmediate (/home/omusil/.vscode/extensions/redhat.java-0.54.0/node_modules/vscode-languageserver-protocol/node_modules/vscode-jsonrpc/lib/main.js:259:13)
	at runCallback (timers.js:694:18)
	at tryOnImmediate (timers.js:665:5)
	at processImmediate (timers.js:647:5)"
```

vscode_java_error

Any idea what could be causing this? Is it maybe a bug in the language server? Cause I'm calling the function exactly the same way like before with same parameters, but now with the DocumentSymbolRequest from vscode-languageclient and the token added.

omusil24:
Actually, now it's working fine. Just had to put in a check for undefined CancellationToken. After that it started working.

src/protocol.ts
```diff
@@ -356,3 +369,5 @@ export interface LinkLocation extends Location {
 export namespace FindLinks {
     export const type = new RequestType<FindLinksParams, LinkLocation[], void, void>('java/findLinks');
 }
+
+export type DocumentSymbolsResponse = DocumentSymbol[] | SymbolInformation[] | null;
```
testforstephen: DocumentSymbolsResponse is just an alias. Move it to the place you want to use.

omusil24:
@testforstephen is there anything else that I need to do? Or when will this get merged?

fbricon:
@omusil24 thanks for your contribution, keep 'em coming!

PR #1159, opened 2019-11-26
Show busy status when there are incomplete tasks

akaroml:


fbricon:
Thanks @akaroml !

PR #1160, opened 2019-11-26
Warn about sensible java preferences in project settings

### LINKED ISSUE #1154: Potential security issue
I think I have found a security issue in this repository. Is there a more secure way of reporting such an issue on this repository?

snjeza:
Fixes #1154 

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -4,7 +4,7 @@ import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
 import { workspace, extensions, ExtensionContext, window, StatusBarAlignment, commands, ViewColumn, Uri, CancellationToken, TextDocumentContentProvider, TextEditor, WorkspaceConfiguration, languages, IndentAction, ProgressLocation, InputBoxOptions, Selection, Position, EventEmitter, OutputChannel } from 'vscode';
-import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, RevealOutputChannelOn, Position as LSPosition, Location as LSLocation, StreamInfo, VersionedTextDocumentIdentifier, ErrorHandler, Message, ErrorAction, CloseAction, InitializationFailedHandler } from 'vscode-languageclient';
+import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, RevealOutputChannelOn, Position as LSPosition, Location as LSLocation, StreamInfo, VersionedTextDocumentIdentifier, ErrorHandler, Message, ErrorAction, CloseAction, InitializationFailedHandler, DidChangeConfigurationNotification } from 'vscode-languageclient';
 import { onExtensionChange, collectJavaExtensions } from './plugin';
 import { prepareExecutable, awaitServerConnection } from './javaServerStarter';
 import { getDocumentSymbolsCommand, getDocumentSymbolsProvider } from './documentSymbols';
@@ -22,7 +22,7 @@ import * as refactorAction from './refactorAction';
 import * as pasteAction from './pasteAction';
 import * as net from 'net';
 import { getJavaConfiguration } from './utils';
-import { onConfigurationChange, excludeProjectSettingsFiles } from './settings';
+import { onConfigurationChange, excludeProjectSettingsFiles, getKey, IS_WORKSPACE_JDK_ALLOWED } from './settings';
 import { logger, initializeLogFile } from './log';
 import glob = require('glob');
 import { serverTasks } from './serverTasks';
@@ -125,7 +125,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 
 	enableJavadocSymbols();
 
-	return requirements.resolveRequirements().catch(error => {
+	return requirements.resolveRequirements(context).catch(error => {
 		// show error
 		window.showErrorMessage(error.message, error.label).then((selection) => {
 			if (error.label && error.label === selection && error.command) {
@@ -137,7 +137,6 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 	}).then(requirements => {
 		return new Promise((resolve, reject) => {
 			const workspacePath = path.resolve(storagePath + '/jdt_ws');
-
 			// Options to control the language client
 			const clientOptions: LanguageClientOptions = {
 				// Register the server for java
@@ -152,7 +151,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 				initializationOptions: {
 					bundles: collectJavaExtensions(extensions.all),
 					workspaceFolders: workspace.workspaceFolders ? workspace.workspaceFolders.map(f => f.uri.toString()) : null,
-					settings: { java: getJavaConfiguration() },
+					settings: { java: getJavaConfig(requirements.java_home) },
 					extendedClientCapabilities: {
 						progressReportProvider: getJavaConfiguration().get('progressReports.enabled'),
 						classFileContentsSupport: true,
@@ -168,6 +167,14 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 						clientHoverProvider: true,
 					},
 					triggerFiles: getTriggerFiles()
+				},
+				middleware: {
+					workspace: {
+					  didChangeConfiguration: () => {
+						languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getJavaConfig(requirements.java_home) });
+						onConfigurationChange(languageClient, context);
+					  }
+					}
 				},
 				revealOutputChannelOn: RevealOutputChannelOn.Never,
 				errorHandler: new ClientErrorHandler(extensionName),
@@ -202,7 +209,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			if (!port) {
 				const lsPort = process.env['JDTLS_CLIENT_PORT'];
 				if (!lsPort) {
-					serverOptions = prepareExecutable(requirements, workspacePath, getJavaConfiguration());
+					serverOptions = prepareExecutable(requirements, workspacePath, getJavaConfig(requirements.java_home), context);
 				} else {
 					serverOptions = () => {
 						const socket = net.connect(lsPort);
@@ -443,6 +450,13 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			context.subscriptions.push(onConfigurationChange(languageClient, context));
 		});
 	});
+}
+
+function getJavaConfig(javaHome: string) {
+	const origConfig = getJavaConfiguration();
+	const javaConfig = JSON.parse(JSON.stringify(origConfig));
+	javaConfig.home = javaHome;
+	return javaConfig;
 }
 
 export function deactivate(): Thenable<void> {
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -5,20 +5,22 @@ import * as glob from 'glob';
 import * as os from 'os';
 import { StreamInfo, Executable, ExecutableOptions } from 'vscode-languageclient';
 import { RequirementsData } from './requirements';
-import { getJavaEncoding } from './settings';
+import { getJavaEncoding, IS_WORKSPACE_VMARGS_ALLOWED, getKey, getJavaagentFlag } from './settings';
 import { logger } from './log';
+import { getJavaConfiguration } from './utils';
+import { workspace, ExtensionContext } from 'vscode';
 
 declare var v8debug;
 const DEBUG = (typeof v8debug === 'object') || startedInDebugMode();
 
-export function prepareExecutable(requirements: RequirementsData, workspacePath, javaConfig): Executable {
+export function prepareExecutable(requirements: RequirementsData, workspacePath, javaConfig, context: ExtensionContext): Executable {
 	const executable: Executable = Object.create(null);
 	const options: ExecutableOptions = Object.create(null);
 	options.env = process.env;
 	options.stdio = 'pipe';
 	executable.options = options;
 	executable.command = path.resolve(requirements.java_home + '/bin/java');
-	executable.args = prepareParams(requirements, javaConfig, workspacePath);
+	executable.args = prepareParams(requirements, javaConfig, workspacePath, context);
 	logger.info(`Starting Java server with: ${executable.command} ${executable.args.join(' ')}`);
 	return executable;
 }
@@ -39,7 +41,7 @@ export function awaitServerConnection(port): Thenable<StreamInfo> {
 	});
 }
 
-function prepareParams(requirements: RequirementsData, javaConfiguration, workspacePath): string[] {
+function prepareParams(requirements: RequirementsData, javaConfiguration, workspacePath, context: ExtensionContext): string[] {
 	const params: string[] = [];
 	if (DEBUG) {
 		params.push('-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=1044,quiet=y');
@@ -60,8 +62,23 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
 	if (DEBUG) {
 		params.push('-Dlog.level=ALL');
 	}
-
-	const vmargs = javaConfiguration.get('jdt.ls.vmargs', '');
+	let vmargsCheck = workspace.getConfiguration().inspect('java.jdt.ls.vmargs').workspaceValue;
+	if (vmargsCheck !== undefined) {
+		const agentFlag = getJavaagentFlag(vmargsCheck);
+		if (agentFlag !== null) {
+			const keyVmargs = getKey(IS_WORKSPACE_VMARGS_ALLOWED, context.storagePath, vmargsCheck);
+			const key = context.globalState.get(keyVmargs);
+			if (key !== true) {
+				vmargsCheck = workspace.getConfiguration().inspect('java.jdt.ls.vmargs').globalValue;
+			}
+		}
+	}
+	let vmargs;
+	if (vmargsCheck !== undefined) {
+		vmargs = vmargsCheck + '';
+	} else {
+		vmargs = '';
+	}
 	const encodingKey = '-Dfile.encoding=';
 	if (vmargs.indexOf(encodingKey) < 0) {
 		params.push(encodingKey + getJavaEncoding());
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -1,12 +1,13 @@
 'use strict';
 
-import { workspace, Uri, env } from 'vscode';
+import { workspace, Uri, env, window, ConfigurationTarget, commands, ExtensionContext } from 'vscode';
 import * as cp from 'child_process';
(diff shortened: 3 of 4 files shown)


src/extension.ts
```diff
@@ -125,6 +127,11 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 
 	enableJavadocSymbols();
 
+	const isAllowed = await checkJavaPreferences(context.globalState);
+	if (!isAllowed) {
```
fbricon: please log the reason of the early exit

src/extension.ts
```diff
@@ -453,6 +460,50 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+			await window.showErrorMessage(`Security Warning! The java.home variable defined in ${env.appName} settings points to (${javaHome}). Do you allow it to be used?`, deny, allow).then(selection => {
+				if (selection === allow) {
+					globalState.update(key, true);
+				} else {
+					globalState.update(key, false);
+					workspace.getConfiguration().update('java.home', null);
+				}
+			});
+		}
+		isAllowed = globalState.get(key);
```
fbricon: in case globalstate was already set, it's read twice. So it should move inside the previous if block.

src/extension.ts
```diff
@@ -453,6 +460,50 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+		const vmargs:string = getJavaConfiguration().get('jdt.ls.vmargs');
+		if (vmargs === undefined) {
+			return isAllowed;
+		}
+		const javaagent = '-javaagent';
+		if (vmargs.indexOf(javaagent) >= 0) {
+			const keyVmargs = IS_WORKSPACE_VMARGS_ALLOWED + "::" + vmargs;
+			isAllowed = globalState.get(keyVmargs);
+			if (isAllowed == undefined) {
+				await window.showErrorMessage(`Security Warning! The java.jdt.ls.vmargs variable defined in ${env.appName} settings \n(${vmargs}) includes the javagent preference. Do you allow it to be used?`, deny, allow).then(selection => {
```
fbricon: includes the `javagent` flag

src/extension.ts
```diff
@@ -453,6 +460,50 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+					if (selection === allow) {
+						globalState.update(keyVmargs, true);
+						isAllowed = true;
+					} else {
```
fbricon: } else if (selection === deny) {

src/extension.ts
snjeza: It has been included.

src/extension.ts

src/extension.ts

src/extension.ts
```diff
@@ -453,6 +460,50 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+	const allow = 'Allow';
+	const deny = 'Deny';
+	if (!isAllowed) {
+		const key = IS_WORKSPACE_JDK_ALLOWED + "::" + javaHome;
+		isAllowed = globalState.get(key);
+		if (isAllowed === undefined) {
+			await window.showErrorMessage(`Security Warning! The java.home variable defined in ${env.appName} settings points to (${javaHome}). Do you allow it to be used?`, deny, allow).then(selection => {
+				if (selection === allow) {
+					globalState.update(key, true);
+				} else {
```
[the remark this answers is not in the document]

snjeza:

src/extension.ts
fbricon: I meant s/includes the javagent preference/includes the javagent flag

snjeza:

snjeza:
test this please

snjeza:

src/extension.ts
```diff
@@ -125,6 +127,12 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 
 	enableJavadocSymbols();
 
+	const isAllowed = await checkJavaPreferences(context.globalState);
+	if (!isAllowed) {
+		logger.error(`Failed to initialize ${extensionName} due to security constraint.`);
```
fbricon: actually I think we need to explain to the user the extension couldn't start because of some project settings were disabled

src/extension.ts
```diff
@@ -125,6 +127,12 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 
 	enableJavadocSymbols();
 
+	const isAllowed = await checkJavaPreferences(context.globalState);
```
fbricon: might be better if checkJavaPreferences threw an exception with a message explaining the reason why the extension is not allowed to start

src/extension.ts
```diff
@@ -453,6 +461,63 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 	});
 }
 
+async function checkJavaPreferences(globalState: Memento) {
+	const javaHome = workspace.getConfiguration().get<string>('java.home', null);
+	let isAllowed = javaHome === undefined || javaHome === null || path.isAbsolute(javaHome);
```
fbricon: I wonder if all paths should be discouraged, even the absolute ones. How is the integrated shell in vscode checked?

snjeza: vscode checks the workspace paths.

snjeza:

akaroml:
Here are my 2 cents for this problem. If I understand correctly, the goal is to avoid the attach when the user clone some repository and the repository contains `java.home` that points to some script that will be executed as an attack.

If the assumption is true, could we try change the scope of this setting entry to `machine`. In the documentation, the `machine` scope looks like designed just for this purpose.

> machine - Machine specific settings that can be set in user or remote settings. For example, ***an installation path which shouldn't be shared across machines***.

Checkout the scope section of this doc:
https://code.visualstudio.com/api/references/contribution-points#contributes.configuration

In this way, putting such a setting in the workspace shouldn't take effect.

The current scope of `java.home` is `window`, which can lead to the attack.

akaroml:
Looks like we already had the discussion in #950 

Here are some further thoughts. To enable the case of using different JDK per workspace, we are at the risk of failing to launch vscode-java extension, or exposing a security hole. I would recommend that we reevaluate the importance of such a use case. 

On the other hand, jenv was already supported by vscode-java via findjavahome (#1104). This could be a nice alternative and we can recommend this approach for such case.

fbricon:
I'm still -1 on giving machine scope to java.home, as I explained before. 
#1104 doesn't provide local jenv support. And I don't necessarily want to depend on jenv for that.
vscode itself handles this use case for the shell: https://github.com/microsoft/vscode/blob/25369137eeabe996d950cb72dd0fda73d96cc35a/src/vs/workbench/contrib/terminal/browser/terminalConfigHelper.ts#L186. If it's good enough for them, it's good enough for us :-)
The long term solution lies in local workspace settings : https://github.com/microsoft/vscode/issues/40233

src/extension.ts
```diff
@@ -125,6 +127,13 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 
 	enableJavadocSymbols();
 
+	try {
+		await checkJavaPreferences(context.globalState);
+	} catch (error) {
+		logger.error(error);
+		return;
```
akaroml: If the user denies the use of `java.home` from workspace settings, we can still fall back to use whatever in the PATH or specified by JAVA_HOME, right?

Shutting down the extension at this point looks unnecessary.

akaroml:
@fbricon while we wait for microsoft/vscode#40233, my suggestion would be trying our best to make the extension running. So I requested the change to use whatever specified in PATH or JAVA_HOME (let findjavahome do its thing) as a fallback mechanism.

src/extension.ts
snjeza: I have fixed it.

src/extension.ts
```diff
@@ -453,6 +466,71 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+		javaHome = workspace.getConfiguration().inspect<string>('java.home').workspaceFolderValue;
+	}
+	let isAllowed = javaHome === undefined || javaHome === null;
+	const allow = 'Allow';
+	const deny = 'Deny';
+	if (!isAllowed) {
+		const key = `${IS_WORKSPACE_JDK_ALLOWED}::${javaHome}`;
+		isAllowed = globalState.get(key);
+		if (isAllowed === undefined) {
+			await window.showErrorMessage(`Security Warning! The java.home variable defined in ${env.appName} settings points to (${javaHome}). Do you allow it to be used?`, deny, allow).then(selection => {
```
fbricon: `Security Warning! Do you allow this workspace to set the java.home variable? \n
java.home: ${javaHome}`

src/extension.ts
```diff
@@ -453,6 +466,71 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+				if (selection === allow) {
+					globalState.update(key, true);
+				} else if (selection === deny) {
+					globalState.update(key, false);
```
fbricon: I don't quite like modifying the workspace settings. VS Code keeps denied "terminal.integrated.shell" settings in the workspace. 
We should just ignore denied project settings in requirements.checkJavaRuntime()

src/extension.ts
```diff
@@ -125,6 +128,16 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 
 	enableJavadocSymbols();
 
+	try {
+		await checkJavaPreferences(context.globalState);
```
fbricon: Since we're not going to prevent the extension from activating, this check should be moved to requirements.checkJavaRuntime()

src/extension.ts
```diff
@@ -125,6 +128,16 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+	} catch (error) {
+		if (error === RELOAD_WINDOW) {
```
fbricon: that should not be necessary

src/extension.ts
```diff
@@ -453,6 +466,71 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 	});
 }
 
+async function checkJavaPreferences(globalState: Memento) {
```
testforstephen: extension.ts is a bit too long, move the logic to the utility file settings.ts?

akaroml: I'm having the same idea but that could be a separate PR.

src/extension.ts
```diff
@@ -453,6 +466,71 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+	let javaHome = workspace.getConfiguration().inspect<string>('java.home').workspaceValue;
+	if (javaHome === undefined || javaHome === null) {
+		javaHome = workspace.getConfiguration().inspect<string>('java.home').workspaceFolderValue;
+	}
+	let isAllowed = javaHome === undefined || javaHome === null;
+	const allow = 'Allow';
+	const deny = 'Deny';
```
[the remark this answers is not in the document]

src/extension.ts

src/extension.ts

src/extension.ts

src/extension.ts

src/extension.ts
snjeza: if we omit https://github.com/redhat-developer/vscode-java/pull/1160/commits/176b9278965d8d0ca0281fcd4db4df9bf98ffcbe#diff-f953b3a2fb6d7cd16d698364c6a1dc45R140 VS Code will start ../poc/jdk/bin/java.

src/settings.ts
```diff
@@ -116,3 +119,74 @@ export function getJavaEncoding(): string {
...
+
+export async function checkJavaPreferences(globalState: Memento) {
+	let javaHome = workspace.getConfiguration().inspect<string>('java.home').workspaceValue;
+    let isAllowed = javaHome === undefined || javaHome === null;
+    if (isAllowed) {
+        javaHome = workspace.getConfiguration().inspect<string>('java.home').globalValue;
+    }
+	const allow = 'Allow';
+    const disallow = 'Disallow';
+    const key = `${IS_WORKSPACE_JDK_ALLOWED}::${javaHome}`;
```
fbricon: should be a per-workspace key so more like ${IS_WORKSPACE_JDK_ALLOWED}::${workspathPath}

src/extension.ts
fbricon: ok, but that's not vscode, it's jdt.ls running the executable. laaame

src/settings.ts
snjeza: Fixed.

snjeza:
@fbricon @testforstephen @akaroml I have updated the PR.

src/settings.ts
```diff
@@ -116,3 +120,85 @@ export function getJavaEncoding(): string {
...
+						globalState.update(keyVmargs, true);
+					} else if (selection === disallow) {
+						globalState.update(keyVmargs, false);
+						throw `Failed to initialize ${extensionName} due denied the javaagent (${agentFlag}) preference.`;
+					} else {
+						throw `Failed to initialize ${extensionName} due disabled the javaagent (${agentFlag}) preference.`;
+					}
+                });
+                isAllowed = globalState.get(keyVmargs);
+			}
```
fbricon: Indentation is wrong

snjeza: I have updated the PR.

src/settings.ts
```diff
@@ -116,3 +120,85 @@ export function getJavaEncoding(): string {
 	}
 	return javaEncoding;
 }
+
+export async function checkJavaPreferences(context: ExtensionContext) {
+	let javaHome = workspace.getConfiguration().inspect<string>('java.home').workspaceValue;
+	let isAllowed = javaHome === undefined || javaHome === null;
```
akaroml: The definition of `isAllowed` is vague. I thought it meant whether users allow the use of the path set by `java.home`. But here it equals to the existence of the workspace value of `java.home`. Does that mean if a user sets a value in workspace settings, it is automatically allowed to be used?

If the answer is no, please use a separate variable to track the source of the value.

src/requirements.ts
```diff
@@ -29,16 +31,25 @@ interface ErrorData {
...
-    return new Promise((resolve, reject) => {
+function checkJavaRuntime(context: ExtensionContext): Promise<string> {
+    return new Promise(async (resolve, reject) => {
         let source: string;
-        let javaHome: string = readJavaConfig();
+        let javaHome;
+        try {
+            javaHome = await checkJavaPreferences(context);
+        } catch (error) {
+            reject({
```
akaroml: As I mentioned before, rejecting will fail the launch of the language server. If a user refuses the use of the workspace value, then we could simply tell them that the value from the environment will be used instead. Shutting down the whole thing is unnecessary.

snjeza: See
- https://github.com/redhat-developer/vscode-java/pull/1160/commits/855ffc864930aa22cd3253c5c839bf1f91e4850c#diff-f953b3a2fb6d7cd16d698364c6a1dc45R164
- https://github.com/redhat-developer/vscode-java/pull/1160/commits/855ffc864930aa22cd3253c5c839bf1f91e4850c#diff-f953b3a2fb6d7cd16d698364c6a1dc45R166
- https://github.com/redhat-developer/vscode-java/pull/1160/commits/855ffc864930aa22cd3253c5c839bf1f91e4850c#diff-f953b3a2fb6d7cd16d698364c6a1dc45R172

snjeza: Fixed.

fbricon:
So when the dialog is closed without a response, the bad java.home is sent to the server, which in turn executes it. 
We should probably do like in vscode-xml and sanitize the settings sent to the server, see  https://github.com/redhat-developer/vscode-xml/blob/master/src/extension.ts#L81

snjeza:

src/extension.ts
```diff
@@ -169,6 +171,14 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 					},
 					triggerFiles: getTriggerFiles()
 				},
+				middleware: {
+					workspace: {
+					  didChangeConfiguration: () => {
+						languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: javaConfig });
```
fbricon: the same initial javaConfig is sent to the server on each preference change

snjeza: I have updated the PR.

src/settings.ts
```diff
@@ -116,3 +120,77 @@ export function getJavaEncoding(): string {
...
+					globalState.update(key, false);
+					await workspace.getConfiguration().update('java.home', undefined, ConfigurationTarget.Workspace);
+				}
+			});
+			isVerified = globalState.get(key);
+		}
+	}
+	const vmargs = workspace.getConfiguration().inspect('java.jdt.ls.vmargs').workspaceValue;
+	if (vmargs === undefined || vmargs === null) {
+		return javaHome;
```
fbricon: if the security warning was closed without a selection, and there's no java.jdt.ls.vmargs, then the workspace java.home is returned (and executed) instead of being ignored

snjeza: Fixed.

src/requirements.ts
```diff
@@ -29,16 +30,16 @@ interface ErrorData {
...
     const javaVersion = await checkJavaVersion(javaHome);
     return Promise.resolve({ java_home: javaHome, java_version: javaVersion });
 }
 
-function checkJavaRuntime(): Promise<string> {
-    return new Promise((resolve, reject) => {
+function checkJavaRuntime(context: ExtensionContext): Promise<string> {
+    return new Promise(async (resolve, reject) => {
         let source: string;
-        let javaHome: string = readJavaConfig();
```
fbricon: the readJavaConfig function is no longer used, you can delete it

snjeza: I have deleted it.

src/javaServerStarter.ts
```diff
@@ -60,8 +62,20 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
...
+	let vmargsCheck = workspace.getConfiguration().inspect('java.jdt.ls.vmargs').workspaceValue;
+	if (vmargsCheck !== undefined) {
+		const keyVmargs = getKey(IS_WORKSPACE_VMARGS_ALLOWED, context.storagePath, vmargsCheck);
+		const key = context.globalState.get(keyVmargs);
+		if (key === undefined || key === null) {
+			vmargsCheck = workspace.getConfiguration().inspect('java.jdt.ls.vmargs').globalValue;
+		}
+	}
+	let vmargs;
+	if (vmargsCheck !== undefined) {
```
fbricon: this misses the case where if vmargs were disallowed, they should be ignored

src/javaServerStarter.ts
```diff
@@ -60,8 +62,20 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
 	if (DEBUG) {
 		params.push('-Dlog.level=ALL');
 	}
-
-	const vmargs = javaConfiguration.get('jdt.ls.vmargs', '');
+	let vmargsCheck = workspace.getConfiguration().inspect('java.jdt.ls.vmargs').workspaceValue;
+	if (vmargsCheck !== undefined) {
```
fbricon: check should be ignored if there's no javaagent flag

snjeza:

fbricon:
Thanks @snjeza. This should be much better. Hope I didn't miss anything

PR #1193, opened 2019-12-12
Remove the unused dependencies

jdneo:
- `tmp` is not used in the project so I removed it
- `path-exists` can be replaced by the existing `fs-extra`


fbricon:
Thanks @jdneo

PR #1081, opened 2019-09-29
Response to the server request workspace/configuration

jdneo:
More details: https://github.com/eclipse/eclipse.jdt.ls/pull/1194

Signed-off-by: Sheng Chen <sheche@microsoft.com>

PR #1196, opened 2019-12-17
Add `java.project.referencedLibraries` setting

Vigilans:
This PR adds a setting `java.project.referencedLibraries` based on https://github.com/eclipse/eclipse.jdt.ls/pull/1305.

### DIFF
--- a/package.json
+++ b/package.json
@@ -222,6 +222,32 @@
           ],
           "scope": "window"
         },
+        "java.project.referencedLibraries": {
+          "type": [
+            "array",
+            "object"
+          ],
+          "description": "Configure glob patterns for referencing local libraries to a Java project.",
+          "default": [
+            "lib/**/*.jar"
+          ],
+          "properties": {
+            "include": {
+              "type": "array"
+            },
+            "exclude": {
+              "type": "array"
+            },
+            "sources": {
+              "type": "object"
+            }
+          },
+          "required": [
+            "include"
+          ],
+          "additionalProperties": false,
+          "scope": "resource"
+        },
         "java.contentProvider.preferred": {
           "type": "string",
           "description": "Preferred content provider (a 3rd party decompiler id, usually)",
--- a/package.json
+++ b/package.json
@@ -246,7 +246,7 @@
             "include"
           ],
           "additionalProperties": false,
-          "scope": "resource"
+          "scope": "window"
         },
         "java.contentProvider.preferred": {
           "type": "string",


Vigilans:
Note that, after moving the library logic to server, the `workspace/didChangedWatchedFiles` won't be sent if file is outside of workspace, because `vscode.createWorksapceListener` will only report events of files in workspace folder.

I original proposed to add additional listener to `vscode-java`, by feeding language client's `synchronize.fileEvents` property with a wrapped chokidar listener. But later I thought it is not worthy putting effort and overhead on what is originally not a good practice - reference libraries outside of workspace. So I added a refresh a button to dependency viewer, to let the user refresh if he finds anything not in sync.

package.json
```diff
@@ -222,6 +222,26 @@
...
           "scope": "window"
         },
+        "java.project.referencedLibraries": {
+          "type": [
+            "array",
+            "object"
+          ],
+          "description": "Configure glob patterns for referencing local libraries to a Java project.",
+          "default": [
+            "lib/**/*.jar"
```
Vigilans: Should the default value be `${workspaceFolder}/lib/**/*.jar` after multi-root workspace support of this setting is done?

fbricon: I don't remember if ${workspaceFolder} is supported at the server level

package.json
```diff
@@ -222,6 +222,32 @@
...
+            },
+            "sources": {
+              "type": "object"
+            }
+          },
+          "required": [
+            "include"
+          ],
+          "additionalProperties": false,
+          "scope": "resource"
```
fbricon: Not sure how to scope that properly. Right now, when a folder is opened as part of a workspace, the server ignores the folder setting and falls back to the default global/workspace settings

package.json
Vigilans: Yes, but in this issue https://github.com/eclipse/eclipse.jdt.ls/issues/1303 I would like to support the `${workspaceFolder:name}` setting at server level.

package.json
Vigilans: Here's my investigation:

Currently, a push model (`workspace/onDidChangeConfiguration`) indeed can't help us scope this setting properly, because the push model cannot scope the setting with a uri.

There is a proposal to add a uri to push model: https://github.com/microsoft/language-server-protocol/issues/457, but it is put in backlog now.

If we could use a pull model (`workspace/configuration`), it can be elegantly handled in such way:
1. Server requests `referencedLibraries` config for a specific project;
2. Client organizes the libraries by merging the workspace setting and workspaceFolder setting. So here workspace setting acts as a shared setting.

Vigilans: Switching to pull model has another persuading reason, in that `workspace/didChangeConfiguration` has been deprecated in vscode client:

[client.ts](https://github.com/microsoft/vscode-languageserver-node/blob/a293b0be6c366427c8fe75ffcc79e334606f751d/client/src/client.ts#L325-L334):

```ts
export interface SynchronizeOptions {
	/**
	 * The configuration sections to synchronize. Pushing settings from the
	 * client to the server is deprecated in favour of the new pull model
	 * that allows servers to query settings scoped on resources. In this
	 * model the client can only deliver an empty change event since the
	 * actually setting value can vary on the provided resource scope.
	 *
	 * @deprecated Use the new pull model (`workspace/configuration` request)
	 */
	configurationSection?: string | string[];


	/**
	 *
	 */
	fileEvents?: VFileSystemWatcher | VFileSystemWatcher[];
}
```


and

https://github.com/microsoft/language-server-protocol/issues/567#issuecomment-420589320:
> Having a push model for configurations was a mistake and got basically replaced by the pull model where the server sends workspace/configuration requests.
> 
> The client should still send workspace/didChangeConfiguration notification so that the server can clear caches if it caches configurations. But it should not send any values since values can differ based on the scope used in the workspace/configuration request.

Vigilans: The workaround of this issue:
####  Immediately in this PR
Change the setting to `window` scope. So when creating a workspace from a folder, the settings will be copied to that workspace, and user will know that the folder setting isn't applicable anymore.

#### After enhanced multi-root workspace support
* If use push model: merge all settings in workspace altogether, prefix settings in each workspace with `${workspaceFolder:name}`, then use the merged setting across projects.
* If use pull model: just as stated in first comment in this review. https://github.com/redhat-developer/vscode-java/pull/1081 also utilize the pull model.

Vigilans:
I've changed the scope to `window`, because `resource` won't come into effect until now. 

If current limitaions is acceptable:
* No multi-root workspace support

Then this feature is ready to publish.

fbricon:
@Vigilans Could you please document how to use this feature in https://github.com/redhat-developer/vscode-java/wiki?

Vigilans:
@fbricon Sure, I am out today and will document it tomorrow.

Vigilans:
## Referenced Libraries support for standalone projects

If you're not using any build tool like maven or gradle, and you want to reference binary jars in local file system, then `java.project.referencedLibraries` will meet your need.

#### Include libraries

The libraries to reference is described using a set of glob patterns:

```json
"java.project.referencedLibraries": [
    "library/**/*.jar",
    "/home/username/lib/foo.jar"
]
```

In this way, all `.jar` files in workspace's `library` folder, and `foo.jar` in the specified absolute path is added to the project's external dependencies.

The referenced libraries is then watched by VS Code, and the project will be refreshed once there's change in these dependent files. 

By default, VS Code will reference all jar files in workspace's `lib` directory using glob `lib/**/*.jar`.

#### Exclude some libraries

If you want to exclude some libraries from the project, just expand `java.project.referencedLibraries` to full `include-exclude-sources` pattern and provide globs to `exclude` field:

```json
"java.project.referencedLibraries": {
    "include": [
        "library/**/*.jar",
        "/home/username/lib/foo.jar"
    ],
    "exclude": [
        "library/sources/**"
    ]
}
```

In this way, any binary jar in `library/sources` folder is ignored from the project's external dependencies.

#### Attach source jars

By default, a referenced `{binary}.jar` will try to search `{binary}-sources.jar` under the same directory, and attach it as source if one match is found.

If you want to manually specify a jar as source attachment, you can provide a key-value map in the `sources` field: 

```json
"java.project.referencedLibraries": {
    "include": [
        "library/**/*.jar",
        "/home/username/lib/foo.jar"
    ],
    "exclude": [
        "library/sources/**"
    ],
    "sources": {
        "library/bar.jar": "library/sources/bar-src.jar"
    }
}
```

In this way, `bar-src.jar` is attached to `bar.jar` as its source.


### Known issues
* VS Code does not support listening to files outside of workspace. If you change a referenced library which locates in somewhere other than current workspace, you should manually trigger a refresh.  
* Multi-root workspace is not well supported yet.

Vigilans:
There will be some new content after `java-dependency-viewer` support of this setting is released, which contains:
* Adding a file or a folder to library
* Removing a library from project
* Refreshing the library

husseinkohy:
this (java.project.referrencedLibs) works on new java projects, however, have difficulty with existing java projects in Mac. this is not a multi-root project, etc. thanks

Vigilans:
@husseinkohy Is your project a eclipse, maven, or gradle one? This setting only works for "invisible project" (which does not have any build file (.classpath, pom.xml, etc...) in your workspace). If you have a build tool in your project, just follow their way to add the dependency.

husseinkohy:
it is actually ant ivy, cannot get ref libs with ant target runner extension.

brunoborges:
Just tried this as of today with latest versions and the property `java.project.referrencedLibs` does not work.

apupier:
> Just tried this as of today with latest versions

do you mean latest released version? or latest snapshot?

Can you precise the files that you have in the workspace?

it is is working with latest released version on my side. (through usage of VS Code Camel K extension)

brunoborges:
Latest released, not snapshot, and I have a naked project (no meta file like .classpath for example).

fbricon:
@brunoborges please provide a sample project reproducing the issue

tomm1e:
Is there a setting to stop .classpath and build.gradle files from being parsed? Both of those will stop the referencedLibraries from loading. I tried disabling gradle in settings and adding those files to every exclusion setting. Only way around it is to delete those two files, clean the server workspace and restart.

PR #1201, opened 2019-12-18
Fix the failure to get trigger file if irrelevant window pops up

### LINKED ISSUE #1198: Java Overview will break the import of invisible project
When opening a folder containing an invisible project, if `Java Overview` pops up, the project will fail to be imported, because `initializationOptions` does not contain a trigger file:

```java
Collection<IPath> triggerFiles = preferencesManager.getPreferences().getTriggerFiles();
if (triggerFiles == null || triggerFiles.isEmpty()) {
    return;
}
```

from https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/managers/InvisibleProjectImporter.java#L72-L75

It is root-caused in `getTriggerFiles`, where it only retrives text editors. When `Java Overview` pops up, the triggered file actually goes to background, and cannot be retrieved in `window.visibleTextEditors`.

https://github.com/redhat-developer/vscode-java/blob/6f82e4d209eccea74149cc4176f4173c0edbc372/src/extension.ts#L846-L852

#### Environment
- Java version: openjdk 1.8
- Visual Studio Code Version:  1.40.2
- Java extension version: master branch

#### Reproduce steps
1. Make sure `Show overview page when using Java` is set
2. Open a folder containing an invisible project.

Sample project: https://github.com/santiontanon/microrts

#### Current Result

"Classpath is incomplete. Only syntax error is reported."

#### Expected Result

Project should successfully build.

Vigilans:
Fix #1198 

Trigger file will now try to get documents from all opened text documents, if all trials from previous version failed.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -135,8 +135,9 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 		});
 		// rethrow to disrupt the chain.
 		throw error;
-	}).then(requirements => {
-		return new Promise((resolve, reject) => {
+	}).then(async (requirements) => {
+		const triggerFiles = await getTriggerFiles();
+		return new Promise<ExtensionAPI>((resolve, reject) => {
 			const workspacePath = path.resolve(storagePath + '/jdt_ws');
 			// Options to control the language client
 			const clientOptions: LanguageClientOptions = {
@@ -167,7 +168,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 						moveRefactoringSupport: true,
 						clientHoverProvider: true,
 					},
-					triggerFiles: getTriggerFiles()
+					triggerFiles,
 				},
 				middleware: {
 					workspace: {
@@ -822,7 +823,7 @@ async function executeRangeFormat(editor, startPosition, lineOffset) {
 	await commands.executeCommand('editor.action.formatSelection');
 }
 
-function getTriggerFiles(): string[] {
+async function getTriggerFiles(): Promise<string[]> {
 	const openedJavaFiles = [];
 	const activeJavaFile = getJavaFilePathOfTextDocument(window.activeTextEditor && window.activeTextEditor.document);
 	if (activeJavaFile) {
@@ -833,7 +834,7 @@ function getTriggerFiles(): string[] {
 		return openedJavaFiles;
 	}
 
-	workspace.workspaceFolders.forEach(rootFolder => {
+	await Promise.all(workspace.workspaceFolders.map(async (rootFolder) => {
 		if (rootFolder.uri.scheme !== 'file') {
 			return;
 		}
@@ -858,7 +859,14 @@ function getTriggerFiles(): string[] {
 				return;
 			}
 		}
-	});
+
+		for (const javaFile of await workspace.findFiles("**/src/**/*.java", undefined, 1)) { // Find at most 1 java file
+            if (isPrefix(rootPath, javaFile.fsPath)) {
+				openedJavaFiles.push(javaFile.toString());
+				return;
+			}
+		}
+	}));
 
 	return openedJavaFiles;
 }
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -833,31 +833,39 @@ function getTriggerFiles(): string[] {
 		return openedJavaFiles;
 	}
 
-	for (const rootFolder of workspace.workspaceFolders) {
+	workspace.workspaceFolders.forEach(rootFolder => {
 		if (rootFolder.uri.scheme !== 'file') {
-			continue;
```
fbricon: why? if subsequent folders have the file:// prefix, why should they be discarded?

fbricon: mmm so here return simply returns from the lambda, so it's fine

Vigilans: I change the `for...of` pattern to `forEach`, and here `return` is equivalent to `continue` here.

fbricon:
I can still see the bug when no java files have been opened and some commands triggered the activation of vscode-java. 
As a last (ugly) resort, we might check for the existence of java files under a src/ folder at the root. Even then, that will not cover all cases, but we should cover more ground

Vigilans:
@fbricon Searching for java file needs `getTriggerFiles` to be async, do you think we need to move `getTriggerFiles` into `resolveRequirements`?

fbricon:
it's not a requirement per se, but I'd be fine performing the lookup asynchronously in a different method

Vigilans:
I've add java file searching in workspace as the last resort.

It searches all src folder in the workspace, whenever one java file is found, it will stop searching.

src/extension.ts
```diff
@@ -858,7 +859,14 @@ function getTriggerFiles(): string[] {
 				return;
 			}
 		}
-	});
+
+		for (const javaFile of await workspace.findFiles("**/src/**/*.java", undefined, 1)) { // Find at most 1 java file
```
fbricon: I'd rather be more conservative and look for src/**/*.java instead, or else if we might have some performance issues in case a very large folder has been opened

Vigilans: I've changed to `src/**/*.java` now. Is trigger file only used for invisible project? If so, they indeed we may assume that `src` will cover most cases, since multi-module project will use a build tool instead.

But we may also search java files under the workspace root directory, since the user of invisible project may only create java files in the outermost place.

fbricon: > I've changed to src/**/*.java now. Is trigger file only used for invisible project? If so, then indeed we can assume that src will cover most cases, since multi-module project will use a build tool instead.

yes src/**/*.java is only for invisible projects. 

> But we may also search java files under the workspace root directory, since the user of invisible project may only create java files in the outermost place.

yes makes sense

Vigilans:
Now trigger file of one specific workspace folder is being searched as following order:
* Active text editor
* Visible text editors
* Workspace text documents
* Java file in workspace root
* Java file in src or test directory

and will early exit whenever one java file is found.

fbricon:
what are workspace.textDocuments?

Vigilans:
@fbricon Whenever `workspace.openTextDocument(uri)` is called, a text document is created in memory and added to `workspace.textDocuments`. It contains properties like `languageId`, `isUntitled`, `isDirty` and `getText()`.

fbricon:
in which circumstance could we have a Java file present in in textDocuments but not in the list of visible text editors, that would need to trigger the language server?

Vigilans:
It is the first situation that this PR tries to resolve: #1198

When opening a java file to activate the extension, the `Java Overview` pops up, and the java file goes to background.

In this circumstance:
* `window.activeTextEditor` is undefined, because `Java Overview` is a webview, not a text editor.
* `window.visibleTextEditors` does not contain the java file, because the java file is hidden by `Java Overview`.
* `workspace.textDocuments` does contain the java file, because this file has been opened, althogh it is not in a visible state.

fbricon:
Thanks @Vigilans!

PR #1252, opened 2020-01-21
Add Assign statement to new variable/field quick action

### LINKED ISSUE #1208: Add Assign statement to new variable/field quick action
using vscode its extract to local variable seem to take the nearest value rather than nearest statement to extract to variable. It should take the nearest statement instead.

This makes it very tedious to use compared to other IDE's to quickly create java code.

##### Steps To Reproduce
1. have code like:

```
        session.createNativeQuery("select * from Gift").scroll();
```
2.  use ctrl+. while cursor is on `session` and use extract to local variable

##### Current Result
```
        StatelessSession session2 = session;
        session2.createNativeQuery("select * from Gift").scroll();
```

##### Expected Result
```
        ScrollableResults scroll = session.createNativeQuery("select * from Gift").scroll();
```

##### Additional Informations

Furthermore the refactoring should be possible to do anywhere in the statement  and not just on a exact value.

snjeza:
Fixes #1208 
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/1319

### DIFF
--- a/src/refactorAction.ts
+++ b/src/refactorAction.ts
@@ -32,10 +32,12 @@ function registerApplyRefactorCommand(languageClient: LanguageClient, context: E
 
     context.subscriptions.push(commands.registerCommand(javaCommands.APPLY_REFACTORING_COMMAND, async (command: string, params: any, commandInfo: any) => {
         if (command === 'extractVariable'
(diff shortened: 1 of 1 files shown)


snjeza:

PR #1255, opened 2020-01-25
Add quickfix for redundant interface

snjeza:
Fixes https://github.com/eclipse/eclipse.jdt.ls/issues/438
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/1325


fbricon:
I'd rather prefer we do not expose jdt compiler settings at this point

PR #1307, opened 2020-02-27
VS Code Java Cannot Run from Read-Only Location

snjeza:
Fixes #1301 

Signed-off-by: Snjezana Peco <snjezana.peco@redhat.com>

PR #1314, opened 2020-03-03
Run from Read-Only Location - fix dev mode

snjeza:
Fixes #1307 
See https://github.com/redhat-developer/vscode-java/pull/1307#issuecomment-594066710


fbricon:
Thanks @snjeza!

PR #1320, opened 2020-03-06
j.i.gradle.arguments and j.i.gradle.jvmArguments aren't properly defined

### LINKED ISSUE #1319: java.import.gradle.arguments and java.import.gradle.jvmArguments aren't properly defined
See
- https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/Preferences.java#L591
- https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/Preferences.java#L593
- https://github.com/redhat-developer/vscode-java/blob/master/package.json#L154
https://github.com/redhat-developer/vscode-java/blob/master/package.json#L160

snjeza:
Fixes #1319 

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,13 +152,19 @@
           "scope": "window"
         },
         "java.import.gradle.arguments": {
-          "type": "string",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
           "default": null,
           "description": "Arguments to pass to Gradle.",
           "scope": "window"
         },
         "java.import.gradle.jvmArguments": {
-          "type": "string",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
           "default": null,
           "description": "JVM arguments to pass to Gradle.",
           "scope": "window"


fbricon:
I understand the issue but I don't think the proposed solution is consistant with other arg preferences (vmArgs comes to mind). As a user, I find it's also less convenient to use than a plain String.

I'd rather prefer https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/MapFlattener.java#L60 was changed to split strings on each whitespace

snjeza:
> I'd rather prefer https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/MapFlattener.java#L60 was changed to split strings on each whitespace

Related to https://github.com/eclipse/eclipse.jdt.ls/pull/1387

PR #1334, opened 2020-03-19
Support refactoring documentation

jdneo:
#1335

Kapture 2020-03-30 at 14 24 47

Note that the two css files `highlight.css` & `markdown.css` are all copied from https://github.com/microsoft/vscode/tree/master/extensions/markdown-language-features/media


jdneo:
Since the API is not ready in stable version yet, so we can first start working on the document structure.

@fbricon @snjeza @akaroml @testforstephen @Eskibear Please take a look at the structure and give your feedbacks. Once the structure is determined and checked-in. All the teammates can start filling the content into it.

I'll raise another PR to add the implementation part once the document is finished.

fbricon:
Question is: how will you map refactoring ids to markdown anchors, such as #extract-to-constant

jdneo:
That a good question!

Currently, the API will categorize the document according to the `CodeActionKind`, which is defined as:

```ts
export class CodeActionKind {
    static readonly Empty: CodeActionKind;
    static readonly QuickFix: CodeActionKind;

    // --- refactor start
    static readonly Refactor: CodeActionKind;
    static readonly RefactorExtract: CodeActionKind;
    static readonly RefactorInline: CodeActionKind;
    static readonly RefactorRewrite: CodeActionKind;
    // --- refactor end

    static readonly Source: CodeActionKind;
    static readonly SourceOrganizeImports: CodeActionKind;
    static readonly SourceFixAll: CodeActionKind;
}
```

This is somehow different from our `JavaCodeActionKind`, so I think we can prepare a general refactor document as a start.

akaroml:
I guess a mapping mechanism will be needed in the follow-up PR.

src/codeActionProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+    public static readonly metadata: CodeActionProviderMetadata = {
+        providedCodeActionKinds: [ 
+            CodeActionKind.Refactor
+        ],
+        documentation: Array.from(javaRefactorKinds.keys()).map(kind => {
+            return {
+                kind,
+                command: {
+                    command: Commands.LEARN_MORE_ABOUT_REFACTORING,
+                    title: 'Learn more about Java refactorings.',
```
fbricon: `Learn more about Java refactorings...`

the ellipsis indicates it'll open a new window

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+- [Rename](#rename)
+- Type change
+  - [Change resolved type to var type](#change-resolved-type-to-var-type)
+  - [Change var type to resolved type](#change-var-type-to-resolved-type)
+
+---
+
+## Introduction
+
+The goal of the Java program refactoring is to make system-wide code changes without affecting the behavior of the program. The **Java Language Support for Visual Studio Code** provides assistance in easily refactoring code.
```
fbricon: The app name should not be hardcoded to **Visual Studio Code** . Needs to be injected via ${vscode.env.appName}. See https://github.com/redhat-developer/vscode-java/pull/1067

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
+## Table of Content
```
fbricon: we jump straight to the TOC but title of the document is missing.  Should be "Java Refactoring"

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+...
+public void test() {
+  assertEquals(expected, actual);
+}
+```
+
+---
+
+## Extract to constant
+Creates a static final field from the selected expression and substitutes a field reference, and rewrites other places where the same expression occurs.
```
fbricon: ... and substitutes a field reference, then rewrites ...

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+  return PI * r * r;
+}
+```
+
+> Also see: [Inline constant](#inline-constant)
+
+---
+
+## Extract to field
+Declare a new field and initialize it with the selected expression. The original expression is replaced with the usage of the field.
```
fbricon: Declares ... initializes...

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+      // do something
+    };
+  }
+```
+> Also see: [Convert to anonymous class creation](#convert-to-anonymous-class-creation)
+
+---
+
+## Convert to static import
+Convert the field or method to static import.
```
fbricon: Converts

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+  for (String book : books) {
+    // do something
+  }
+}
+```
+
+---
+
+## Convert to lambda expression
+Convert an anonymous class creation to the lambda expression.
```
fbricon: Converts

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+  };
+}
+```
+
+> Also see: [Convert to lambda expression](#convert-to-lambda-expression)
+
+---
+
+## Convert to enhanced for loop
+Convert the simple `for` loop to `for-each` style.
```
fbricon: Converts

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+    final boolean isValid = true;
+    return new MyInterface(isValid);
+  }
+}
+```
+
+---
+
+## Convert to anonymous class creation
+Convert lambda expression to anonymous class creation.
```
fbricon: Converts

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+```
+#### After
+```java
+List<String> fruits = Arrays.asList("apple", "lemon", "banana");
+```
+
+---
+
+## Change modifiers to final
+Add `final` modifier to all the variables and parameters in the current source file.
```
fbricon: Adds

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+```
+> Also see: [Convert to anonymous class creation](#convert-to-anonymous-class-creation)
+
+---
+
+## Convert to static import
+Convert the field or method to static import.
+
+### Example
+Let's trasform the `Assert.assertEquals()` invocation to a static import.
```
fbricon: transform

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+    // do somethinf
+  }
+}
+```
+#### After
+```java
+public void method() {
+  boolean isMac = platform.equalsIgnoreCase("MAC");
+  if (isMac) {
+    // do somethinf
```
fbricon: something

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+
+### Examples
+Let's replace the variable `isMac` directly to the boolean expression.
+#### Before
+
+```java
+public void method() {
+  boolean isMac = platform.equalsIgnoreCase("MAC");
+  if (isMac) {
+    // do somethinf
```
fbricon: something

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+  }
+}
+```
+#### After
+```java
+public void method() {
+  if (platform.equalsIgnoreCase("MAC")) {
+    // do somethinf
```
fbricon: something

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+Creates a new variable assigned to the expression currently selected and replaces the selection with a reference to the new variable.
+
+### Examples
+Let's extract the expression `platform.equalsIgnoreCase("MAC")` to a new variable.
+#### Before
+
+```java
+public void method() {
+  if (platform.equalsIgnoreCase("MAC")) {
+    // do somethinf
```
fbricon: something

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+  }
+}
+```
+
+> Also see: [Extract to local variable](#extract-to-local-variable)
+
+---
+
+## Inline method
+Replace calls to the method with the method’s body.
```
fbricon: Replaces

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+  int area = height * width;
+}
+```
+
+> Also see: [Extract to method](#extract-to-method)
+
+---
+
+## Invert conditions
+Invert the boolean expression in the conditions.
```
fbricon: Inverts

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+  if (value <= 5 || value >= 15) {
+    // do something
+  }
+}
+```
+
+---
+
+## Invert local variable
+Invert the local boolean variable.
```
fbricon: Inverts

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+
+public void myMethod() {
+  Bar myClass = new Bar();
+}
+```
+
+---
+
+## Change resolved type to var type
+Use `var` to declare local variables.
```
fbricon: Uses

document/_java.learnMoreAboutRefactorings.md
```diff
@@ -0,0 +1,593 @@
...
+```java
+var s = "";
+```
+
+> Also see: [Change var type to resolved type](#change-var-type-to-resolved-type)
+
+---
+
+## Change var type to resolved type
+Use the resolved type to declare local variables.
```
fbricon: Uses

jdneo:
Update: the API has been finalized and should be available in the next VS Code stable version, I'll update the PR after the new version released.

jdneo:
PR updated...

To navigate to the specific section in the refactoring document, you need to add a keyboard shortcut for a certain refactoring, for example:

```
{
        "key": "ctrl+shift+alt+e",
        "command": "editor.action.codeAction",
        "args": {
            "kind": "refactor.extract.function",
            "apply": "never"
        }
    }
```

fbricon:
Can you please fix the conflict?

jdneo:
@fbricon Done. Travis linux image failed, seems caused by java 8 runtime.

fbricon:
@jdneo please open a different PR

jdneo:
@fbricon @snjeza The new refactor `Introduce Parameter` is added. Please take a look.

src/standardLanguageClient.ts
```diff
@@ -295,6 +298,13 @@ export class StandardLanguageClient {
 				});
 			}
 			excludeProjectSettingsFiles();
+
+			context.subscriptions.push(markdownPreviewProvider);
+			context.subscriptions.push(languages.registerCodeActionsProvider({ scheme: 'file', language: 'java' }, new RefactorDocumentProvider(), RefactorDocumentProvider.metadata));
+			context.subscriptions.push(commands.registerCommand(Commands.LEARN_MORE_ABOUT_REFACTORING, async (kind: CodeActionKind) => {
+				const sectionId: string = javaRefactorKinds.get(kind) || '';
+				markdownPreviewProvider.show(context.asAbsolutePath(path.join('document', `${Commands.LEARN_MORE_ABOUT_REFACTORING}.md`)), 'Java Refactoring', sectionId, context);
+			}));
```
fbricon, line 307: Can we check the new API is available (think Theia/Che) before registering the new provider?

jdneo: Is there any guidance that we could follow to do such checks?

fbricon: I assume something along the lines of `if (CodeActionProviderMetadata !== null)`. But I'm no typescript expert

jdneo: `CodeActionProviderMetadata` is only used as a prarm when register the code action provider, so probably it's not able to be used for API checking. Using `env.appName.includes('VS Code')`?

// BTW, will theia/che honor the `engines` field in package.json

fbricon: env.appName.includes('VS Code') is bad, since we'd need to rerelease vscode-java once Theia/Che support the feature.
> // BTW, will theia/che honor the engines field in package.json?

That's a question for @benoitf 

I don't understand why we can't do a check on CodeActionProviderMetadata (&& CodeActionProviderMetadata.documentation) like we do for https://github.com/redhat-developer/vscode-java/blob/master/src/semanticTokenProvider.ts#L8

benoitf: `vscode.version` is provided at runtime within Eclipse Che for VS Code extensions but usually it's bumped even if not 100% of the new API is provided. (so it may not be a good check)

registerCodeActionsProvider  method and `CodeActionProviderMetadata` interface are there but `documentation` field is missing for now in Theia/Che.

jdneo: Not sure if I understand your proposal correctly.

Different from the example of semanticTokenProvider, `languages.registerDocumentSemanticTokensProvider` works because `registerDocumentSemanticTokensProvider()` is a member of the object `languages`. While `CodeActionProviderMetadata` is just a interface definition, in the runtime, which is javascript, there is no such type actually.

image

fbricon: if (RefactorDocumentProvider?.metadata?.documentation) then

jdneo: That will always be true because `RefactorDocumentProvider.metadata.document` is defined by ourselves, you can think it's just an object regardless what type we declared for it.

fbricon: Since CodeActionProviderMetadata exists, we should be able to do:

let capm : CodeActionProviderMetadata;
if ('documentation' in capm) {...}

fbricon: doesn't work. I can't see how to check that at runtime. 
So what's the worst thing that could happen? The documentation field will be unused and we'll registered a RefactorDocumentProvider for nothing?

jdneo: I think so.

Is it possible to install a vsix in theia? I tried to launch extension in theia but failed to do that.

fbricon: @jdneo if you link to a vsix build, maybe @benoitf can test it (display the refactoring menu, see there are no errors)

jdneo: Hi @benoitf, 

Would you mind to help test this feature? Here is the vsix [link](https://drive.google.com/file/d/1cb3NB6Gw6y0aYBrv2Nb0zBZ_m7btX9zJ/view?usp=sharing). To test it, you can open a Java file, right click in the editor and click `Learn more about Java refactorings` (if it exists in the context menu).

More details can be found in the [gif](https://user-images.githubusercontent.com/6193897/77882127-ede1a080-7292-11ea-8cb8-953c738e4d51.gif)

benoitf: here is a link to try this vsix with che online: http://che.openshift.io/f/?url=https://gist.githubusercontent.com/benoitf/dcc2fa363b37161f5c8105a8e439a389/raw/55c6c7b117614c45d12a576d8159437c3afd923e/devfile.yaml

benoitf: refactoring

custom documentation does not appear but it's expected as not yet supported

jdneo: Thank you @benoitf! 

@fbricon Seems that a RefactorDocumentProvider is registered for nothing.

Screen Shot 2020-07-20 at 9 38 58 PM

fbricon:
Thanks for the cool feature @jdneo!

PR #1340, opened 2020-03-25
Auto fill file contents when new a Java file

### LINKED ISSUE #1222: Fill in package for newly created file
When creating a new file, e.g `Foo.java` inside the `com.mycomp.app`-folder then the extension should use the new `vscode.workspace.on[Will|Did]CreateFile`-api to fill in some default content, esp the `package`-declaration `package com.mycomp.app` etc.

##### Environment
- Java extension pack 0.8.1

##### Steps To Reproduce
1. Use the vscode file explorer to create a new file

##### Current Result
File is empty

##### Expected Result
File should be pre-populated

testforstephen:

Fix #1222

- If the language server is not initialized yet, new file will only generate the class structure.
- If the language server is ready, then resolve the package name and populate both the package statement and class contents.
- If the new file is not belonged to any known source paths, then leave the package as default.

newfile

### DIFF
--- a/src/fileEventHandler.ts
+++ b/src/fileEventHandler.ts
@@ -8,16 +8,16 @@ import { Commands } from './commands';
 
 let serverReady: boolean = false;
 
-export function setServerStatus(onReady: boolean) {
-    serverReady = onReady;
+export function setServerStatus(ready: boolean) {
+    serverReady = ready;
 }
 
 export function registerFileEventHandlers(client: LanguageClient, context: ExtensionContext, ) {
-    context.subscriptions.push(workspace.onDidCreateFiles(didCreateFiles));
+    context.subscriptions.push(workspace.onDidCreateFiles(handleNewJavaFiles));
 }
 
-async function didCreateFiles(e: FileCreateEvent) {
-    const emptyFiles: string[] = [];
+async function handleNewJavaFiles(e: FileCreateEvent) {
+    const emptyFiles: Uri[] = [];
     const textDocuments: TextDocument[] = [];
     for (const uri of e.files) {
         if (!uri.fsPath || !uri.fsPath.endsWith(".java")) {
@@ -29,7 +29,7 @@ async function didCreateFiles(e: FileCreateEvent) {
             continue;
         }
 
-        emptyFiles.push(uri.fsPath);
+        emptyFiles.push(uri);
         textDocuments.push(textDocument);
     }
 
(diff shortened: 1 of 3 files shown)


src/fileEventHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+        const snippets: string[] = [];
+        const packageName = resolvePackageName(sourcePaths, emptyFiles[i]);
+        if (packageName) {
+            snippets.push(`package ${packageName};`);
+        }
+
+        const typeName: string = resolveTypeName(textDocuments[i].fileName);
+        snippets.push(...[
+            "",
+            `public \${1|class,interface,enum|} ${typeName} {`,
```
fbricon: it should also propose record, if the project is compiled against Java 14 or more recent. See https://github.com/eclipse/eclipse.jdt.ls/pull/1395

src/fileEventHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import * as path from 'path';
+import { workspace, FileCreateEvent, ExtensionContext, FileRenameEvent, window, TextDocument, SnippetString, commands, Uri } from 'vscode';
+import { LanguageClient } from 'vscode-languageclient';
+import { ListCommandResult } from './buildpath';
+import { Commands } from './commands';
+
+let serverReady: boolean = false;
+
+export function setServerStatus(onReady: boolean) {
```
fbricon: s/onReady/ready/g

src/fileEventHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+    serverReady = onReady;
+}
+
+export function registerFileEventHandlers(client: LanguageClient, context: ExtensionContext, ) {
+    context.subscriptions.push(workspace.onDidCreateFiles(didCreateFiles));
+}
+
+async function didCreateFiles(e: FileCreateEvent) {
```
fbricon: s/didCreateFiles/handleNewJavaFile/g

src/fileEventHandler.ts
fbricon: I don't think this is actually doable, as, when `record` is selected, the template should be changed to insert something like `${typeName}(${0})`

testforstephen: yes, the static snippet is not that strong. I just thought of a workaround to make the ${typeName}(${0}) to be part of the choice text, too. See the screenshot:
image

fbricon: does it actually insert the cursor placeholder after selecting the record value?

testforstephen: The placeholder **$0** in the choice item won't take effect.

After selected, the cursor will stop after the closing parenthesis, e.g. "public record Hello()**|** {"

testforstephen:
PR updated.

fbricon:
Merged as https://github.com/redhat-developer/vscode-java/commit/ce48e3fb6de6040afa9f6b81fbcbdade17237703.

Thanks @testforstephen

PR #1101, opened 2019-10-16
Enable client snippet before the language server is ready

jdneo:
Since we have moved some snippets from the client side to server side. The moved snippets will not be available until the language server is ready to provide the completion items.

This PR is trying to resolve this problem. It will register a completion provider to provide those snippets which has been moved to the server side. Once the client received the notification that the language server is ready, the client side provider will be disposed.

Signed-off-by: Sheng Chen <sheche@microsoft.com>

PR #1351, opened 2020-03-30
Snippets for import statements

AtharvaThorve:
Code contains snippets for import statements like
- `imp` becomes `import java..`
- `impu` becomes `import java.util.*`
- `impl` becomes `import java.lang.*`
- `impio` becomes `import java.io.*`<br>
import-statement

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -748,7 +748,7 @@
     },
     "bl": {
       "version": "1.2.2",
-      "resolved": "https://registry.npmjs.org/bl/-/bl-1.2.2.tgz",
+      "resolved": "http://registry.npmjs.org/bl/-/bl-1.2.2.tgz",
       "integrity": "sha512-e8tQYnZodmebYDWGH7KMRvtzKXaJHx3BbilrgZCfvyLUYdKpK1t5PSPmpkny/SgiTSCnjfLW7v5rlONXVFkQEA==",
       "dev": true,
       "requires": {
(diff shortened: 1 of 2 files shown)


fbricon:
@AtharvaThorve while I appreciate the effort, unfortunately, I don't think this PR can be applied:

- this is pretty much something that addresses one's personal needs/tastes (for instance, some people prefer to avoid using wildcard imports). Since vscode-java is already capable of performing auto-import when typing, or auto-import can already be executed on demand (Alt+Shift+o), having import snippets doesn't bring much value and rather bloats the completion list. 
- this change only works when the Java language hasn't finished starting. Once started, the server takes over snippet contributions (see https://github.com/redhat-developer/vscode-java/pull/1101.) So snippets would also need to be implemented on the server side (see https://github.com/eclipse/eclipse.jdt.ls/pull/1201).
- Snippets need to be context sensitive: no need to provide import snippets in the middle of a method. That would also need to be handled on the server side level.

All-in-all, I think this snippet list rather belongs to one's [personal snippets file](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_create-your-own-snippets) or a  snippets extension on the [marketplace](https://marketplace.visualstudio.com/search?term=java&target=VSCode&category=Snippets&sortBy=Relevance).

If you'd like to contribute useful snippets, there are certainly plenty you can help porting from Eclipse IDE to https://github.com/eclipse/eclipse.jdt.ls:
Screen Shot 2020-03-30 at 12 27 22 PM

AtharvaThorve:
Thank you very much for the feedback. I'll go through the  https://github.com/eclipse/eclipse.jdt.ls repository as suggested.

PR #1359, opened 2020-04-02
Sort formatter settings

rafaelrenanpacheco:
Sorting the formatter settings makes it easier to find the desired setting among hundred lines.


snjeza:
@rafaelrenanpacheco Thanks.

PR #1360, opened 2020-04-03
fix checkJavaVersion for Windows

Eskibear:
Originally reported in  https://github.com/microsoft/vscode-java-pack/issues/351

The pack-installer uses the same logic to check java version, error below:
```
{
  "killed": false,
  "code": 1,
  "signal": null,
  "cmd": "C:\\Program Files\\AdoptOpenJDK\\jdk-11.0.6.10-hotspot\\/bin/java -version"
}
```


fbricon:
Thanks @Eskibear !

PR #1368, opened 2020-04-06
workspace/notify notification creates no vscode command as expected

### LINKED ISSUE #1367: workspace/notify notification creates no vscode command as expected
I'm trying to send an notification on JDT LS extension side by sending a notification `workspace/notify` with:

https://github.com/eclipse/eclipse.jdt.ls/blob/cf6358ba4c166fa7d4c45aba5b7b22ea08516524/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/lsp/ExecuteCommandProposedClient.java#L26

On TypeScript client side, I'm trying to consume the expected notification but it doesn't work although it works with `workspace/executeClientCommand` which is a request and not a notification.

It seems that there is a bug in https://github.com/redhat-developer/vscode-java/blob/master/src/extension.ts#L367

```ts
languageClient.onRequest(SendNotificationRequest.type, (params) => {
					});
```

but it should use `onNotification` (not onRequest):

```ts
languageClient.onNotification(SendNotificationRequest.type, (params) => {
					});
```

angelozerr:
workspace/notify notification creates no vscode command as expected 

Fixes #1367

Signed-off-by: azerr <azerr@redhat.com>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -364,7 +364,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 						return commands.executeCommand(params.command, ...params.arguments);
 					});
 
-					languageClient.onRequest(SendNotificationRequest.type, (params) => {
+					languageClient.onNotification(SendNotificationRequest.type, (params) => {
 						return commands.executeCommand(params.command, ...params.arguments);
 					});
 


angelozerr:
I close this PR (which doesn't comes from my vscode-java github repo).

PR #1369, opened 2020-04-06
workspace/notify notification creates no vscode command as expected

### LINKED ISSUE #1367: workspace/notify notification creates no vscode command as expected

angelozerr:
workspace/notify notification creates no vscode command as expected

Fixes #1367

Signed-off-by: azerr azerr@redhat.com

### DIFF
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -112,8 +112,8 @@ export namespace ExecuteClientCommandRequest {
     export const type = new RequestType<ExecuteCommandParams, any, void, void>('workspace/executeClientCommand');
 }
 
-export namespace SendNotificationRequest {
-    export const type = new RequestType<ExecuteCommandParams, any, void, void>('workspace/notify');
+export namespace ServerNotification {
+    export const type = new NotificationType<ExecuteCommandParams, void>('workspace/notify');
 }
 
 export interface SourceAttachmentRequest {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -11,7 +11,7 @@ import { getDocumentSymbolsCommand, getDocumentSymbolsProvider } from './documen
 import * as requirements from './requirements';
 import { Commands } from './commands';
 import {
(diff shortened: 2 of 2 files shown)


src/protocol.ts
```diff
@@ -112,8 +112,8 @@ export namespace ExecuteClientCommandRequest {
     export const type = new RequestType<ExecuteCommandParams, any, void, void>('workspace/executeClientCommand');
 }
 
-export namespace SendNotificationRequest {
-    export const type = new RequestType<ExecuteCommandParams, any, void, void>('workspace/notify');
+export namespace SendNotificationNotification {
```
fbricon, line 115: ServerNotification

fbricon:
@testforstephen @yaohaizh @jdneo please check this is compatible with vscode-java-debug, as you guys introduced it.

src/protocol.ts
angelozerr: fixed.

testforstephen:
See https://github.com/microsoft/vscode-java-test/issues/260 and https://github.com/microsoft/vscode-java-test/pull/309/files, Test runner is consuming this api. @jdneo  Could you verify whether this break the test runner?

jdneo:
The usage of `sendNotification()` API was removed long time ago, This PR won't break the Java Test Runner.

Issue #641, opened 2018-09-14
Refactor when moving files


fwcd:
Automatically adjust imports and package declarations when moving a class in the file explorer. (As implemented in eclipse/eclipse.jdt.ls#683 ?)

#### Environment
- Operating System: `Windows 10`
- JDK version: `1.8.0_144`
- Visual Studio Code version: `1.27.2`
- Java extension version: `0.30.0`

#### Steps To Reproduce
1. Move a Java source file to another folder in the explorer

#### Current Result
1. The imports in other source files are not updated

#### Expected Result
1. The imports should be updated accordingly

julianosam:
That's literally the one thing that's stopping this (amazing) tool to become the #1 choice for Java development. I am the only developer in a project of near 100 Java developers and I am the only one using VSCode for Java. People are very impressed when I show my Java code in VSCode and also skeptical. So they ask "can it do this?" or "I bet if you try this it won't let you...". And refactoring is a key point. I confess I've been relying on Eclipse to more complex code refactoring... It'd be great to have this incorporated.

SpenceUK:
Automatically fixing imports across the project and package declarations would be a killer feature for Java in vscode.

testforstephen:
See https://github.com/microsoft/vscode/issues/43768, vscode will fire an FileWillRenameEvent when the user is renaming or moving files in file explorer. By listening at this event, the language server is able to update package declaration and references.

Currently the change is proposed at vscode insider. When it's officially ready, it will unblock us on the refactoring features such as move/rename file/rename package name.

nistvan86:
What is the current status on this? There's some work merged here https://github.com/microsoft/vscode/pull/85830
but I see an open ticket for a language server protocol extension https://github.com/microsoft/language-server-protocol/issues/873

liudonghua123:
I also like this feature

noel-yap:
Is there a way I can get notification for when this fix is released?

fwcd:
@noel-yap You can subscribe to this issue and choose to only get notified once it closes.

Sayyiditow:
Hi any update on this?

PR #1372, opened 2020-04-07
update its class name and references after filename renamed

testforstephen:

it requires https://github.com/eclipse/eclipse.jdt.ls/pull/1406

### DIFF
--- a/src/fileEventHandler.ts
+++ b/src/fileEventHandler.ts
@@ -1,10 +1,11 @@
 'use strict';
 
 import * as path from 'path';
-import { workspace, FileCreateEvent, ExtensionContext, window, TextDocument, SnippetString, commands, Uri } from 'vscode';
+import { workspace, FileCreateEvent, ExtensionContext, window, TextDocument, SnippetString, commands, Uri, FileRenameEvent, ProgressLocation } from 'vscode';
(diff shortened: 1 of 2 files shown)


testforstephen:
rename

fbricon:
@testforstephen so while this only works for a simple file rename within the same directory, any chance we can fix #641, i.e. when moving file to a different location?

testforstephen:
Have a look at the moving case, but there is still some problem needed to fix. 

Since the file event api at the client side will listen at file explorer operation and workspace edit, that means the file move event is probably from two sources: drag and drop at file explorer, move a CU through move refactoring. For the second operation, it's unnecessary to re-handle it in file event handler. Currently VS Code didn't provide the event source yet (there is a feature request for it https://github.com/microsoft/vscode/issues/93470), the extension itself has to find a way to figure out whether to handle the event.

In the onDidRenameFiles event, we can do that through comparing the package declaration and file location. But the upstream JDT move refactoring implementation won't work in the didRenameFiles, because it requires the references update calculation to be done before the actual move happens. I'm still trying to modify the jdt code to see whether it can reach what you want.

On the other side, it's hard for the extension to detect the event source in the onWillRenameFiles event. If the vscode feature request is ready, we can handle the move case in the onWillRenameFiles event.

PR #1375, opened 2020-04-10
preview the updates before apply the changes caused by rename file

testforstephen:

### DIFF
--- a/src/fileEventHandler.ts
+++ b/src/fileEventHandler.ts
@@ -1,11 +1,12 @@
 'use strict';
 
 import * as path from 'path';
-import { workspace, FileCreateEvent, ExtensionContext, window, TextDocument, SnippetString, commands, Uri, FileRenameEvent, ProgressLocation } from 'vscode';
-import { LanguageClient } from 'vscode-languageclient';
+import { workspace, FileCreateEvent, ExtensionContext, window, TextDocument, SnippetString, commands, Uri, FileRenameEvent, ProgressLocation, WorkspaceEdit as CodeWorkspaceEdit } from 'vscode';
+import { LanguageClient, WorkspaceEdit as LsWorkspaceEdit, CreateFile, RenameFile, DeleteFile, TextDocumentEdit } from 'vscode-languageclient';
 import { ListCommandResult } from './buildpath';
 import { Commands } from './commands';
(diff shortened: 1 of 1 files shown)


testforstephen:
previewRename

src/fileEventHandler.ts
```diff
@@ -173,3 +175,58 @@ async function isVersionLessThan(fileUri: string, targetVersion: number): Promis
 
     return javaVersion < targetVersion;
 }
+
+/**
+ * This function reference the implementation of asWorkspaceEdit() from 'vscode-languageclient/lib/protocolConverter'.
+ */
```
fbricon, line 181: anything safeguard we can add to check the preview updates are supported by the underlying client (eg. Theia)?

src/fileEventHandler.ts
```diff
@@ -98,7 +99,8 @@ async function handleRenameFiles(e: FileRenameEvent, client: LanguageClient) {
                 const edit = await client.sendRequest(DidRenameFiles.type, {
                     files: javaRenameEvents
                 });
-                const codeEdit = client.protocol2CodeConverter.asWorkspaceEdit(edit);
```
fbricon, line 101: keep the above line if preview refactoring is not supported

fbricon:
there's now a conflict in package.json

src/fileEventHandler.ts
testforstephen: This helper is safe even though the client doesn't support preview. Because the helper is almost same as the standard asWorkspaceEdit helper, except that it will append some additional [WorkspaceEditEntryMetadata](https://github.com/microsoft/vscode/blob/master/src/vs/vscode.d.ts#L2918) attribute to the existing WorkspaceEdit. It won't break the applyEdit API if your client doesn't recognize the new property.

I tried to the install a private vscode-java in an old **vscode-1.41** (**vscode-1.42 introduces the preview feature**), and it will apply the rename edit directly. But if i run it in vscode-1.42 and above, it will preview the rename edit first. I assume Theia would have the similar behavior as vscode-1.41 if it doesn't support preview.

src/fileEventHandler.ts
testforstephen: see the comment above.

PR #1394, opened 2020-04-16
extracted element can not be renamed

### LINKED ISSUE #1391: extracted element can not be renamed
Extracting a String to a field or a method, the generated element can not be renamed:
Failed rename

This is in vscode-java 0.59.1. This used to work since #333

snjeza:
Fixes #1391 

### DIFF
--- a/src/refactorAction.ts
+++ b/src/refactorAction.ts
@@ -113,6 +113,7 @@ async function applyRefactorEdit(languageClient: LanguageClient, refactorEdit: R
     }
 
     if (refactorEdit.command) {
+        await new Promise(resolve => setTimeout(resolve, 400));
         if (refactorEdit.command.arguments) {
             await commands.executeCommand(refactorEdit.command.command, ...refactorEdit.command.arguments);
         } else {


src/refactorAction.ts
```diff
@@ -109,6 +109,7 @@ async function applyRefactorEdit(languageClient: LanguageClient, refactorEdit: R
         const edit = languageClient.protocol2CodeConverter.asWorkspaceEdit(refactorEdit.edit);
         if (edit) {
             await workspace.applyEdit(edit);
+            await new Promise(resolve => setTimeout(resolve, 400));
```
testforstephen, line 112: `400` is the debounce interval at server side, do you use the await to make the changes is synced to the server, right?

i think you should move the await down a bit. Delay is only needed if the rename is executed immediately after the applyEdit operation.
```
    if (refactorEdit.edit) {
        const edit = languageClient.protocol2CodeConverter.asWorkspaceEdit(refactorEdit.edit);
        if (edit) {
            await workspace.applyEdit(edit);
        }
    }

    if (refactorEdit.command) {
        >>> await new Promise(resolve => setTimeout(resolve, 400));
        if (refactorEdit.command.arguments) {
            await commands.executeCommand(refactorEdit.command.command, ...refactorEdit.command.arguments);
        } else {
            await commands.executeCommand(refactorEdit.command.command);
        }
    }
```

snjeza: > 400 is the debounce interval at server side, do you use the await to make the changes is synced to the server, right?

Right.

> i think you should move the await down a bit. Delay is only needed if the rename is executed immediately after the applyEdit operation.

Done.

fbricon:
Nice @snjeza! thanks!

PR #1403, opened 2020-04-21
Refactor package name from file explorer

### LINKED ISSUE #823: Refactor package name
Is there a way to refactor the package name which will trigger a change in all the classes inside that package?
Couldn't find a way...

testforstephen:

Fix #823

This PR requires https://github.com/eclipse/eclipse.jdt.ls/pull/1414, it will rename the files and subpackages when renaming the package.

Rename package:
renamePackage

Rename subpackages:
renameSubPackage

### DIFF
--- a/package.json
+++ b/package.json
@@ -544,16 +544,18 @@
           "type": "string",
           "enum": [
             "Never",
-            "AutoApply",
-            "PreviewBeforeApply"
+            "AlwaysAutoApply",
+            "AlwaysPreview",
+            "Prompt"
           ],
           "enumDescriptions": [
             "Don't enable refactoring for rename operations on File Explorer.",
             "Always automatically update the imports and package declarations.",
-            "Always preview the changes before applying."
+            "Always preview the changes before applying.",
+            "Ask user to confirm whether to bypass refactor preview."
           ],
           "description": "Specifies whether to update imports and package declarations when renaming files from File Explorer.",
-          "default": "PreviewBeforeApply",
+          "default": "Prompt",
           "scope": "window"
         }
       }
--- a/package.json
+++ b/package.json
@@ -543,10 +543,10 @@
         "java.refactor.renameFromFileExplorer": {
           "type": "string",
           "enum": [
-            "Never",
-            "AlwaysAutoApply",
-            "AlwaysPreview",
-            "Prompt"
+            "never",
+            "autoApply",
+            "preview",
+            "prompt"
           ],
           "enumDescriptions": [
             "Don't enable refactoring for rename operations on File Explorer.",
@@ -555,7 +555,7 @@
             "Ask user to confirm whether to bypass refactor preview."
           ],
           "description": "Specifies whether to update imports and package declarations when renaming files from File Explorer.",
-          "default": "Prompt",
+          "default": "prompt",
           "scope": "window"
         }
       }
--- a/package.json
+++ b/package.json
@@ -539,6 +539,22 @@
           "description": "Enable/disable a warning about the impending requirement of Java 11.",
           "default": true,
           "scope": "window"
+        },
+        "java.refactor.renameFromFileExplorer": {
+          "type": "string",
+          "enum": [
+            "Never",
+            "AutoApply",
+            "PreviewBeforeApply"
+          ],
+          "enumDescriptions": [
+            "Don't enable refactoring for rename operations on File Explorer.",
+            "Always automatically update the imports and package declarations.",
+            "Always preview the changes before applying."
+          ],
+          "description": "Specifies whether to update imports and package declarations when renaming files from File Explorer.",
+          "default": "PreviewBeforeApply",
+          "scope": "window"
         }
       }
     },
--- a/src/fileEventHandler.ts
+++ b/src/fileEventHandler.ts
@@ -89,8 +89,16 @@ async function handleNewJavaFiles(e: FileCreateEvent) {
     }
 }
 
+function isRenameRefactoringEnabled(): boolean {
+    return workspace.getConfiguration().get("java.refactor.renameFromFileExplorer") !== "Never";
+}
+
+function isPreviewRenameRefactoring(): boolean {
+    return workspace.getConfiguration().get("java.refactor.renameFromFileExplorer") === "PreviewBeforeApply";
+}
+
 async function handleWillRenameFiles(e: FileWillRenameEvent, client: LanguageClient) {
-    if (!serverReady) {
+    if (!serverReady || !isRenameRefactoringEnabled()) {
         return;
     }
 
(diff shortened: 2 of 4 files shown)


testforstephen:
Not yet. This is also what i want. We're discussing the UX improvement with the VS Code team, see https://github.com/microsoft/vscode/issues/94650

fbricon:
Is there a way to bypass the preview then, apply the changes directly?

testforstephen:
Here is the latest updates to improve the UX:
- Add a new user setting to bypass preview.
  `java.refactor.renameFromFileExplorer`: Specifies whether to update imports and package declarations when renaming files from File Explorer. Defaults to `PreviewBeforeApply`.
    - `Never`: Don't enable refactoring for rename operations on File Explorer.
    - `AutoApply`: Always automatically update the imports and package declarations.
    - `PreviewBeforeApply`: Always preview the changes before applying.

- When generating the previewable WorkspaceEdit, add a dummy needsConfirmation node and make all other changes not need confirm. This will ensure the preview panel can be triggered and all valid changes pre-selected.

image

List changes by group:
image

testforstephen:
Add a notification to ask user whether to always preview the changes.

image

fbricon:
So I got an issue when applying refactoring after clicking on the message. Happens with either choices:
Screen Shot 2020-04-29 at 11 57 41 PM

Screencast: https://streamable.com/eoyh9w

package.json
```diff
@@ -539,6 +539,24 @@
           "description": "Enable/disable a warning about the impending requirement of Java 11.",
           "default": true,
           "scope": "window"
+        },
+        "java.refactor.renameFromFileExplorer": {
+          "type": "string",
+          "enum": [
```
fbricon, line 545: never, autoApply, preview, prompt

src/fileEventHandler.ts
```diff
@@ -1,14 +1,24 @@
...
 import { Commands } from './commands';
-import { DidRenameFiles } from './protocol';
+import { DidRenameFiles, WillRenameFiles } from './protocol';
 import { Converter as ProtocolConverter } from 'vscode-languageclient/lib/protocolConverter';
 
 let serverReady: boolean = false;
+let pendingEditPromise: Promise<LsWorkspaceEdit>;
+
+const PREFERENCE_KEY = "java.refactor.renameFromFileExplorer";
+const enum Preference {
```
fbricon, line 16: Never = "never",
    AlwaysAutoApply = "autoApply",
    AlwaysPreview = "preview",
    Prompt = "prompt"

src/fileEventHandler.ts
```diff
@@ -111,10 +211,55 @@ async function handleRenameFiles(e: FileRenameEvent, client: LanguageClient) {
     });
 }
 
+function askForConfirmation() {
+    if (needsConfirmRenameRefactoring()) {
+        window.showInformationMessage("Do you want to always preview the changes for rename refactoring?",
```
fbricon, line 216: Do you want to automatically apply refactoring when renaming?

src/fileEventHandler.ts
```diff
@@ -111,10 +211,55 @@ async function handleRenameFiles(e: FileRenameEvent, client: LanguageClient) {
...
+            "Yes", "No").then(async (answer) => {
```
fbricon, line 217: "Refactor automatically, "Always preview changes"

fbricon: Can we also apply the changes once the "Refactor automatically" button is pressed?

testforstephen: Nice idea. Yes, we can.

We can use `commands.executeCommand("refactorPreview.apply")` to apply the selected changes in the preview panel. If the preview panel is already executed, then this command will do nothing.

testforstephen:
> So I got an issue when applying refactoring after clicking on the message. Happens with either choices:
> Screen Shot 2020-04-29 at 11 57 41 PM
> 
> Screencast: https://streamable.com/eoyh9w

Aha, i used to see this dialog before. It seems vscode is too strict when [detecting conflicts](https://github.com/microsoft/vscode/blob/15c02dca3a95c1515d112b0f1bc97b3090402fdb/src/vs/workbench/services/bulkEdit/browser/conflicts.ts), they will check file changes and model changes. i suspect they have issues for model changes detection.

That's why i added a 400ms wait in my previous commit. After 400ms wait, I rarely see this conflicts message box in my machine. But I have to admit, this doesn't guarantee that others won't see this problem.
```typescript
await new Promise((resolve) =>setTimeout(resolve, 400));
const codeEdit = asPreviewWorkspaceEdit(edit, client.protocol2CodeConverter, isPreviewRenameRefactoring(), "Rename updates", e.files);
workspace.applyEdit(codeEdit);
```

However, I opened an issue in the VS Code https://github.com/microsoft/vscode/issues/96654.

testforstephen:
**Update:**
The root cause for the error "Cannot apply refactoring because 'x' file has changed in the meantime." is from eclipse.jdt.ls itself.

When renaming a package or applyEdit, vscode will send **didOpen** notifications for these sub files to the language server. Because these files are out-of-sync compilation units, the eclipse.jdt.ls will refreshLocal and then recreate the compilation unit (via `pkg.createCompilationUnit()` api). However, the underlying implementation of this API `pkg.createCompilationUnit()`  actually will call `unit.save()` method to modify the disk file. That's why vscode detects conflicts happening. 

See the code of [DocumentLifeCycleHandler](https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/BaseDocumentLifeCycleHandler.java):
```java
public ICompilationUnit handleOpen(DidOpenTextDocumentParams params) {
    String uri = params.getTextDocument().getUri();
    ICompilationUnit unit = resolveCompilationUnit(uri);
    if (unit == null || unit.getResource() == null || unit.getResource().isDerived()) {
        return unit;
    }
    try {
        // The open event can happen before the workspace element added event when a new file is added.
        // checks if the underlying resource exists and refreshes to sync the newly created file.
        if (!unit.getResource().isAccessible()) {
            try {
                unit.getResource().refreshLocal(IResource.DEPTH_ONE, new NullProgressMonitor());
                if (unit.getResource().exists()) {
                    IJavaElement parent = unit.getParent();
                    if (parent instanceof IPackageFragment) {
                        IPackageFragment pkg = (IPackageFragment) parent;
                        unit = pkg.createCompilationUnit(unit.getElementName(), unit.getSource(), true, new NullProgressMonitor());
                    }
                }
            } catch (CoreException e) {
                // ignored
            }
        }
```

// @snjeza, the extra logic is introduced to fix a previous bug. https://github.com/eclipse/eclipse.jdt.ls/pull/486

fbricon:
Rebased and merged as https://github.com/redhat-developer/vscode-java/commit/e0490ac2da0e3ffa0ec9926211ef0269d1ae454d

Thanks @testforstephen!

SUMIT0733:
Hello, 
When I add java.refactor.renameFromFileExplorer in settings.json, it says "unknown configuration settings", is there anything I need to add other than add this settings in json file?

testforstephen:
No, this setting is removed by another PR https://github.com/redhat-developer/vscode-java/pull/1815.

PR #1418, opened 2020-04-29
Prompt the user for importing during initialization

jdneo:
@fbricon This is the PR to prompt the user for importing.

_// It contains the commit of this [PR](https://github.com/redhat-developer/vscode-java/pull/746) I'll do the rebase after #746 is merged._


jdneo:
Close because we have #1501

PR #1419, opened 2020-04-29
Prompt to enable Semantic Hightlighting on startup

fbricon:
Display this prompt when the `java.semanticHighlighting.enabled` preference has not be set explicitely:

Message links to https://github.com/redhat-developer/vscode-java/wiki/Semantic-Highlighting


fbricon:
FYI @akaroml  @Eskibear

Eskibear:
LGTM. 

But sample setting of color customization in Wiki page should be updated. The standard modifier "readonly" has been used instead of "final".

Issue #1416, opened 2020-04-29
Provide goToDefinition API 


KaihangXing:
#### [This is not an issue but a requirement]

I want to get the symbol's defination but not the Document defination through ExtensionAPI in another vscode plugin. Such as a method's class, package and parameters. 

I can't find suitable method in current ExtensionAPI. 

Could anyone help me?  Thanks a lot~

fbricon:
Sorry, this is currently not possible at the moment

KaihangXing:
> Sorry, this is currently not possible at the moment

I just commit a PR for this enhancement https://github.com/redhat-developer/vscode-java/pull/1425

fbricon:
@KaihangXing this will be part of the next release, mid-May 2020

PR #1425, opened 2020-04-30
Added goToDefinition call to extension API

KaihangXing:
For this enhancement: https://github.com/redhat-developer/vscode-java/issues/1416

At Alibaba we are working on a VSCode extension and in it we need finding a method's defination in Java files. Your extension has everything we need, but there just wasn't a way to call the language server you are using from outside the extension. 

This PR add function gotoDefinitionSymbols to the extension API, which can then be called by other extensions. This function performs textDocument/definition call to the language server and return a promise with the result.

If you have any questions just ask, or if this PR has incorrect style etc. then I can fix that, no problem.

Signed-off-by: kaihang.xkh <kaihang.xkh@alibaba-inc.com>

### DIFF
--- a/src/extension.api.ts
+++ b/src/extension.api.ts
@@ -1,4 +1,5 @@
 import { getDocumentSymbolsCommand } from './documentSymbols';
+import { goToDefinitionCommand } from './goToDefinition';
 import { RequirementsData } from './requirements';
 import { TextDocumentPositionParams } from 'vscode-languageclient';
 import { CancellationToken, Command, ProviderResult, Uri, Event } from 'vscode';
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -297,7 +299,8 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 									getProjectSettings,
 									getClasspaths,
 									isTestFile,
-									onDidClasspathUpdate
+									onDidClasspathUpdate,
+									gotoDefinitionCommand: gotoDefinitionSymbols
```
fbricon, line 303: why not simply `gotoDefinitionSymbols`?

src/extension.api.ts
```diff
@@ -77,4 +78,5 @@ export interface ExtensionAPI {
 	 *   3. The Uri points to the project root path.
 	 */
 	readonly onDidClasspathUpdate: Event<Uri>;
+	readonly gotoDefinitionCommand: gotoDefinitionCommand;
```
fbricon, line 81: goToDefinition: goToDefinitionCommand

fbricon:
Thanks for your contribution @KaihangXing!

PR #1435, opened 2020-05-06
return semantic tokens as Uint32Array

Eskibear:
currently it's returned as `number[]`, causing random failures of semantic highlighting.

See [vscode's API]( https://github.com/microsoft/vscode/blob/bd0efff9e3f36d6b3e1045cee9887003af8034d7/src/vs/vscode.d.ts#L3242) and https://github.com/microsoft/vscode/issues/96664#issuecomment-623311367


fbricon:
Thanks @Eskibear

PR #1454, opened 2020-05-15
Enable refactor when moving files

testforstephen:

It requires https://github.com/eclipse/eclipse.jdt.ls/pull/1445

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,8 +158,8 @@ The following settings are supported:
 * `java.sources.organizeImports.starThreshold`: Specifies the number of imports added before a star-import declaration is used, default is 99.
 * `java.sources.organizeImports.staticStarThreshold`: Specifies the number of static imports added before a star-import declaration is used, default is 99.
 * `java.semanticHighlighting.enabled`: Enable/disable [Semantic Highlighting](https://github.com/redhat-developer/vscode-java/wiki/Semantic-Highlighting) for Java files. Defaults to `true`.
-* `java.refactor.renameFromFileExplorer`: Specifies whether to update imports and package declarations when renaming files from File Explorer. Defaults to `prompt`.
-  - `never`: Don't enable refactoring for rename operations on File Explorer.
+* `java.refactor.renameFromFileExplorer`: Specifies whether to update imports and package declarations when you rename or move files from File Explorer. Defaults to `prompt`.
+  - `never`: Don't enable refactoring for rename and move operations on File Explorer.
   - `autoApply`: Always automatically update the imports and package declarations.
   - `preview`: Always preview the changes before applying.
   - `prompt`: Ask user to confirm whether to bypass refactor preview.
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,12 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@types/debounce-promise": {
+      "version": "3.1.2",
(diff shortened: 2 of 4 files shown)


fbricon:
I got pretty inconsistent results using this feature.
- one time I got the "file was already modified" error after clicking on the autoapply button
- most of the time, I get stale error diagnostics on modified files. Needs to save them so they disappear
- moving files once more, looks like package refactoring didn't work

testforstephen:
> one time I got the "file was already modified" error after clicking on the autoapply button

I also got this error several times. After a look at the verbose log, i found there are didChange events occurring after files moved. That's the cause of the conflicts. I'm still in digging the main culprit.

testforstephen:
> one time I got the "file was already modified" error after clicking on the autoapply button

I find out this is a vscode bug. Open an issue https://github.com/microsoft/vscode/issues/98057 in upstream. When moving files back and forth several times, there are stale text models in memory and when you move back, vscode tries to sync the stale model so as to trigger didChange events in background.

fbricon:
@testforstephen looks like upstream bug has been fixed, maybe time to revisit this issue

testforstephen:
Sure. Let me check. i remember the upstream still has some issue when moving the unsaved files to new location.

The original issue mentioned above is about there is dirty text model hanging in memory when the file is deleted or moved. It's fixed.

A new issue is https://github.com/microsoft/vscode/issues/100255. When moving an unsaved file to new location, VS Code will close the old file first, then open a new file, and trigger a `didChange` event to restore the unsaved contents. The extra `didChange` event may cause conflicts when we're applying the previewed workspace edit to the new location.

I will check again to see whether there is workaround for this. Also ping the upstream again about the progress.

testforstephen:
Rebased the code, here is a private vsix https://raw.githubusercontent.com/testforstephen/vscode-test/master/java-0.69.1.vsix, welcome to have a try with the latest **vscode insider**.
move


> - one time I got the "file was already modified" error after clicking on the autoapply button

 i didn't meet this problem in the latest insider.
> - most of the time, I get stale error diagnostics on modified files. Needs to save them so they disappear

The DiagnosticsHandler in jdt.ls will refresh the diagnostics for all existing working copies every time there are documents didChange. I didn't observe the stale error diagnostics on modified files (they are in working copies) after applying the edit. But it's possible to get stale diagnostics on the unmodified files, because the DiagnosticsHandler in DocumentLifeCycleHandler only refresh workingcopies. Need save the files to trigger auto build.

> - moving files once more, looks like package refactoring didn't work

I didn't observe this problem in spring-petclinic project, but it probably can happen in some more complex project. The reason is VS Code only gives a default 5 seconds to the extension to compute the refactoring in the onWillRenameFile event listener. If your project is complex, or your machine is not powerful, and the move refactoring computation may not get result in the default time limit. Need manually increase the timeout setting `files.participants.timeout`.

testforstephen:
> > - moving files once more, looks like package refactoring didn't work

> I didn't observe this problem in spring-petclinic project, but it probably can happen in some more complex project. The reason is VS Code only gives a default 5 seconds to the extension to compute the refactoring in the onWillRenameFile event listener. If your project is complex, or your machine is not powerful, and the move refactoring computation may not get result in the default time limit. Need manually increase the timeout setting `files.participants.timeout`.

Opened an issue https://github.com/microsoft/vscode/issues/111208 to discuss the 5-second timeout issue.

testforstephen:
I'll send a new pull request to rework the rename/move refactoring feature.

PR #1456, opened 2020-05-17
Update the API's comments

jdneo:
require: https://github.com/eclipse/eclipse.jdt.ls/pull/1446


jdneo:
Hi @fbricon, just to remind that this requires PR in JDT.LS: https://github.com/eclipse/eclipse.jdt.ls/pull/1446

Issue #1474, opened 2020-05-28
[Feature request] Add getCompletionItems to extensionAPI


HaoboGu:
Hello, 

I am an extension developer, I am writing a code auto-completion extension for Java. 

vscode-java is a great extension which provides amazing completion results. But those result cannot be used by other extensions. 

Can you please add getCompletionItems API so that other extensions can use it for providing better completion service?

Or, is there any way to contribute this feature? 

Thanks if you can help!

fbricon:
Can you explain your use case in more details?

HaoboGu:
> Can you explain your use case in more details?

I am writing an auto-complete extension. The extension would generate some completionItems, I want to re-rank my list by combining with the result given by your extension. Besides, I also want to enrich the documentation of my list using the doc/detail field of the result from your extension.

fbricon:
Would your autocompletion be triggered inside java files? My concern is potential conflicts between several java completion invocations

HaoboGu:
> Would your autocompletion be triggered inside java files? 

Yes

> My concern is potential conflicts between several java completion invocations

I will remove the duplicate results from my list. And VSCode would do the merging of results from multiple extensions, I think it's not a risk.

fbricon:
I don't understand, why do you want to have the same completion triggered twice for Java files, only to filter out the second set of results?

fbricon:
my point is Java completion requests are pretty expensive. Doing it twice will certainly result in a much degraded UX

HaoboGu:
> I don't understand, why do you want to have the same completion triggered twice for Java files, only to filter out the second set of results?

It's not only to do filtering, the result also helps to enhance and rerank my list, which is different from the list from your extension. 

For example, if there are two results from my extension with leading `re`:
```
1. reader.Close();
2. return null;
```
And, if I found in your list, `reader` exists but `return` not, I would give `reader.Close()` a higher rank.

The results from my extension will have no effect on vscode-java's result. 

What's more, if we can request once and then do both things(I mean showing the raw results in vscode and pass them to my extension), it would be great!

PR #1478, opened 2020-06-01
Add provideCompletionItems to extension API

HaoboGu:
This PR adds `provideCompletionItems` call to extension API. 

This PR also fixes the this issue: #1474 

Signed-off-by: Haobo Gu <haobogu@outlook.com>


src/extension.api.ts
```diff
@@ -80,4 +81,5 @@ export interface ExtensionAPI {
 	 */
 	readonly onDidClasspathUpdate: Event<Uri>;
 	readonly goToDefinition: goToDefinitionCommand;
+	readonly provideCompletionItems: provideCompletionItemsCommand;
```
fbricon, line 84: getCompletionItems: getCompletionItemsCommand

src/extension.ts
```diff
@@ -307,7 +309,8 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 									getClasspaths,
 									isTestFile,
 									onDidClasspathUpdate,
-									goToDefinition: goToDefinition
+									goToDefinition: goToDefinition,
+									provideCompletionItems: provideCompletionItems,
```
fbricon, line 313: getCompletionItems: getCompletionItems

src/extension.ts
```diff
@@ -324,7 +327,8 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
```
fbricon, line 331: getCompletionItems: getCompletionItems

src/extension.ts
```diff
@@ -35,6 +35,7 @@ import { registerClientProviders, ClientHoverProvider } from './providerDispatch
 import * as fileEventHandler from './fileEventHandler';
 import { registerSemanticTokensProvider } from './semanticTokenProvider';
 import { runtimeStatusBarProvider } from './runtimeStatusBarProvider';
+import { provideCompletionItemsCommand, completionItemsProvider } from './provideCompletionItems';
```
fbricon, line 38: import {getCompletionItemsCommand, completionItemsProvider} from './completionItems.ts'

src/extension.ts
```diff
@@ -265,6 +266,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 				languageClient.registerProposedFeatures();
 				const getDocumentSymbols: getDocumentSymbolsCommand = getDocumentSymbolsProvider(languageClient);
 				const goToDefinition: goToDefinitionCommand = goToDefinitionProvider(languageClient);
+				const provideCompletionItems: provideCompletionItemsCommand = completionItemsProvider(languageClient);
```
fbricon, line 269: const getCompletionItems: getCompletionItemsCommand

src/provideCompletionItems.ts
```diff
@@ -0,0 +1,22 @@
+'use strict';
```
fbricon, line 1: completionItems.ts

src/provideCompletionItems.ts
```diff
@@ -0,0 +1,22 @@
...
+	CancellationToken,
+	CompletionList,
+	CompletionItem,
+	LanguageClient,
+	CompletionParams,
+	CompletionRequest,
+} from 'vscode-languageclient';
+
+type CompletionResponse = CompletionList | CompletionItem[] | null;
+export type provideCompletionItemsCommand = (params: CompletionParams, token?: CancellationToken) => Promise<CompletionResponse>;
```
fbricon, line 13: export type getCompletionItemsCommand

jdneo:
Just tried in `Standard` and `LightWeight` mode. Since nowadays, the light weight server does not have the capability of completion, I got an error when sending the completion request to the server:

```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.request(GenericEndpoint.java:120)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleRequest(RemoteEndpoint.java:261)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:190)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:194)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor...
```

testforstephen:
I don't think it's a good idea to expose this API to ts side. Because the completion handler in server side already did some ranking in advance and then returned the top 50 results to the client. I highly doubt your re-rank algorithm will work in the cooked 50 results.

HaoboGu:
@testforstephen I won't change the results from language server -- I just use them as the evidence to improve my sequential CompletionItems. 

Because my completion model might give some out-of-topic or illegal results, context-aware results from language server would be quite helpful for filtering and rerank my CompletionItems

jdneo:
@HaoboGu Just interested, would you mind to share some data points about the performance of the completion in your implementation?

As @fbricon has said in the original issue, resolving completion lists is an expensive operation. Users may get confused if they found the slowness for resolving the completion list -- which makes this API useless if it's not useable because of the performance downgrade.

HaoboGu:
@jdneo Sure. I am implementing an ensemble auto-completion model, which contains a deep model and a local model. 

The performance of deep model depends on the file size. For a file which contains 4000 lines, calculating completionItems costs 200-300ms. I also set a 300ms timeout. We'll keep working on improving the performance of deep model. 

As for the local model, the completion costs 20-50ms on average. I believe this won't have negative effects on user experience.

jdneo:
Thank you @HaoboGu for the sharing. What I'm afraid is not your deep/local model but the API itself. I would suggest you to try to integrate this API locally and see the performance of completion. By default, it will return 50 items. But when you set the setting `java.completion.maxResults` to 0 it will return a full list, which will cost more times (before you can do real the ranking actually).

HaoboGu:
@jdneo I just tested the performance of the API on a file with 4000 lines in a large project(more than 3000 java files). `java.completion.maxResults` was set to 0.

In the first few completionRequests, the API is quite slow. It took about 5s to fetch results from the API. 

And I got an error in **Language Support for Java** output channel:
```
An internal error occurred during: "Publish Diagnostics".
Java heap space
java.lang.OutOfMemoryError: Java heap space
```

After the first few requests, the performance of the API was acceptable:  the time cost of fetching results from added API was from 200ms to 600ms, the overall completion call cost less than 1s.

jdneo:
@HaoboGu Thank you for those data points, the project and the idea look very cool and I like it.

While, as you can see for the performance results, in worst case it will take 5s to get the full list. Even the average time is less than 1s, considering the completion is a very frequent action, the experience is still not that good. So, I think we can export this API as long as either one of the following conditions meet:
- The performance of the completion get boosted in VS Code Java, then we are confident to export this API.
- More requests for this APIs from different users.

Anyway, it's just my personal opinion, @fbricon @testforstephen please add your comments if you have.

HaoboGu:
@jdneo Thanks for the comment. 

I just did more tests and I can confirm that the performance issue on large project still exists even though I disabled all other extensions except vscode-java. 

I agree to suspend this PR until eclipse.jdt addresses the problem.

HaoboGu:
See  [eclipse.jdt.ls#1469](https://github.com/eclipse/eclipse.jdt.ls/issues/1469), the performance issues was addressed after I removed the `onDidOpenTextDocument` watcher. 

According to my test, now the API works great, [here](http://haobo-markdown.oss-cn-zhangjiakou.aliyuncs.com/markdown/2020-06-18-example-1.gif) is an example. You can see that with eclipse.jdt.ls, `getElementType()` ranks higher than `getTypeName()`, which is not legal. 

 Thank you all for the help @jdneo @testforstephen @fbricon 

Can this PR be merged right now?

fbricon:
I'm still very, very cagey about this PR. I believe triggering the completion twice is inherently a wrong idea. 
The Java IntelliCode extension is capable of changing the completion ranking using a different approach. @jdneo @testforstephen @Eskibear can correct me if I'm wrong, but, if I remember correctly, a custom jdt.ls extension [registers](https://github.com/eclipse/eclipse.jdt.ls/blob/c2bdb489d305ece226da4e55f6e349d65ce34383/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/JDTLanguageServer.java#L222) another completionProvider, that basically calls a fork of https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/CompletionHandler.java, then calls [getProposals()](https://github.com/eclipse/eclipse.jdt.ls/blob/c2bdb489d305ece226da4e55f6e349d65ce34383/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/contentassist/CompletionProposalRequestor.java#L390-L392)  to properly rank the results against the complete list of proposals. See https://github.com/eclipse/eclipse.jdt.ls/pull/1344.

Eskibear:
Exactly as @fbricon metioned above, IntelliCode does below things:
* turn off default completion
```java
Unregistration unregistration = new Unregistration(Preferences.COMPLETION_ID, Preferences.COMPLETION);
UnregistrationParams unregistrationParams = new UnregistrationParams(Collections.singletonList(unregistration));
clientConnection.unregisterCapability(unregistrationParams);
```
* register its own via vscode's API `vscode.languages.registerCompletionItemProvider`
* fetches list of CompletionItems from the language server, through a [DelegateCommand](https://github.com/eclipse/eclipse.jdt.ls/blob/cf6358ba4c166fa7d4c45aba5b7b22ea08516524/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/IDelegateCommandHandler.java#L25).
* re-rank them by changing the `sortText`

jdneo:
Hey @HaoboGu,

Strongly recommend to take a look at the approach @Eskibear has mentioned. I think that is much better than simply exposing the completion in the APIs.

Eskibear:
> a custom jdt.ls extension registers another completionProvider
Yes, it would be better to do all in the jdt.ls extension, then you won't bother to convert them between LSP and VSCode. 

Anyway, it doesn't make sense to me to expose such API in this extension, because triggering the completion twice is a wrong idea.

HaoboGu:
> * fetches list of CompletionItems from the language server, through a [DelegateCommand](https://github.com/eclipse/eclipse.jdt.ls/blob/cf6358ba4c166fa7d4c45aba5b7b22ea08516524/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/IDelegateCommandHandler.java#L25).

@Eskibear Could you provide more details about how to do this?  In my understanding, I have to write a custom `eclipse.jdt.ls` and extend `DelegateCommand` to pass the completion request to my own completion server. Then, I can resort the result in custom `eclipse.jdt.ls`. Am I right?

Eskibear:
a custom [jdt.ls extension ](https://github.com/eclipse/eclipse.jdt.ls/wiki/Contribute-an-extension-bundle#create-an-jdtls-extension)is what you want, follow https://github.com/redhat-developer/vscode-java/pull/1478#issuecomment-646505044

HaoboGu:
The IntelliCode's solution seems a better solution than exposing completion API, which leads to redundant completion request. So I'm going to close this PR.

Thank you all @jdneo @fbricon @Eskibear @testforstephen for the help and suggestions. Learned a lot from the discussion above :D

HaoboGu:
Hi guys, I just started to implement a jdt.ls extension according to previous discussion. However, I could not set up the extension project, using [jdt.ls-extension-archetype](https://github.com/open-archetypes/jdt.ls-extension-archetype). 
When I executed `mvn clean verify`, maven reports the following error:
```
[ERROR] Internal error: org.eclipse.tycho.core.osgitools.OsgiManifestParserException: Exception parsing OSGi MANIFEST /Users/haobogu/.m2/repository/p2/osgi/bundle/org.eclipse.jdt.ls.core/0.60.0.202008140759/org.eclipse.jdt.ls.core-0.60.0.202008140759.jar!/META-INF/MANIFEST.MF: Unknown OSGi execution environment: 'JavaSE-11' -> [Help 1]
org.apache.maven.InternalErrorException: Internal error: org.eclipse.tycho.core.osgitools.OsgiManifestParserException: Exception parsing OSGi MANIFEST /Users/haobogu/.m2/repository/p2/osgi/bundle/org.eclipse.jdt.ls.core/0.60.0.202008140759/org.eclipse.jdt.ls.core-0.60.0.202008140759.jar!/META-INF/MANIFEST.MF: Unknown OSGi execution environment: 'JavaSE-11'
```
@Eskibear do you have any idea about this? Thank you very much!

Issue #1480, opened 2020-06-02
Export public API in LightWeight mode


jdneo:
### Background
We recently introduced [`LightWeight` mode](https://github.com/redhat-developer/vscode-java/pull/1329) in VS Code Java. The `LightWeight` mode is used for those who just want to do some 'easy' work in VS Code, for example: reading the code...

But the users might find that if they take a certain action (i.e. starting a debug session) in `LightWeight` mode, nothing happens...

### Root Cause
In the `activate()` method of VS Code Java, we will return the public APIs as a promise, which will only be resolved when the [`Standard` language server is ready](https://github.com/redhat-developer/vscode-java/blob/master/src/extension.ts#L300). That means the promise will never be resolved if VS Code Java is in `LightWeight` mode. Thus, the extension will be in an inactivated status (You can tell it by triggerring the command **Developer: Show Running Extensions**). All the third-party extensions which depend on VS Code Java will then, keep waiting.

### Proposed change
Also resolve the public API in whatever server mode, besides that, adding the following new API:

```jsonc
{
  serverMode: ServerMode
  onDidServerModeChange: Event<ServerMode>
}
```

**Note:** 
- The APIs like: `getProjectSettings`, `getClasspaths`, `isTestFile` and `onDidClasspathUpdate` will return an **empty-like** value in `LightWeight` mode, since so far there is no scenarios shows these APIs need to work in `LightWeight` mode. (We can change the impelmentation when necessary)
- In `LightWeight` mode, all the calls to the command: `java.execute.workspaceCommand` will just show a warning message in the console and directly return.
- It's recommended for all the the third-party extensions to consume the `serverMode` field in the APIs set and change the behavior if it's `LightWeightServiceReady`.

### How to adopt the new APIs
Check the serverMode when get the resolved extension API, and run properly. For example, you may not want to provide some features when Java Language Server is in lightweight mode.

Also do not forget to register a listener to listen the server mode changing event:

```ts
const extension: Extension<any> | undefined = extensions.getExtension('redhat.java');
if (extension && extension.isActive) {
    const extensionApi: any = extension.exports;
    if (!extensionApi) {
        return;
    }

    if (extensionApi.onDidChangeServerMode) {
        const onDidChangeServerMode: Event<string> = extensionApi.onDidChangeServerMode;
        context.subscriptions.push(onDidChangeServerMode(async (mode: string) => {
            // do something according to current server mode
        }));
    }
}
```

[...]

PR #1484, opened 2020-06-08
Export the public APIs in light weight mode

jdneo:
Address #1480 

> Note: The cases for lightweight mode could not run currently because the syntax server needs to send started signal to the client, see: https://github.com/eclipse/eclipse.jdt.ls/pull/1472


fbricon:
When started in standard mode, I get this command:
Screen Shot 2020-06-08 at 12 21 41 PM

fbricon:
Travis fails on both mac and linux

src/commands.ts
```diff
@@ -200,4 +200,8 @@ export namespace Commands {
      * Temporary command to fetch Semantic Tokens Legend. To remove when LSP v3.16 is ready.
      */
     export const GET_SEMANTIC_TOKENS_LEGEND = 'java.project.getSemanticTokensLegend';
+    /**
+     * Command to switch between standard mode and lightweight mode.
+     */
+    export const SWITCH_SERVER_MODE = 'java.server.mode.switch';
```
fbricon, line 206: this is not accurate. The command only says "switch to Standard" when you're in lightweight mode. There should be a similar switch to LightWeight, when in standard mode

jdneo:
@fbricon Yes I should hide the 'switch' command when it's in standard mode.

I guess the travis fails because it's not using the latest bits including https://github.com/eclipse/eclipse.jdt.ls/pull/1472. Without the lightweight server `started` notification to clients, the public APIs never get resolved in `activate()` methods. Thus VS Code keeps waiting for VS Code Java to start up.

src/commands.ts
jdneo: The current implementation only supports switch from `LightWeight` to `Standard` once. Because so far I cannot find out a valid scenario that the users need to switch back to `LeightWeight`.

But I named the command without specifying the mode in case in the future we need to reuse this command to switch back and forth. A param: `mode: ServerMode` could be added to it then.

src/apiManager.ts
```diff
@@ -0,0 +1,94 @@
...
+            javaRequirement: requirements,
+            status: null,
+            registerHoverCommand,
+            getDocumentSymbols,
+            goToDefinition,
+            getProjectSettings,
+            getClasspaths,
+            isTestFile,
+            onDidClasspathUpdate,
+            serverMode: getJavaServerMode() === ServerMode.STANDARD ? ServerMode.STANDARD : ServerMode.LIGHTWEIGHT,
```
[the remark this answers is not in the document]
testforstephen: serverMode here is a little different with the serverMode user setting. It looks more like the active server which is serving language features, right? For example, in Hybrid mode, at first it's lightweight mode serve features, then switched to standard mode.

src/extension.ts
```diff
@@ -536,30 +266,48 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+
+			const snippetProvider: SnippetCompletionProvider = new SnippetCompletionProvider();
+			context.subscriptions.push(languages.registerCompletionItemProvider({ scheme: 'file', language: 'java' }, snippetProvider));
+
+			registerClientProviders(context, { contentProviderEvent: jdtEventEmitter.event });
+
+			fileEventHandler.registerFileEventHandlers(standardClient.getClient(), context);
 
+			apiManager.getApiInstance().onDidChangeServerMode((event: ServerMode) => {
+				if (event === ServerMode.STANDARD) {
```
testforstephen, line 304: When do you stop lightweight server in Hybrid mode? I didn't see the logic for that.

src/extension.ts
```diff
@@ -217,315 +205,57 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+			if (requireStandardServer) {
+				standardClient.start();
+			}
 
-				languageClient.onReady().then(() => {
-					languageClient.onNotification(StatusNotification.type, (report) => {
+			// TODO: Currently only resolve the promise when the server mode is explicitly set to lightweight.
+			// This is to avoid breakings
+			if (serverMode === ServerMode.LIGHTWEIGHT) {
+				const lightWeightClient: LanguageClient = syntaxClient.getClient();
```
testforstephen, line 238: how about moving the notification callback to syntaxLanguageClient, just like what the standardLanguageClient did.

src/providerDispatcher.ts
```diff
@@ -34,23 +33,26 @@ export function registerClientProviders(context: ExtensionContext, options: Prov
...
+		const serverMode: ServerMode = apiManager.getApiInstance().serverMode;
+		if (serverMode === ServerMode.LIGHTWEIGHT) {
 			const params = {
 				textDocument: languageClient.code2ProtocolConverter.asTextDocumentIdentifier(document),
 				position: languageClient.code2ProtocolConverter.asPosition(position)
 			};
 			const hoverResponse = await languageClient.sendRequest(HoverRequest.type, params);
 			return languageClient.protocol2CodeConverter.asHover(hoverResponse);
-		} else if (this.delegateProvider) {
+		} else if (serverMode === ServerMode.STANDARD) {
```
testforstephen, line 55: No need to expose `initializeForStandardServer` to external, you can just lazy initialize it if delegateProvider is undefined.

testforstephen:
```typescript
registerHoverCommand(callback: provideHoverCommandFn) {
	if (this.delegateProvider) {	
		this.delegateProvider.registerHoverCommand(callback);				
	}
}		
```

Besides, you can decouple `delegateProvider.registerHoverCommand(callback)` to a standalone utility. What it does is saving the callbacks into an array, and it doesn't invoke the language client.

src/standardLanguageClient.ts
```diff
@@ -0,0 +1,364 @@
...
+
+const extensionName = 'Language Support for Java';
+const GRADLE_CHECKSUM = "gradle/checksum/prompt";
+
+export class StandardLanguageClient {
+
+	private languageClient: LanguageClient;
+	private initialized: boolean = false;
+
+	public initialize(context: ExtensionContext, requirements: RequirementsData, clientOptions: LanguageClientOptions, workspacePath: string, jdtEventEmitter: EventEmitter<Uri>, resolve): void {
```
testforstephen, line 34: resolve: (value: ExtensionAPI) => void

src/apiManager.ts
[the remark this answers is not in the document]
jdneo: call it `activeServer`?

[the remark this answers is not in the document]
testforstephen: Oh it represents the final server that is alive when the java extension is resolved. I'm OK with the current name, but you'd better document its meaning in API doc.

src/extension.api.ts
```diff
@@ -80,9 +81,18 @@ export interface ExtensionAPI {
 	 */
 	readonly onDidClasspathUpdate: Event<Uri>;
 	/**
-	 * An event fires on projects imported.
+	 * An event fires on projects imported. This API is not supported in light weight server mode so far.
 	 * The Uris in the array point to the project root path.
 	 */
 	readonly onDidProjectsImport: Event<Uri[]>;
 	readonly goToDefinition: goToDefinitionCommand;
+	serverMode: ServerMode;
```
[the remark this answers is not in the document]
jdneo: It's in an interface, no need to add `let`

src/providerDispatcher.ts
jdneo: The `createClientHoverProvider` needs the `ExtensionContext` as a param to register a command (go to super implementation). Do we also need to move that command registration to `standardLanguageClient` to get rid of the `context` param in `createClientHoverProvider`?

src/standardLanguageClient.ts
```diff
@@ -85,8 +85,10 @@ export class StandardLanguageClient {
 			this.languageClient.onNotification(StatusNotification.type, (report) => {
 				switch (report.type) {
 					case 'ServiceReady':
-						apiManager.updateServerMode(ServerMode.STANDARD);
-						apiManager.emitDidChangeServerMode(ServerMode.STANDARD);
+						if (apiManager.getApiInstance().serverMode !== ServerMode.STANDARD) {
+							apiManager.updateServerMode(ServerMode.STANDARD);
+							apiManager.fireDidServerModeChange(ServerMode.STANDARD);
+						}
```
testforstephen, line 91: when in Hybrid mode, the initialize value of apiManager.getApiInstance().serverMode is already Standard. So it seems never trigger didServerModeChange event and no chance to stop lightweight server, right?

src/apiManager.ts
```diff
@@ -0,0 +1,96 @@
...
+import { commands, Uri } from "vscode";
+import { Commands } from "./commands";
+import { Emitter } from "vscode-languageclient";
+import { ServerMode, getJavaServerMode } from "./settings";
+import { registerHoverCommand } from "./hoverAction";
+
+class ApiManager {
+
+    private api: ExtensionAPI;
+    private onDidClasspathUpdateEmitter: Emitter<Uri> = new Emitter<Uri>();
```
testforstephen, line 16: This variable has Emitter as suffix, but other variables don't have. Do you want to make them consistent?

src/extension.ts
```diff
@@ -571,11 +293,28 @@ export function getJavaConfig(javaHome: string) {
...
+export function deactivate(): void {
+	standardClient.stop();
+	syntaxClient.stop();
+}
+
+export async function getActiveLanguageClient(): Promise<LanguageClient | undefined> {
+	let languageClient: LanguageClient;
+
+	const api: ExtensionAPI = apiManager.getApiInstance();
+	if (api.serverMode === ServerMode.LIGHTWEIGHT) {
```
testforstephen, line 305: In hybrid mode, the apiManager.getApiInstance() always returns standard, no chance to return syntax client to serve features. Looks a breaking change.

src/apiManager.ts
```diff
@@ -0,0 +1,96 @@
...
+        this.api.serverMode = mode;
+    }
+
+    public updateStatus(status: "Started" | "Error"): void {
+        this.api.status = status;
+    }
+
+    private initializeServerMode(): ServerMode {
+        const serverLaunchMode: ServerMode = getJavaServerMode();
+        if (serverLaunchMode === ServerMode.HYBRID) {
```
testforstephen, line 87: I think you should return hybrid to api directly. Only when standard server is ready, then change status to standard.

package.json
```diff
@@ -729,6 +738,14 @@
...
           "command": "java.project.removeFromSourcePath",
           "when": "false"
+        },
+        {
+          "command": "java.show.server.task.status",
+          "when": "serverMode != LightWeight"
+        },
+        {
+          "command": "java.server.mode.switch",
+          "when": "serverMode != Standard"
```
testforstephen, line 748: serverMode == Lightweight

src/providerDispatcher.ts
```diff
@@ -34,60 +33,56 @@ export function registerClientProviders(context: ExtensionContext, options: Prov
...
+			return undefined;
 		}
-	}
 
-	async provideHover(document: TextDocument, position: Position, token: CancellationToken): Promise<Hover> {
-		if (this.options.syntaxClient && this.options.syntaxClient.isAlive()) {
-			const languageClient: LanguageClient = this.options.syntaxClient.getClient();
-			await languageClient.onReady();
+		const serverMode: ServerMode = apiManager.getApiInstance().serverMode;
+		if (serverMode === ServerMode.LIGHTWEIGHT) {
```
testforstephen, line 44: providerDispatcher is used to work as a load balancer to dispatch the request to the active client of hybrid mode. With this new change, looks like it doesn't work for hybrid any more.

src/extension.ts
```diff
@@ -190,12 +176,12 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
 					workspace: {
 						didChangeConfiguration: () => {
-							languageClient.sendNotification(DidChangeConfigurationNotification.type, {
+							standardClient.getClient().sendNotification(DidChangeConfigurationNotification.type, {
 								settings: {
 									java: getJavaConfig(requirements.java_home),
 								}
 							});
-							onConfigurationChange(languageClient, context);
+							onConfigurationChange(standardClient.getClient(), context);
```
testforstephen, line 184: I see `onConfigurationChange` is called again at the later code. Actually what it does is just registering a listener via workspace.onDidChangeConfiguration(). Is it necessary to call it again here?

testforstephen: Besides, `onConfigurationChange(languageClient: LanguageClient, context: ExtensionContext)` defines two arguments, but its implementation never uses these arguments.

fbricon:
Thanks @jdneo!

PR #1501, opened 2020-06-23
Ask users to import projects when opening a new folder

jdneo:
Kapture 2020-06-23 at 19 58 17

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,14 +158,16 @@ The following settings are supported:
 * `java.sources.organizeImports.staticStarThreshold`: Specifies the number of static imports added before a star-import declaration is used, default is 99.
 * `java.semanticHighlighting.enabled`: Enable/disable [Semantic Highlighting](https://github.com/redhat-developer/vscode-java/wiki/Semantic-Highlighting) for Java files. Defaults to `false`.
 * `java.requirements.JDK11Warning`: Enable/disable a warning about the impending requirement of Java 11. Defaults to `true`.
-
-New in 0.62.0:
 * `java.refactor.renameFromFileExplorer`: Specifies whether to update imports and package declarations when renaming files from File Explorer. Defaults to `prompt`.
   - `never`: Don't enable refactoring for rename operations on File Explorer.
   - `autoApply`: Always automatically update the imports and package declarations.
   - `preview`: Always preview the changes before applying.
   - `prompt`: Ask user to confirm whether to bypass refactor preview.
 * `java.imports.gradle.wrapper.checksums`: Defines allowed/disallowed SHA-256 checksums of Gradle Wrappers.
+
+New in 0.64.0:
+* `java.project.importOnFirstTimeStartup`: Specifies whether to import the Java projects in workspace on first-time startup. Supported values are `disabled` (never imports), `interactive` (asks to import or not), `automatic` (always imports).
+* `java.project.importHint`: Enable/disable the information for switching server mode when the user skips importing the Java projects on startup. Defaults to `true`.
 
 Semantic Highlighting
 ===============
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ The following settings are supported:
 * `java.imports.gradle.wrapper.checksums`: Defines allowed/disallowed SHA-256 checksums of Gradle Wrappers.
 
 New in 0.64.0:
-* `java.project.importOnFirstTimeStartup`: Specifies whether to import the Java projects in workspace on first-time startup. Supported values are `disabled` (never imports), `interactive` (asks to import or not), `automatic` (always imports).
+* `java.project.importOnFirstTimeStartup`: Specifies whether to import the Java projects in workspace for the first-time startup in hybrid mode. Supported values are `disabled` (never imports), `interactive` (asks to import or not), `automatic` (always imports). Default to `interactive`.
 * `java.project.importHint`: Enable/disable the information for switching server mode when the user skips importing the Java projects on startup. Defaults to `true`.
 
 Semantic Highlighting
--- a/README.md
+++ b/README.md
@@ -166,8 +166,8 @@ The following settings are supported:
 * `java.imports.gradle.wrapper.checksums`: Defines allowed/disallowed SHA-256 checksums of Gradle Wrappers.
 
 New in 0.64.0:
-* `java.project.importOnFirstTimeStartup`: Specifies whether to import the Java projects in workspace for the first-time startup in hybrid mode. Supported values are `disabled` (never imports), `interactive` (asks to import or not), `automatic` (always imports). Default to `interactive`.
-* `java.project.importHint`: Enable/disable the information for switching server mode when the user skips importing the Java projects on startup. Defaults to `true`.
+* `java.project.importOnFirstTimeStartup`: Specifies whether to import the Java projects, when opening the folder in Hybrid mode for the first time. Supported values are `disabled` (never imports), `interactive` (asks to import or not), `automatic` (always imports). Default to `interactive`.
+* `java.project.importHint`: Enable/disable the server-mode switch information, when Java projects import is skipped on startup. Defaults to `true`.
 
 Semantic Highlighting
 ===============
--- a/package.json
+++ b/package.json
@@ -591,12 +591,12 @@
             "automatic"
           ],
           "default": "interactive",
-          "description": "Specifies whether to import the Java projects in workspace for the first-time startup in hybrid mode.",
+          "description": "Specifies whether to import the Java projects, when opening the folder in Hybrid mode for the first time.",
           "scope": "application"
         },
         "java.project.importHint": {
           "type": "boolean",
-          "description": "Enable/disable the information for switching server mode when the user skips importing the Java projects on startup.",
+          "description": "Enable/disable the server-mode switch information, when Java projects import is skipped on startup.",
           "default": true,
           "scope": "application"
         }
--- a/package.json
+++ b/package.json
@@ -30,8 +30,7 @@
     "onLanguage:java",
     "workspaceContains:pom.xml",
     "workspaceContains:build.gradle",
-    "workspaceContains:.classpath",
-    "onCommand:java.project.import"
+    "workspaceContains:.classpath"
   ],
   "main": "./dist/extension",
   "contributes": {
@@ -582,6 +581,23 @@
           },
           "description": "Defines allowed/disallowed SHA-256 checksums of Gradle Wrappers",
           "default": [],
+          "scope": "application"
+        },
+        "java.project.importOnFirstTimeStartup": {
+          "type": "string",
+          "enum": [
+            "disabled",
+            "interactive",
+            "automatic"
+          ],
+          "default": "interactive",
+          "description": "Specifies whether to import the Java projects in workspace on first-time startup.",
+          "scope": "application"
+        },
+        "java.project.importHint": {
+          "type": "boolean",
+          "description": "Enable/disable the information for switching server mode when the user skips import the Java projects on startup.",
+          "default": true,
           "scope": "application"
         }
       }
--- a/package.json
+++ b/package.json
@@ -596,7 +596,7 @@
         },
         "java.project.importHint": {
           "type": "boolean",
-          "description": "Enable/disable the information for switching server mode when the user skips import the Java projects on startup.",
+          "description": "Enable/disable the information for switching server mode when the user skips importing the Java projects on startup.",
           "default": true,
           "scope": "application"
         }
--- a/package.json
+++ b/package.json
@@ -591,7 +591,7 @@
             "automatic"
           ],
           "default": "interactive",
-          "description": "Specifies whether to import the Java projects in workspace on first-time startup.",
+          "description": "Specifies whether to import the Java projects in workspace for the first-time startup in hybrid mode.",
           "scope": "application"
         },
         "java.project.importHint": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -354,7 +354,7 @@ async function promptUserForStandardServer(config: WorkspaceConfiguration): Prom
 			const showHint: boolean = config.get(importHintSection);
 			if (showHint) {
 				const showRocketEmoji: boolean = process.platform === "win32" || process.platform === "darwin";
-				const message: string = `Java Language Server is running in LightWeight mode. Click the 'Rocket' button${showRocketEmoji ? ': 🚀' : ''} in the status bar if you want to import the projects later.`;
+				const message: string = `Java Language Server is running in LightWeight mode. Click the ${showRocketEmoji ? '🚀' : 'Rocket'} icon in the status bar if you want to import the projects later.`;
 				window.showInformationMessage(message, dontShowAgain)
 					.then(selection => {
 						if (selection && selection === dontShowAgain) {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,21 +3,22 @@
 import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
-import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, Selection, Position, EventEmitter, OutputChannel, TextDocument, RelativePattern } from 'vscode';
+import * as fse from 'fs-extra';
+import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, Selection, Position, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration } from 'vscode';
 import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, RevealOutputChannelOn, ErrorHandler, Message, ErrorAction, CloseAction, DidChangeConfigurationNotification } from 'vscode-languageclient';
 import { collectJavaExtensions } from './plugin';
 import { prepareExecutable } from './javaServerStarter';
 import * as requirements from './requirements';
 import { Commands } from './commands';
 import { ExtensionAPI } from './extension.api';
-import { getJavaConfiguration, deleteDirectory } from './utils';
+import { getJavaConfiguration, deleteDirectory, getInclusionFilePatterns, getInclusionPatternsFromNegatedExclusion, getInclusionBlob, getExclusionBlob } from './utils';
 import { onConfigurationChange, getJavaServerMode, ServerMode } from './settings';
 import { logger, initializeLogFile } from './log';
 import glob = require('glob');
 import { SyntaxLanguageClient } from './syntaxLanguageClient';
 import { registerClientProviders } from './providerDispatcher';
 import * as fileEventHandler from './fileEventHandler';
-import { StandardLanguageClient } from './standardLanguageClient';
+import { StandardLanguageClient, ClientStatus } from './standardLanguageClient';
 import { apiManager } from './apiManager';
 import { SnippetCompletionProvider } from './snippetCompletionProvider';
 import { runtimeStatusBarProvider } from './runtimeStatusBarProvider';
@@ -131,7 +132,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 		throw error;
 	}).then(async (requirements) => {
 		const triggerFiles = await getTriggerFiles();
-		return new Promise<ExtensionAPI>((resolve) => {
+		return new Promise<ExtensionAPI>(async (resolve) => {
 			const workspacePath = path.resolve(storagePath + '/jdt_ws');
 			const syntaxServerWorkspacePath = path.resolve(storagePath + '/ss_ws');
 
@@ -139,7 +140,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			commands.executeCommand('setContext', 'java:serverMode', serverMode);
 			const isDebugModeByClientPort = !!process.env['SYNTAXLS_CLIENT_PORT'] || !!process.env['JDTLS_CLIENT_PORT'];
 			const requireSyntaxServer = (serverMode !== ServerMode.STANDARD) && (!isDebugModeByClientPort || !!process.env['SYNTAXLS_CLIENT_PORT']);
-			const requireStandardServer = (serverMode !== ServerMode.LIGHTWEIGHT) && (!isDebugModeByClientPort || !!process.env['JDTLS_CLIENT_PORT']);
+			let requireStandardServer = (serverMode !== ServerMode.LIGHTWEIGHT) && (!isDebugModeByClientPort || !!process.env['JDTLS_CLIENT_PORT']);
 
 			// Options to control the language client
 			const clientOptions: LanguageClientOptions = {
@@ -196,16 +197,15 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 				outputChannelName: extensionName
 			};
 
+			apiManager.initialize(requirements);
+
 			if (requireSyntaxServer) {
 				if (process.env['SYNTAXLS_CLIENT_PORT']) {
 					syntaxClient.initialize(requirements, clientOptions, resolve);
 				} else {
 					syntaxClient.initialize(requirements, clientOptions, resolve, prepareExecutable(requirements, syntaxServerWorkspacePath, getJavaConfig(requirements.java_home), context, true));
 				}
-			}
-
-			if (requireStandardServer) {
-				standardClient.initialize(context, requirements, clientOptions, workspacePath, jdtEventEmitter, resolve);
+				syntaxClient.start();
 			}
 
 			context.subscriptions.push(commands.registerCommand(Commands.EXECUTE_WORKSPACE_COMMAND, (command, ...rest) => {
@@ -220,16 +220,6 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 				};
 				return standardClient.getClient().sendRequest(ExecuteCommandRequest.type, params);
 			}));
-
-			apiManager.initialize(requirements);
-
-			if (requireSyntaxServer) {
-				syntaxClient.start();
-			}
-
-			if (requireStandardServer) {
-				standardClient.start();
-			}
 
 			const cleanWorkspaceExists = fs.existsSync(path.join(workspacePath, cleanWorkspaceFileName));
 			if (cleanWorkspaceExists) {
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -583,6 +582,23 @@
...
+          ],
+          "default": "interactive",
+          "description": "Specifies whether to import the Java projects in workspace on first-time startup.",
+          "scope": "application"
+        },
+        "java.project.importHint": {
+          "type": "boolean",
+          "description": "Enable/disable the information for switching server mode when the user skips import the Java projects on startup.",
+          "default": true,
+          "scope": "application"
```
fbricon, line 601: new prefs need to be documented in the readme.md

src/utils.ts
```diff
@@ -45,3 +45,70 @@ export function ensureExists(folder) {
...
+	const config = getJavaConfiguration();
+	const isMavenImporterEnabled: boolean = config.get<boolean>("import.maven.enabled");
+	const isGradleImporterEnabled: boolean = config.get<boolean>("import.gradle.enabled");
+	const patterns: string[] = [];
+	if (isMavenImporterEnabled) {
+		patterns.push("**/pom.xml");
+	}
+	if (isGradleImporterEnabled) {
+		patterns.push("**/build.gradle");
+	}
```
fbricon, line 59: we need to figure out a way to let 3rd party extensions register patterns/import prefs from their package.json (something similar to extension jars)

jdneo: Something like this?

```js
"contributes": {
    "javaConfigurationFiles": [
        {
            "name": "BUILD",
            "contentPattern": "java_binary\\s*\\([\\s\\S]*\\)"  <--- optional
        }
    ],
    "javaExtensions": [
        ...
    ],
    ...
}
```

where `contentPattern` is optional, since for `Bazel`, the `BUILD` file can also be used to build other language's project, so we need a regex pattern to test towards the file content.

testforstephen: IMO, you doesn't need test against the file content, because vscode-java won't be automatically activated when detecting BUILD file. 

```json
"contributes": {
  "javaBuildFiles": [
      {
          "fileNamePattern": "**/BUILD"
      }
  ]
}

```

jdneo: Hard to say... If the third-party extension depends on vscode-java, then it will be activated.

package.json
```diff
@@ -583,6 +582,23 @@
...
+        },
+        "java.project.importOnFirstTimeStartup": {
+          "type": "string",
+          "enum": [
+            "disabled",
+            "interactive",
+            "automatic"
+          ],
+          "default": "interactive",
+          "description": "Specifies whether to import the Java projects in workspace on first-time startup.",
```
testforstephen, line 594: this setting only takes effect in hybrid mode. do you want to mention this info in some place?

src/syntaxLanguageClient.ts
```diff
@@ -96,4 +101,24 @@ export class SyntaxLanguageClient {
 	public getClient(): LanguageClient {
 		return this.languageClient;
 	}
+
+	public registerUIComponents(): void {
+		const item = window.createStatusBarItem(StatusBarAlignment.Right, Number.MIN_VALUE);
+		item.text = "LightWeight Mode";
```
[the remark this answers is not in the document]
jdneo: Looks good:

image

package.json
jdneo: Done

package.json
```diff
@@ -30,8 +30,7 @@
     "onLanguage:java",
     "workspaceContains:pom.xml",
     "workspaceContains:build.gradle",
-    "workspaceContains:.classpath",
-    "onCommand:java.project.import"
```
fbricon, line 34: why would you remove it? If you start from an empty folder, then add a pom.xml, you'd want to be able to trigger import even though vscode-java hasn't started yet

package.json
```diff
@@ -583,6 +582,23 @@
...
+        },
+        "java.project.importOnFirstTimeStartup": {
+          "type": "string",
+          "enum": [
+            "disabled",
+            "interactive",
+            "automatic"
+          ],
+          "default": "interactive",
+          "description": "Specifies whether to import the Java projects in workspace for the first-time startup in hybrid mode.",
```
fbricon, line 594: Specifies whether to import the Java projects, when opening the folder in hybrid mode for the first time.

package.json
```diff
@@ -583,6 +582,23 @@
...
+          "scope": "application"
+        },
+        "java.project.importHint": {
+          "type": "boolean",
+          "description": "Enable/disable the information for switching server mode when the user skips importing the Java projects on startup.",
```
fbricon, line 599: Enable/disable the server-mode switch information, when Java projects import is skipped on startup.

test/standard-mode-suite/utils.test.ts
```diff
@@ -0,0 +1,130 @@
...
+	test('getBuildFilePatterns() - no importers is enabled', async function () {
+		await config.update(IMPORT_MAVEN_ENABLED, false);
+		await config.update(IMPORT_GRADLE_ENABLED, false);
+
+		const result: string[] = getBuildFilePatterns();
+
+		assert.deepEqual(result, []);
+	});
+
+	test('getInclusionPatternsFromNegatedExclusion() - no negeted exclusions', async function () {
```
fbricon, line 52: negated

test/standard-mode-suite/utils.test.ts
```diff
@@ -0,0 +1,130 @@
...
+			"**/archetype-resources/**",
+			"**/META-INF/maven/**"
+		]);
+
+		const result: string[] = getInclusionPatternsFromNegatedExclusion();
+
+		assert.deepEqual(result, []);
+	});
+
+	test('getInclusionPatternsFromNegatedExclusion() - have negeted exclusions', async function () {
```
fbricon, line 65: negated

test/standard-mode-suite/utils.test.ts
```diff
@@ -0,0 +1,130 @@
...
+			"**/archetype-resources/**",
+			"**/META-INF/maven/**"
+		]);
+
+		const result: string[] = getInclusionPatternsFromNegatedExclusion();
+
+		assert.deepEqual(result, ["**/node_modules/test/**"]);
+	});
+
+	test('getExclusionBlob() - no negeted exclusions', async function () {
```
fbricon, line 79: negated

test/standard-mode-suite/utils.test.ts
```diff
@@ -0,0 +1,130 @@
...
+			"**/archetype-resources/**",
+			"**/META-INF/maven/**"
+		]);
+
+		const result: string = getExclusionBlob();
+
+		assert.equal(result, "{**/node_modules/**,**/.metadata/**,**/archetype-resources/**,**/META-INF/maven/**}");
+	});
+
+	test('getExclusionBlob() - have negeted exclusions', async function () {
```
fbricon, line 92: has negated

test/standard-mode-suite/utils.test.ts
```diff
@@ -0,0 +1,130 @@
...
+		const result: string = convertToGlob([]);
+		assert.equal(result, "");
+	});
+
+	test('convertToGlob() - no base patterns', async function () {
+		const result: string = convertToGlob(["**/pom.xml", "**/build.gradle"]);
+		assert.equal(result, "{**/pom.xml,**/build.gradle}");
+	});
+
+	test('convertToGlob() - have base patterns', async function () {
```
fbricon, line 116: has

README.md
```diff
@@ -158,15 +158,17 @@ The following settings are supported:
...
   - `never`: Don't enable refactoring for rename operations on File Explorer.
   - `autoApply`: Always automatically update the imports and package declarations.
   - `preview`: Always preview the changes before applying.
   - `prompt`: Ask user to confirm whether to bypass refactor preview.
 * `java.imports.gradle.wrapper.checksums`: Defines allowed/disallowed SHA-256 checksums of Gradle Wrappers.
 
+New in 0.64.0:
+* `java.project.importOnFirstTimeStartup`: Specifies whether to import the Java projects in workspace for the first-time startup in hybrid mode. Supported values are `disabled` (never imports), `interactive` (asks to import or not), `automatic` (always imports). Default to `interactive`.
+* `java.project.importHint`: Enable/disable the information for switching server mode when the user skips importing the Java projects on startup. Defaults to `true`.
+
```
fbricon, line 171: update descriptions based on feedback

package.json
testforstephen: For all UI message, please keep the consistent capitalization with the server mode terms. For example, it should be **H**ybrid, **L**ight**W**eight, **S**tandard.

src/extension.ts
```diff
@@ -278,13 +284,83 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+			return true;
+		case "Yes":
+			return true;
+		case "Later":
+		default:
+			const importHintSection: string = "project.importHint";
+			const dontShowAgain: string = "Don't Show Again";
+			const showHint: boolean = config.get(importHintSection);
+			if (showHint) {
+				const message: string = "Java Language Server is running in LightWeight mode. Click the button 'LightWeight Mode' in the status bar if you want to import the projects later.";
```
testforstephen, line 352: Click the button 🚀 ...

src/syntaxLanguageClient.ts
```diff
@@ -96,4 +101,24 @@ export class SyntaxLanguageClient {
...
+
+	public registerUIComponents(): void {
+		const item = window.createStatusBarItem(StatusBarAlignment.Right, Number.MIN_VALUE);
+		item.text = "$(rocket)";
+		item.command = {
+			title: "Switch to Standard mode",
+			command: Commands.SWITCH_SERVER_MODE,
+			arguments: [ServerMode.STANDARD],
+		};
+		item.tooltip = "Click to switch the Java language server to standard mode";
```
testforstephen, line 113: Java language server is running in LightWeight mode, click to switch to Standard mode

src/extension.ts
```diff
@@ -254,15 +244,31 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
 					return;
 				}
 
-				standardClient.initialize(context, requirements, clientOptions, workspacePath, jdtEventEmitter, resolve);
-				standardClient.start();
+				let choice: string;
+				if (force) {
+					choice = "Yes";
+				} else {
+					choice = await window.showInformationMessage("Are you sure you want to switch the Java language server to standard mode?", "Yes", "No");
```
testforstephen, line 264: standard -> Standard

package.json
jdneo: There is no API to let the extension know which event activates it. When VS Code Java is activated by this command, the user will be asked again to confirm the import again. Is it acceptable?


Maybe a better solution is to let VS Code check the adding files for `workspaceContains`: https://github.com/microsoft/vscode/issues/28122.

src/syntaxLanguageClient.ts
jdneo: Will it too long for a hover message?

src/extension.ts
jdneo: Not sure how will the emoji looks like on different platform. I'll do some tests tomorrow...

src/syntaxLanguageClient.ts
testforstephen: Is there any display issue about long tooltip? i believe user doesn't know what the rocket icon means.

jdneo: I see. Make sense.

src/extension.ts
jdneo: The rocket emoji has some problem on `Ubuntu 18.04` + `xfce4` desktop env, I suggest to still use the plain text instead.

Ubuntu 18.04 + xfce4:
Screen Shot 2020-06-30 at 12 58 49 PM


Windows 10: 
image

Mac OS 10.15.5:
Screen Shot 2020-06-30 at 1 01 49 PM

jdneo: Changed the behavior to: Only show rocket emoji on windows and mac:

image

src/syntaxLanguageClient.ts
```diff
@@ -96,4 +101,24 @@ export class SyntaxLanguageClient {
 	public getClient(): LanguageClient {
 		return this.languageClient;
 	}
+
+	public registerUIComponents(): void {
+		const item = window.createStatusBarItem(StatusBarAlignment.Right, Number.MIN_VALUE);
```
testforstephen, line 106: Can we let all server status share one StatusBarItem instance? Currently standard server created one, and lightweight server created one. Given they doesn't exist at the same time.

jdneo: Sure I can do that. But that needs some refactoring: decouple the status bar item out of each language client. Should I do it in this PR or another one?

fbricon:
@jdneo can you please fix the conflict?

src/syntaxLanguageClient.ts
testforstephen: A new PR is OK.

jdneo: Tracking here: https://github.com/redhat-developer/vscode-java/issues/1506

jdneo:
@fbricon Sure, rebased.

src/extension.ts
```diff
@@ -285,10 +291,81 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+		case "Yes":
+			return true;
+		case "Later":
+		default:
+			const importHintSection: string = "project.importHint";
+			const dontShowAgain: string = "Don't Show Again";
+			const showHint: boolean = config.get(importHintSection);
+			if (showHint) {
+				const showRocketEmoji: boolean = process.platform === "win32" || process.platform === "darwin";
+				const message: string = `Java Language Server is running in LightWeight mode. Click the 'Rocket' button${showRocketEmoji ? ': 🚀' : ''} in the status bar if you want to import the projects later.`;
```
fbricon, line 357: Click the ${showRocketEmoji ? ': 🚀':'Rocket'} icon

fbricon:
Everytime I close the vscode window, the syntax server fails to stop on its own, looks like it never receives the shutdown request

jdneo:
I tried using `jps` after closing the vs code and not found any language server process. @testforstephen Could you reproduce that?

testforstephen:
> I tried using `jps` after closing the vs code and not found any language server process. @testforstephen Could you reproduce that?

I observed the lightweight server is still running for a while after i forced to close VS Code window. However, it would automatically exit after about 1 minute.

Below is the log.

```
!ENTRY org.eclipse.jdt.ls.core 1 0 2020-07-01 13:20:12.657
!MESSAGE Parent process stopped running, forcing server exit

!ENTRY org.eclipse.jdt.ls.core 1 0 2020-07-01 13:20:12.658
!MESSAGE Shutdown received... waking up main thread

!ENTRY org.eclipse.jdt.ls.core 1 0 2020-07-01 13:20:12.665
!MESSAGE class org.eclipse.jdt.ls.core.internal.JavaLanguageServerPlugin is stopping:

!ENTRY org.eclipse.core.resources 2 10035 2020-07-01 13:20:12.678
!MESSAGE The workspace will exit with unsaved changes in this session.
```

testforstephen:
The Standard mode looks like the same behavior as "Parent process stopped running, forcing server exit".

src/extension.ts
```diff
@@ -278,13 +284,83 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+				return await promptUserForStandardServer(config);
+			}
+		}
+	}
+
+	return true;
+}
+
+async function promptUserForStandardServer(config: WorkspaceConfiguration): Promise<boolean> {
+	const choice: string = await window.showInformationMessage("The workspace contains Java projects, would you like to import them?", "Yes", "Later", "Always");
```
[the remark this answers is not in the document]
jdneo: Done:

WeChat Screenshot_20200701133203

fbricon:
I see the standard server receiving 
```
!ENTRY org.eclipse.jdt.ls.core 1 0 2020-07-01 11:53:35.884
!MESSAGE >> shutdown

!ENTRY org.eclipse.jdt.ls.core 1 0 2020-07-01 11:53:35.923
!MESSAGE >> exit
```

but never for the syntax server

fbricon:
From the client log, I can see the shutdown command is never sent for the syntax client, but it is for the standard client, which explains my previous comment. But looking at the code, I don't see why this is not working the same for both.
Very easy to reproduce the issue: simply execute the reload window command (Cmd+R when running in dev mode).

fbricon:
The server shutting down on its own after about 30 sec is thanks to https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/ParentProcessWatcher.java#L49

snjeza:
@fbricon @testforstephen @akaroml @jdneo could you try the following Java LS patch:

```
diff --git a/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/JDTLanguageServer.java b/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/JDTLanguageServer.java
index a6710c75..d2e23a1a 100644
--- a/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/JDTLanguageServer.java
+++ b/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/JDTLanguageServer.java
@@ -161,6 +161,29 @@ public class JDTLanguageServer extends BaseJDTLanguageServer implements Language
 
 	private ProgressReporterManager progressReporterManager;
 
+	private Job shutdownJob = new Job("Shutdown...") {
+
+		@Override
+		protected IStatus run(IProgressMonitor monitor) {
+			try {
+				JavaRuntime.removeVMInstallChangedListener(jvmConfigurator);
+				if (workspaceDiagnosticsHandler != null) {
+					workspaceDiagnosticsHandler.removeResourceChangeListener();
+					workspaceDiagnosticsHandler = null;
+				}
+				if (classpathUpdateHandler != null) {
+					classpathUpdateHandler.removeElementChangeListener();
+					classpathUpdateHandler = null;
+				}
+				ResourcesPlugin.getWorkspace().save(true, monitor);
+			} catch (CoreException e) {
+				logException(e.getMessage(), e);
+			}
+			return Status.OK_STATUS;
+		}
+
+	};
+
 	@Override
 	public LanguageServerWorkingCopyOwner getWorkingCopyOwner() {
 		return workingCopyOwner;
@@ -348,20 +371,7 @@ public class JDTLanguageServer extends BaseJDTLanguageServer implements Language
 	public CompletableFuture<Object> shutdown() {
 		logInfo(">> shutdown");
 		return computeAsync((monitor) -> {
-			try {
-				JavaRuntime.removeVMInstallChangedListener(jvmConfigurator);
-				if (workspaceDiagnosticsHandler != null) {
-					workspaceDiagnosticsHandler.removeResourceChangeListener();
-					workspaceDiagnosticsHandler = null;
-				}
-				if (classpathUpdateHandler != null) {
-					classpathUpdateHandler.removeElementChangeListener();
-					classpathUpdateHandler = null;
-				}
-				ResourcesPlugin.getWorkspace().save(true, monitor);
-			} catch (CoreException e) {
-				logException(e.getMessage(), e);
-			}
+			shutdownJob.schedule();
 			return new Object();
 		});
 	}
@@ -372,6 +382,11 @@ public class JDTLanguageServer extends BaseJDTLanguageServer implements Language
 	@Override
 	public void exit() {
 		logInfo(">> exit");
+		try {
+			shutdownJob.join();
+		} catch (InterruptedException e) {
+			JavaLanguageServerPlugin.logException(e.getMessage(), e);
+		}
 		JavaLanguageServerPlugin.getLanguageServer().exit();
 		Executors.newSingleThreadScheduledExecutor().schedule(() -> {
 			logInfo("Forcing exit after 1 min.");
diff --git a/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/syntaxserver/SyntaxLanguageServer.java b/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/syntaxserver/SyntaxLanguageServer.java
index 44daf7f1..d3582e89 100644
--- a/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/syntaxserver/SyntaxLanguageServer.java
+++ b/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/syntaxserver/SyntaxLanguageServer.java
@@ -29,7 +29,9 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jdt.ls.core.internal.BaseJDTLanguageServer;
 import org.eclipse.jdt.ls.core.internal.JDTUtils;
@@ -92,6 +94,19 @@ public class SyntaxLanguageServer extends BaseJDTLanguageServer implements Langu
 	private ContentProviderManager contentProviderManager;
 	private ProjectsManager projectsManager;
 	private PreferenceManager preferenceManager;
+	private Job shutdownJob = new Job("Shutdown...") {
+
+		@Override
+		protected IStatus run(IProgressMonitor monitor) {
+			try {
+				ResourcesPlugin.getWorkspace().save(true, monitor);
+			} catch (CoreException e) {
+				logException(e.getMessage(), e);
+			}
+			return Status.OK_STATUS;
+		}
+
+	};
 
 	public SyntaxLanguageServer(ContentProviderManager contentProviderManager,
 								ProjectsManager projectsManager,
@@ -120,12 +135,7 @@ public class SyntaxLanguageServer extends BaseJDTLanguageServer implements Langu
 	public CompletableFuture<Object> shutdown() {
 		logInfo(">> shutdown");
 		return computeAsync((monitor) -> {
-			try {
-				ResourcesPlugin.getWorkspace().save(true, monitor);
-			} catch (CoreException e) {
-				logException(e.getMessage(), e);
-			}
-
+			shutdownJob.schedule();
 			return new Object();
 		});
 	}
@@ -133,6 +143,11 @@ public class SyntaxLanguageServer extends BaseJDTLanguageServer implements Langu
 	@Override
 	public void exit() {
 		logInfo(">> exit");
+		try {
+			shutdownJob.join();
+		} catch (InterruptedException e) {
+			JavaLanguageServerPlugin.logException(e.getMessage(), e);
+		}
 		JavaLanguageServerPlugin.getLanguageServer().exit();
 		Executors.newSingleThreadScheduledExecutor().schedule(() -> {
 			System.exit(1);
```

snjeza:
The issue happens when ResourcesPlugin.getWorkspace().save(true, monitor) takes more than five seconds in which case VS Code doesn't send the exit command.

jdneo:
OK, I can now repro this on Mac. Seems this issue is not reproduceable on Windows.

jdneo:
@snjeza Thank you for the patch! It works while it has a limitation. It ensures the language serve will be shut down if the initialize request is finished.

If the VS Code is closed before the initialize response is sent from the language server, the client will not send the shutdown event to the server -- which is a [restriction of LSP](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize). Thus the servers will still be running. (to simply repro this, you can close the VS Code at the moment when the spinning icon shows at the staus-bar).

_// BTW, I suggest to discuss this topic in another new issue thread, since it's out of this PR's scope._

fbricon:
so I found out the extension.deactivate() function is never called when the server is started in hybrid mode but the projects have not been imported. So the syntax server client is never stopped, so never sends the shutdown signal.

I just can't figure out why

src/extension.ts
```diff
@@ -285,10 +291,81 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+
+			if (serverMode === ServerMode.HYBRID && !await fse.pathExists(path.join(workspacePath, ".metadata", ".plugins"))) {
+				requireStandardServer = await isStandardServerRequired();
+			}
+
+			if (requireStandardServer) {
+				startStandardServer(context, requirements, clientOptions, workspacePath, resolve);
+			} else {
+				apiManager.getApiInstance().serverMode = ServerMode.LIGHTWEIGHT;
+				apiManager.fireDidServerModeChange(ServerMode.LIGHTWEIGHT);
```
fbricon, line 303:
```typescript
resolve(apiManager.getApiInstance());
```

There's our culprit, the extension was never registered as deactivable unless the Promise was resolved

snjeza: It works for me.

fbricon:
> @fbricon @testforstephen @akaroml @jdneo could you try the following Java LS patch:
> 
please open a PR in jdt.ls, that's a different issue

jdneo:
In case the user may not select an option when asking import. I did some more changes to make sure the vscode java will resolve the API when the asking is needed.

src/extension.ts
```diff
@@ -285,10 +292,88 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
 	});
 }
 
+function startStandardServer(context: ExtensionContext, requirements: requirements.RequirementsData, clientOptions: LanguageClientOptions, workspacePath: string, resolve: (value?: ExtensionAPI | PromiseLike<ExtensionAPI>) => void) {
+	standardClient.initialize(context, requirements, clientOptions, workspacePath, jdtEventEmitter, resolve);
+	standardClient.start();
+	syntaxClient.disposeUIComponents();
+}
+
+async function isStandardServerRequired(resolve: (value: ExtensionAPI) => void): Promise<boolean> {
```
fbricon, line 313: A method called isStandardServer that resolves the extension activation seems wrong

src/syntaxLanguageClient.ts
```diff
@@ -52,48 +55,84 @@ export class SyntaxLanguageClient {
...
+	}
+
+	public disposeUIComponents(): void {
+		for (const disposable of this.disposables) {
+			disposable.dispose();
+		}
+		this.disposables = [];
+	}
+
+	public resolveApi(resolve: (value: ExtensionAPI) => void) {
```
fbricon, line 127: public resolveApi(resolve: (value: ExtensionAPI) => void): void

src/extension.ts
```diff
@@ -285,10 +292,82 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+	if (inclusionBlob && (await workspace.findFiles(inclusionBlob, exclusionBlob, 1 /*maxResults*/)).length > 0) {
+		return true;
+	}
+
+	return false;
+}
+
+async function promptUserForStandardServer(config: WorkspaceConfiguration): Promise<boolean> {
+	const choice: string = await window.showInformationMessage("The workspace contains Java projects, would you like to import them?", "Yes", "Always", "Later");
+	switch (choice) {
```
fbricon, line 346: should we add Never?

jdneo: Hmm... Personal option is not including Never here. Since our final goal is to let users to use all the features of VS Code Java.

fbricon:
Thanks @jdneo!

snjeza:
> please open a PR in jdt.ls, that's a different issue

@fbricon See https://github.com/eclipse/eclipse.jdt.ls/pull/1495

PR #1504, opened 2020-06-28
Use java:serverMode to avoid conflicts

jdneo:
Since the context value is shared between all the extensions. Other extensions can leverage this value in the when clause in the package.json.

It would be better to call it `java:serverMode` to avoid conflicts.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -135,8 +135,6 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			const syntaxServerWorkspacePath = path.resolve(storagePath + '/ss_ws');
 
 			const serverMode = getJavaServerMode();
-			// TODO: deptecated this in the future, use 'java:serverMode' instead
-			commands.executeCommand('setContext', 'serverMode', serverMode);
 			commands.executeCommand('setContext', 'java:serverMode', serverMode);
 			const isDebugModeByClientPort = !!process.env['SYNTAXLS_CLIENT_PORT'] || !!process.env['JDTLS_CLIENT_PORT'];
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -135,7 +135,9 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			const syntaxServerWorkspacePath = path.resolve(storagePath + '/ss_ws');
 
 			const serverMode = getJavaServerMode();
+			// TODO: deptecated this in the future, use 'java:serverMode' instead
```
testforstephen, line 138: typos. deptecated -> deprecated.

src/extension.ts
```diff
@@ -135,7 +135,9 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
 			commands.executeCommand('setContext', 'serverMode', serverMode);
```
testforstephen, line 139: Since this key is exposed a few moments ago, is there anyone adopting this key? if not, it's ok to remove it directly.

jdneo: Removed

src/extension.ts
```diff
@@ -278,7 +278,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 					snippetProvider.setActivation(false);
 					fileEventHandler.setServerStatus(true);
 					runtimeStatusBarProvider.initialize(context.storagePath);
-					commands.executeCommand('setContext', 'serverMode', event);
+					commands.executeCommand('setContext', 'java:serverMode', serverMode);
```
testforstephen, line 281: The last parameter `event` is changed to `serverMode`, On purpose?

jdneo: My bad, it should be `event`

fbricon:
Thanks @jdneo

Issue #1276, opened 2020-02-17
Intellisense / autocomplete list incomplete


spottedmahn:
The autocomplete listing on the Elasticsearch `RestHighLevelClient` is incomplete.

##### Environment
- Operating System: Win 10 v1909
- JDK version: openjdk version "11.0.5" 2019-10-15
- Visual Studio Code version: 1.42.1 (user setup)
- Java extension version: version: 0.55.1

##### Steps To Reproduce
1. Start a project w/ the [Elasticsearch RestHighLevelClient](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/master/java-rest-high-getting-started-initialization.html) 
1. Initialize a `RestHighLevelClient`
1. Perform intellisense on the client

##### Current Result
![image](https://user-images.githubusercontent.com/14079228/74659089-a9c29280-5161-11ea-9a89-bdc79a6cf09b.png)

##### Expected Result
A full listing of all accessible methods (other things, fields, etc.).

For example, [`search()`](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/master/java-rest-high-search.html) is missing.

##### Additional Informations
Pretty sure I've seen this on other objects and this is **not** specific to the `RestHighLevelClient`.

fbricon:
The completion list is limited to 50 items by default, for performance purposes. You can either continue typing `client.s`, to fetch the next matching results or increase the `java.completion.maxResults` preference value. See https://github.com/eclipse/eclipse.jdt.ls/pull/1298 for more details.

spottedmahn:
>The completion list is limited to 50 items by default, for performance purposes

Fair enough.

What about an **indicator** that the list is **incomplete**?  As a newbie to Java w/ VS Code it is not obvious, at least not to me!

And on that "incomplete list indictator", what about a "load all option"?

---

And thanks for the super quick reply! ⚡🤝

fbricon:
Not sure that is possible with vscode. We'll look into it.

akaroml:
If `isIncomplete` is set to `true`, then vscode will request for another completion list whenever users type something new. When set to `false`, then vscode will consider the list complete and will be able to do filtering as users type more without requesting the LS for more completion suggestions.

In short,
`isIncomplete == true` => LS does the filtering (by handling more requests)
`isIncomplete == false` => vscode does the filtering

jdneo:
The `java.completion.maxResults` is now default to `0`, which means the limitation is disabled.

[...]

PR #1507, opened 2020-06-30
Disable completion limit

fbricon:
following https://github.com/redhat-developer/vscode-java/pull/1503, we can set java.completion.maxResults to a much larger value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ The following settings are supported:
 * `java.selectionRange.enabled`: Enable/disable Smart Selection support for Java. Disabling this option will not affect the VS Code built-in word-based and bracket-based smart selection.
 * `java.showBuildStatusOnStart.enabled`: Automatically show build status on startup. Defaults to `false`.
 * `java.project.referencedLibraries`: Configure glob patterns for referencing local libraries to a Java project.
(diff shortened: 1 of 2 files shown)


fbricon:
@Eskibear @jdneo @testforstephen wdyt?

jdneo:
Or maybe just use `0`?

Eskibear:
0 should be fine. 2000 doesn't solve the inconvinience brought by the "incomplete list" approach, like in #1276
Known max value for a typical project (spring-petclinic) is ~3000, with the 2000 limit I don't feel the difference, in both case there's a little bit lagging like hundreds of ms.

2000 is indeed a safe choice for large projects with more packages and classes (e.g. >10K candidates). We can change the description to tell users, they can set the limit for better performance but in trade of completeness. 

The more the number is, the more cases where it gives a complete list. I'm ok with both, but prefer 0(infinite) for simplicity and less confusion. Then we only need to explain one thing:
* When users get bad perf for extremely large projects, we point them to the setting.

Otherwise we'll also have to explain why the completion list **sometimes** is incomplete.

fbricon:
Ok, set limit to 0 (aka ∞)

PR #1513, opened 2020-07-08
bugfix: Click 'Always' has no response

jdneo:
If not specify `ConfigurationTarget.Global`, VS Code will try to save it to workspace setting then fail.


fbricon:
please update package.json* to 0.64.1, as well as the changelog. I'll release it later today

jdneo:
@fbricon Updated.

PR #1522, opened 2020-07-14
Do not show hint when standard server is already activated

jdneo:
This change is to avoid showing the import hint dialog if the standard server is already activated. Following steps shows how to reproduce the issue without this change:

- Open a fresh workspace
- Ignore the message box that asks to import the projects in workspace
- Click the rocket icon in status bar to switch the server mode
- Now, click the buttons(later/close) in the message box that asks to import.

> Note: There is no API to get the handler of the message box in VS Code. Thus we cannot dispose it when some certain conditions are met. But at least what we can do here is to avoid showing the hint message box if the standard server is already started.

### DIFF
--- a/src/serverStatusBarProvider.ts
+++ b/src/serverStatusBarProvider.ts
@@ -24,12 +24,10 @@ class ServerStatusBarProvider implements Disposable {
 	}
 
 	public showStandardStatus(): void {
-		if (this.statusBarItem.text === StatusIcon.LightWeight) {
-			this.statusBarItem.text = StatusIcon.Busy;
-			this.statusBarItem.command = Commands.SHOW_SERVER_TASK_STATUS;
-			this.statusBarItem.tooltip = "";
-			this.statusBarItem.show();
-		}
+		this.statusBarItem.text = StatusIcon.Busy;
+		this.statusBarItem.command = Commands.SHOW_SERVER_TASK_STATUS;
+		this.statusBarItem.tooltip = "";
+		this.statusBarItem.show();
 	}
 
 	public updateText(text: string): void {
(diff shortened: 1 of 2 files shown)


src/serverStatusBarProvider.ts
```diff
@@ -24,26 +24,28 @@ class ServerStatusBarProvider implements Disposable {
 	}
 
 	public showStandardStatus(): void {
-		this.statusBarItem.text = "$(sync~spin)";
-		this.statusBarItem.command = Commands.SHOW_SERVER_TASK_STATUS;
-		this.statusBarItem.tooltip = "";
-		this.statusBarItem.show();
+		if (this.statusBarItem.text === StatusIcon.LightWeight) {
```
testforstephen, line 27: what's the purpose of this condition check?

jdneo: It's to avoid the icon becoming spinning when the method is called multi times

testforstephen: Is there user scenario which needs to call it multi times?

jdneo: the method `startStandardServer()` might be called multi times. For example:
1. User ignores the message box which asks to import
2. User clicks the rocket icon to switch the server mode
3. User pops up the message box and click `Yes` or `Always`.
4. In this case, `startStandardServer()` will be executed again.

So, maybe another way to better handle this situation is to check the client state at the beginning of `startStandardServer()`?

testforstephen: Yes, regarding the single responsibility principle, it's better to check it at the outside.

fbricon:
Thanks @jdneo!

PR #1524, opened 2020-07-15
Require Java 11 to run the extension

fbricon:
follow up on https://github.com/eclipse/eclipse.jdt.ls/pull/1509

### DIFF
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -85,29 +85,10 @@ function checkJavaVersion(javaHome: string): Promise<number> {
         const javaBin = path.join(javaHome, "bin", JAVA_FILENAME);
         cp.execFile(javaBin, ['-version'], {}, (error, stdout, stderr) => {
             const javaVersion = parseMajorVersion(stderr);
-            if (javaVersion < 8) {
-                openJDKDownload(reject, 'Java 8 or more recent is required to run. Please download and install a recent JDK');
-            } else {
-                if (javaVersion < 11) {
-                    const section = "requirements.JDK11Warning";
-                    const dontShowAgain = "Don't Show Again";
-                    const getJDK = "Get the Java Development Kit";
-                    const jdkWarning = getJavaConfiguration().get(section);
-                    if (jdkWarning) {
-                        const message = `Java 11 [will soon be the minimum requirement](https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements#jdk11.requirement) to run. Consider installing a recent JDK now.`;
-                        window.showInformationMessage(`${message}`, getJDK, dontShowAgain)
-                            .then(selection => {
-                                if (selection === dontShowAgain) {
-                                    getJavaConfiguration().update(section, false, ConfigurationTarget.Global);
-                                }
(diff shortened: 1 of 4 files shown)


src/requirements.ts
```diff
@@ -85,29 +85,9 @@ function checkJavaVersion(javaHome: string): Promise<number> {
...
-                                if (selection === dontShowAgain) {
-                                    getJavaConfiguration().update(section, false, ConfigurationTarget.Global);
-                                }
-                                if (selection === getJDK) {
-                                    commands.executeCommand(Commands.OPEN_BROWSER, Uri.parse(getJdkUrl()));
-                                }
-                            });
-                    }
-                }
-                resolve(javaVersion);
```
snjeza, line 109: You should add this line.

fbricon: Yes obviously :man_facepalming:. 
I wonder why the extension did work in regular vscode though. I must have seriously fucked up

demolixion2/src/main/java/foo/bar/demo/App.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Hello world!
+ *
+ */
+public class App 
+{
+    public static void main( String[] args )
+    {
+        System.out.println( "Hello World!"
```
Eskibear, line 11: just wondering what's this file for?

fbricon: good catch. Removed it

frankxls:
I'm not able to run some projects in Java 8 anymore. I guess is because of this "enhancement", however I still need to run such version. If not, there's no point in using VSCode and this extension. Thanks anyway.

fbricon:
https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements#do-i-need-to-migrate-my-projects-to-java-11

qq849546387:
I have setting "java.configuration.runtimes", but no effect, so I hava to change it to v0.64.1.

tigerinus:
> I have setting "java.configuration.runtimes", but no effect, so I hava to change it to v0.64.1.

How did you revert to previous version? I could only find v0.61 at https://github.com/redhat-developer/vscode-java/releases

tigerinus:
@frankxls @qq849546387 - from the description it sounds like JDK11 is only needed for this extension to run. It should not impact your JDK8 based development work if you setup is correct.

JDK8 is mandatory to my work too. I do not need to change either my `JDK_HOME` or `JAVA_HOME` pointing to JDK8/JRE8. My setup below works fine for me:

```json
    "java.home": "/opt/jdk11",
    "java.configuration.runtimes": [
        {
            "name": "JavaSE-1.8",
            "path": "/opt/jdk8",
            "default": true
        }
    ]
```

On the other hand for owners of this extension, maybe you should consider adding a new setting like `jdt.java.home` (in the same way like `xml.java.home` from XML extension and `metals.javaHome` from Scala extension), instead of using `java.home`.

gajduk:
enhancement - Is that the new word for breaking change?

PR #1534, opened 2020-07-20
Tune the user settings about specifying gradle distribution

### LINKED ISSUE #1495: The gradle wrapper validator won't work in the machine without internet access
To reproduce it with the latest CI build:
1. Turn off your wifi and keep it without internet access.
2. Clean up `~/tooling/gradle`, local gradle user home `~/.gradle`, and leave all gradle vscode settings to the default value.
3. Open a sample gradle wrapper project below in VS Code.
[springbootgradledemo.zip](https://github.com/redhat-developer/vscode-java/files/4784624/springbootgradledemo.zip)
4. You would be prompted about security warning, and click `Trust` to continue.
5. Even though the LS status is 👍, but actually it fails to import the gradle project. No .classpath is generated. No projects are displayed in the Java Dependency View.

Both GradleProjectImporter and update project configuration failed with exception. See [log.txt](https://github.com/redhat-developer/vscode-java/files/4784679/log.txt) for more details.

testforstephen:

Close #1495

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,11 +109,14 @@ The following settings are supported:
 * `java.contentProvider.preferred` : Preferred content provider (see 3rd party decompilers available in [vscode-java-decompiler](https://github.com/dgileadi/vscode-java-decompiler)).
 * `java.import.exclusions` : Exclude folders from import via glob patterns. Use `!` to negate patterns to allow subfolders imports. You have to include a parent directory. The order is important.
 * `java.import.gradle.enabled` : Enable/disable the Gradle importer.
-* `java.import.gradle.home`: setting for GRADLE_HOME.
+* Specify the Gradle distribution used by the Java extension:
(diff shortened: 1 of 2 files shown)


testforstephen:
// @akaroml @jdneo @Eskibear

Eskibear:
> `java.import.gradle.version`: Use Gradle from the specific version if the Gradle wrapper is missing or disabled.

Just to confirm that, if `java.import.gradle.wrapper.enabled` is true, we don't honor `java.import.gradle.version`, right?

testforstephen:
Also f your project has gradle wrapper, then we don't respect `java.import.gradle.version`.

The extension follows the order below to find the gradle distribution:
1. Check whether gradle.wrapper.enable preference is enabled and your workspace contains gradlew script.
2. Check gradle.version perference.
3. Check gradle.home preference and GRADLE_HOME environment variable.

fbricon:
Thanks @testforstephen

PR #1537, opened 2020-07-20
Declare semantic tokens in contribution points

0dinD:
Relates to eclipse/eclipse.jdt.ls#1501

This PR ensures backwards compatibility for the "old" semantic token types and enables more flexible styling of semantic tokens by declaring their supertypes in the `semanticTokenTypes` contribution point. For example, with the changes to the semantic highlighting I made in eclipse/eclipse.jdt.ls#1501, styling `type` will yield no results, since the language server now gives more specific tokens like `class`, `interface` and `enum`. By adding a supertype of `type` to these three tokens, they can still be styled with the "old" `type` token type.

Here is a more detailed view of the type hierarchy I have declared for the new semantic token types in eclipse/eclipse.jdt.ls#1501:
- `namespace`
- `type`
  - `class`
  - `interface`
  - `enum`
  - `annotation`
  - `typeParameter`
- `member`
  - `function`
    - `annotationMember`
  - `property`
    - `enumMember`
- `variable`
  - `parameter`

This hierarchy should probably be documented in the semantic highlighting wiki page, but I will leave that up to someone else, since I don't know if or how I can contribute to the wiki.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,108 @@
   ],
   "main": "./dist/extension",
   "contributes": {
+    "semanticTokenTypes": [
+      {
+        "id": "namespace",
+        "description": "Style for packages."
+      },
+      {
+        "id": "type",
+        "description": "Style for types."
+      },
+      {
+        "id": "class",
+        "superType": "type",
+        "description": "Style for classes."
+      },
+      {
+        "id": "interface",
+        "superType": "type",
+        "description": "Style for interfaces."
+      },
+      {
+        "id": "enum",
+        "superType": "type",
+        "description": "Style for enums."
+      },
+      {
+        "id": "annotation",
+        "superType": "type",
+        "description": "Style for annotations."
+      },
+      {
+        "id": "typeParameter",
+        "superType": "type",
+        "description": "Style for type parameters."
+      },
+      {
+        "id": "member",
+        "description": "Style for members."
+      },
+      {
+        "id": "function",
+        "superType": "member",
+        "description": "Style for methods."
+      },
+      {
+        "id": "annotationMember",
+        "superType": "function",
+        "description": "Style for annotation members."
+      },
+      {
+        "id": "property",
+        "superType": "member",
+        "description": "Style for fields."
+      },
+      {
+        "id": "enumMember",
+        "superType": "property",
+        "description": "Style for enum members."
+      },
+      {
+        "id": "variable",
+        "description": "Style for variables."
+      },
+      {
+        "id": "parameter",
+        "superType": "variable",
+        "description": "Style for parameters."
+      }
+    ],
+    "semanticTokenModifiers": [
+      {
+        "id": "static",
+        "description": "Style for symbols that are static."
+      },
+      {
+        "id": "readonly",
+        "description": "Style for symbols that are final."
+      },
+      {
+        "id": "deprecated",
+        "description": "Style for symbols that are deprecated."
+      },
+      {
+        "id": "public",
+        "description": "Style for symbols that are public"
+      },
+      {
+        "id": "private",
+        "description": "Style for symbols that are private."
+      },
+      {
+        "id": "protected",
+        "description": "Style for symbols that are protected"
+      },
+      {
+        "id": "abstract",
+        "description": "Style for symbols that are abstract"
+      },
+      {
+        "id": "declaration",
+        "description": "Style for symbols that are declarations."
+      }
+    ],
     "languages": [
       {
         "id": "java",


fbricon:
@Eskibear can you review this please?

package.json
```diff
@@ -34,6 +34,108 @@
...
+      {
+        "id": "namespace",
+        "description": "Style for packages."
+      },
+      {
+        "id": "type",
+        "description": "Style for types."
+      },
+      {
+        "id": "class",
```
Eskibear, line 47: I agree that, for new subtypes, the inheritance (e.g. `annotation` is sub type of `type`) should be defined here.

But for standard tokens like `class`, I'm wondering whether vscode itself has defined / should define the inheritance. Otherwise they would be redundantly defined in multiple extensions. @aeschli What do you think?

package.json
```diff
@@ -34,6 +34,108 @@
...
+        "id": "typeParameter",
+        "superType": "type",
+        "description": "Style for type parameters."
+      },
+      {
+        "id": "member",
+        "description": "Style for members."
+      },
+      {
+        "id": "function",
```
Eskibear, line 76: For the moment, there's no detailed description of each standard token in [LSP3.16](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_semanticTokens). But per vscode's [token scope mapping](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#semantic-token-scope-map), `member` looks more like a subtype of `function`.

token: member | textMate scope: entity.name.function.member
-- | --

Besides "member function/method", we also have "member variable" here in Java, which is definitely not a function. Before defining the inheritance, we have to clarify the concept `member` first.  /cc @aeschli

0dinD:
I have looked into the issues Eskibear described in his review comments, and this is what I found:
- VS Code does not seem to declare any supertypes by default, see [these lines](https://github.com/microsoft/vscode/blob/33b51eb210d2ae1c67721df3cd5d1ab4b6a6da17/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L506).
- If an extension contributes a token type with an existing ID, that token type is overwritten. For example, if we declare that classes have a supertype of `type`, that will be true not only for `class` tokens provided by this extension, but also for `class` tokens provided by semantic highlighting in TypeScript or Javascript. 

In my opinion, there should be a way to declare supertypes for the default token types without it affecting other languages that might want to assign another supertype. For example, a method in Java is guaranteed to also be a member, so it makes sense to declare that type hierarchy. But in TypeScript or JavaScript, a function can also be a variable, so there is no clear supertype. Another example is annotations, which are functions in TypeScript but types in Java. If both of these supertypes are declared, only one of them will "win", overriding the other even for the language where another supertype is desired.

I think the best way to solve this problem would be to add a "language" property to all contributed token types and modifiers, like there is for token scope mapping, so that an extension can override the default tokens if necessary, but only for the language that it applies to. Another solution could be to only allow overriding the default token supertypes for the tokens provided by the extension that overrode them. For example, if vscode-java overrides some default token supertypes, it will only take effect for tokens also provided by vscode-java.

I can see that @aeschli is the author of the [tokenClassificationRegistry](https://github.com/microsoft/vscode/blob/master/src/vs/platform/theme/common/tokenClassificationRegistry.ts), which is the piece of code that handles the default token types and modifiers. Maybe you could take a look at implementing this?

Some clarification on the `member` type would be great. But in my opinion, it makes much more sense for the relationship between `member` and `function` to be the opposite of what the textMate scope suggests. Again, this is probably language specific to Java, so I think it would good to have a way to specify that when overriding default tokens.

fbricon:
This PR brings changes that are out-of-scope of vscode-java. Given there are no immediate regressions without applying it, I'll wait for more feedback from @aeschli before merging it.

0dinD:
Well, there is a slight problem introduced by adding the new `annotation` and `annotationMember` token types. Since they are not part of the default tokens provided by VS Code, they are not colored at all unless you have a theme that supports them. So those tokens look a bit worse now with the default theme:
annotation
annotationMember
Note that colors are applied to annotations in the body, by a textMate scope, which is what you see with `SuppressWarnings` in the second screenshot. So the problem of missing color for annotations only seems to exist for import statements and their members. 

Since we know that `annotation` and `annotationMember` tokens are not declared by default in VS Code, we could probably declare at least those two token types while we wait for what aeschli has to say. Again, I still think that we need to be able to declare this for Java files only, because it is entirely possible for a language like TypeScript to also introduce `annotation` tokens, but with a different supertype (in which case there would be a conflict).

@fbricon what do you think? Should I submit another pull request with just the `annotation` and `annotationMember` token types declared, in order to fix the missing colors? Or should we wait for this PR instead?

fbricon:
So without this PR, annotation stuff looks bad:
Screen Shot 2020-07-21 at 6 01 17 PM

with vscode-java 0.58.0 / this PR:
Screen Shot 2020-07-21 at 6 00 01 PM

fbricon:
@0dinD so yes please add annotations specific types

0dinD:
I have now opened an issue (microsoft/vscode#103097) for allowing declaration of language-specific token types and modifiers in the contribution points. When/If it gets implemented, I could update this PR to declare these tokens for the `java` language only, which would avoid conflicts with other semantic token providers and the default token types.

0dinD:
After the discussion with aeschli in microsoft/vscode#103097, some things have been cleared up which makes me think that this PR is no longer a good idea. The main concern I initially had was about using the `member` token type, which I thought was way too generalized to use unless it could be supertyped to `function` and `property`.

This is what aeschli said about the `member` token type in our conversation.
> `member` is only intended for member functions. Maybe a clearer name would have been `memberFunction`.

So I just misunderstood the meaning of the `member` token type, because of the unclear name.

With that cleared up, I feel like it's time to close this PR. I will submit another PR to change the `function` token type to `member` in eclipse.jdt.ls, when I have time.

PR #1539, opened 2020-07-21
Request semantic tokens for class files

### LINKED ISSUE #1538: No semantic tokens are available in class files
When viewing a class file, such as one from an external library, no semantic tokens are available. 

##### Steps To Reproduce
1. Reference a class from a library in your Java project. (It could be from the Java standard library)
2. Use CTRL + click on the class name to navigate to said class declaration.

##### Current Result
The opened class file does not have any semantic tokens, even with semantic highlighting enabled.

##### Expected Result
The opened class file has semantic tokens, if semantic highlighting is enabled.

0dinD:
Fixes #1538
Requires eclipse/eclipse.jdt.ls#1511

This is a very simple change to the semantic tokens provider to make it request semantic tokens not only for source files, but also for class files in libraries, using the `jdt://...` protocol.

### DIFF
--- a/src/semanticTokenProvider.ts
+++ b/src/semanticTokenProvider.ts
@@ -21,7 +21,11 @@ export function registerSemanticTokensProvider(context: vscode.ExtensionContext)
     }
     if (isSemanticHighlightingEnabled()) {
         getSemanticTokensLegend().then(legend => {
-            const semanticTokensProviderDisposable = vscode.languages.registerDocumentSemanticTokensProvider({ scheme: 'file', language: 'java' }, semanticTokensProvider, legend);
+            const documentSelector = [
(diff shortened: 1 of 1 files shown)


fbricon:
mmm I still don't get any semantic highlighting when opening String.class

0dinD:
@fbricon that is strange, for me it works like expected. Are you sure that you are using both the changes to vscode-java in this PR and also the changes to eclipse.jdt.ls I made in eclipse/eclipse.jdt.ls#1511? Even if vscode requests the semantic tokens for class files, the language server must also be able to provide them, which is what that PR addresses.

fbricon:
For sure, it works much better when applying both PRs ;-)

fbricon:
Thanks @0dinD!

PR #1540, opened 2020-07-21
Declare semantic token types for annotations

0dinD:
This is a partial implementation of #1537 , in order to fix token colorization for annotations whilst not causing conflicts with the default token types.


fbricon:
Thanks @0dinD!

PR #1542, opened 2020-07-22
Add cache for markdown provider

jdneo:
This PR is to fix a bug for the markdown preview.

When using the API `markdown.api.render` to render the same markdown string to html more than once. The `id` field in the rendered result will be appended `-1`, `-2`, ...

Which means, if users opens the document more than once in the same session, he cannot navigate from TOC.

So a cache is introduced here. This will also boost the loading time since the render process will be skipped if the cache hits.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -35,6 +35,11 @@ export namespace Commands {
     export const GOTO_LOCATION = 'editor.action.goToLocations';
 
     /**
+     * Render markdown string to html string
+     */
+    export const MARKDOWN_API_RENDER = 'markdown.api.render';
+
+    /**
      * Update project configuration
      */
     export const CONFIGURATION_UPDATE = 'java.projectConfiguration.update';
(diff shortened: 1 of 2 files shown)


fbricon:
Thanks for the fix @jdneo!
However, I don't think having a cache is the most appropriate solution in the long term. That's unnecessary memory usage for a very infrequent operation. Ideally, the rendering command should be useable with a parameter that prevents generation of those unique ids. You should ask upstream vscode about it.

jdneo:
> Thanks for the fix @jdneo!
> However, I don't think having a cache is the most appropriate solution in the long term. That's unnecessary memory usage for a very infrequent operation. Ideally, the rendering command should be useable with a parameter that prevents generation of those unique ids. You should ask upstream vscode about it.

@fbricon  Make sense. Just created one: https://github.com/microsoft/vscode/issues/103076

Issue #1517, opened 2020-07-10
REFACTOR SETTING NOT WORKING  --  Java>Refactor: Rename From File Explorer


maleko48:
#### Environment #####
- Operating System: Windows 10 Home x64 v1909 / Build: 18363.900
- JDK version: 1.8.0_161
- Visual Studio Code version: 1.47.0
- Java extension version:  WORKS in 0.62.0 - 0.63.0  //  BROKEN in v0.64.0 - 0.64.1

##### Steps To Reproduce #####
1.)  Set Java>Refactor: Rename From File Explorer setting to: "AutoApply"

2.)  Create new java file: test.java via VSCode File Explorer

3.)  Save test.java

4.)  Select test.java via VSCode File Explorer

5.)  Press F2 to rename test.java to test1.java and complete rename by pressing ENTER

6.)  class name in test1.java should automatically refactor from "public class test" to "public class test1" along with other relevant references in the class definition, but does not.

testforstephen:
@maleko48 Nice catch!

PR #1574, opened 2020-08-12
Add test case for rename file

jdneo:
This is to prevent the regression in the future, for the issue: #1517

### DIFF
--- a/test/standard-mode-suite/rename.test.ts
+++ b/test/standard-mode-suite/rename.test.ts
@@ -1,0 +1,60 @@
+'use strict';
+
+import * as path from 'path';
+import * as fse from "fs-extra";
+import { Uri, extensions, TextDocument, workspace, WorkspaceEdit } from 'vscode';
+import { getJavaConfiguration } from '../../src/utils';
+import { constants } from '../common';
+
+const originFilePath: string = path.join(constants.projectFsPath, 'src', 'main', 'java', 'java', 'Foo.java');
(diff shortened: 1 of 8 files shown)


test/standard-mode-suite/rename.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+	suiteSetup(async function() {
+		getJavaConfiguration().update("refactor.renameFromFileExplorer", "autoApply");
+		await extensions.getExtension('redhat.java').activate();
+	});
+
+	test('rename on file will update the class name', async function () {
+		const workspaceEdit: WorkspaceEdit = new WorkspaceEdit();
+		workspaceEdit.renameFile(Uri.file(originFilePath), Uri.file(newFilePath));
+		await workspace.applyEdit(workspaceEdit);
```
testforstephen, line 24: The unit test will modify the test project file in the source code directly. Can you copy it to some temporary directory and modify there?

fbricon (changes requested):
Change the test so that it doesn't mutate the sources

test/standard-mode-suite/rename.test.ts
jdneo: ok

jdneo:
@fbricon Would you mind to tag this PR to a later milestone? I need sometime to figure out if there is some elegant way to prepare the test folder and launch the tests. (Launch from `npm run test` and VS Code launch configuration have different entrances).

jdneo:
Updated

fbricon:
Thanks @jdneo!

PR #1578, opened 2020-08-13
Added sysout

devmohit-live:
Println snippet for java

### DIFF
--- a/snippets/java.json
+++ b/snippets/java.json
@@ -17,6 +17,13 @@
 		],
 		"description": "Public constructor"
 	},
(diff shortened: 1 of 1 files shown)


fbricon:
sysout snippet is already provided by the server

fbricon:
Sorry but this PR is redudant, as the snippet is already available even while the server is starting: See https://github.com/redhat-developer/vscode-java/blob/f3b402671b746e84faadbf616edb27658a6fc265/snippets/server.json#L2-L8

Closing.

Issue #1571, opened 2020-08-11
(Extension API) Detecting if language server will launch in Standard Mode during extension activation


rzgry:
For MicroProfile tools for vscode (https://github.com/redhat-developer/vscode-microprofile) we require that the java language server is started in standard mode, and we want to display a warning / prompt users to switch to standard if they are in lightweight. We want to do this when MicroProfile tools activates before it launches its own language server that depends on vscode-java.

I am wondering the best way to do this.

Looking at the `serverMode` from vscode-java's api (https://github.com/redhat-developer/vscode-java/blob/master/src/extension.api.ts#L103). Before a project has been imported `serverMode` is returned as `LightWeight` initially and then switches over to `Standard` later if the user selects to import the project. So if i am checking for LightWeight while MicroProfile tools is starting up and display a warning, the user gets an unnecessary warning while the language server is starting up. Because MicroProfile tools will see that the user is in `LightWeight` mode and display a warning to switch to `Standard` when really the Standard mode server is in the process of being launched.

I also looked at using the `java.server.launchMode` setting to see if it is set to lightweight, since for the most part if it is set to `Hybrid` or `Standard` the language server will eventually end up in standard mode. The one problem I see with this is that before a project is imported if someone has their launch mode set to hybrid, and they decide to click import later  i think the server will stay in lightweight mode. So vscode-microprofile will not work and they would not get an error message in that case since the `java.server.launchMode` is set to `Hybrid`

Is there a way to detect if the java language server has launched in standard more or is in the process of launching in standard mode so I do not need to display a warning to the user.

@fbricon mentioned that @testforstephen, @jdneo and @Eskibear may be able to provide some guidance

jdneo:
Let's say if the user set the launch mode to Hybrid, and for that moment, lightweight server is ready and standard server is still activating. you will get a notification that `serverMode` is in `hybrid`:

https://github.com/redhat-developer/vscode-java/blob/master/src/extension.ts#L274-L276

Once the standard server is ready, another notification is emitted that `serverMode` changes to `standard`:

https://github.com/redhat-developer/vscode-java/blob/d6b090fcaa40ffe12c15caf4ec63e891d9709a79/src/standardLanguageClient.ts#L88

So I think you can get the initial value of the `serverMode` in the resolved API. And listen to the `onDidServerModeChange` event. Only activate the features of your MicroProfile tools when it's standard.

Feel free to let us know if you have more question. :)

rzgry:
[...]

PR #1581, opened 2020-08-14
bugfix: Emit Hybrid mode if both language server is running

jdneo:
fix an issue found in #1571.

The `hybrid` should be emitted when Lightweight server is ready and standard server is going to start.

So I moved the emit action into the function which starts the standard server, and only emit the event if the current server mode is lightweight.


fbricon:
Thanks @jdneo!

PR #1584, opened 2020-08-19
enable semantic highlighting by default

Eskibear:
Semantic highlighting has been polished for a while, and most theme authors have added related support. Looks it's not bad to enable it by default.


fbricon:
Agreed, it's much better now than when we introduced it. Thanks @Eskibear

PR #1586, opened 2020-08-20
Document 'Hybrid' serverMode value in api

rzgry:
Found it confusing that the `serverMode` api could return any of the server mode types defined in the `ServerMode ` enum, but the extension api docs only mention Standard and Lightweight. 

Signed-off-by: Ryan Zegray <ryan.zegray@ibm.com>


testforstephen:
@jdneo Can you review this PR?

PR #1591, opened 2020-08-24
Support CancellationToken when calling LSP workspaceCommand

testforstephen:

This PR makes the workspace command cancellable. The usage is passing an instance of vscode.CancellationToken as the last parameter of the workspace command. Cancel the token will cancel the LSP request.

A sample:
```ts
vscode.commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, commands.JAVA_RESOLVE_MAINMETHOD, uri.toString(), token);
```


testforstephen:
// @jdneo @Eskibear 
The codenlens and completion provider registered at vscode side can benefit from the cancellation mechanism. Also you should check the progress monitor to return early at the java plugin command handler.

fbricon:
@testforstephen thanks!

PR #1599, opened 2020-08-29
Punctuation fix in import message

pcantrell:
Minor punctuation fix: these are two independent clauses, and thus joining them with a comma is incorrect in formal written English.

    The workspace contains Java projects, would you like to import them?
                                        ↑

An alternative to the one in this PR would be two separate sentences:

    The workspace contains Java projects. Would you like to import them?

…but I proposed the semicolon because it’s Java!

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -359,7 +359,7 @@ async function workspaceContainsBuildFiles(): Promise<boolean> {
 }
 
 async function promptUserForStandardServer(config: WorkspaceConfiguration): Promise<boolean> {
-	const choice: string = await window.showInformationMessage("The workspace contains Java projects, would you like to import them?", "Yes", "Always", "Later");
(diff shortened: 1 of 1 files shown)


fbricon:
I think it'll be more consistent (at least, less weird) to use a period instead of a semicolon.
Can you also please sign-off your commit? (see https://github.com/redhat-developer/vscode-java/pull/1599/checks?check_run_id=1045549072)

pcantrell:
That semicolon usage [is standard English](https://writing.wisc.edu/handbook/grammarpunct/semicolons/), though perhaps it looks strange to a French speaker? In any cause, I'm happy to split it into two sentences. Force-pushing the change shortly.

fbricon:
It's also standard French ;-) But we'd prolly need to  rewrite a whole bunch of sentences all over the place to be consistent with that style.

pcantrell:
Fair enough!

fbricon:
Thanks @pcantrell!

PR #1603, opened 2020-08-31
Remove mention of java.requirements.JDK11Warning from README

0dinD:
The JDK11 warning popup and it's related setting was completely removed last version, as far as I'm aware. But the README still mentions the `java.requirements.JDK11Warning` setting.


fbricon:
Thanks @0dinD

PR #1632, opened 2020-09-22
Fix semantic tokens offset due to document updates

### LINKED ISSUE #1597: Since July update cause java highlight break
I love to use VSC Java plugin but I had so much pain working with java now.
Now the highlight is glitchy, it don't highlight properly when Semantic Highlight is turn on.
Comment don't get highlighted :( I feel sad
Also I would appreciate it if it is easier to use junit assertEqual, when I try to open a single .java file, it cannot import org.junit

##### Environment
- JDK version: 14
- Visual Studio Code version: 1.48.2 before update also the same
- Java extension version: -

##### Steps To Reproduce
1. Turn on semantic highlight
2. start debugging

email

0dinD:
Fixes #1597, together with eclipse/eclipse.jdt.ls#1552

The fix is basically taken straight from the implementation of the TypeScript semantic token provider, [see this line](https://github.com/microsoft/vscode/blob/50cc1d0e9777f3f76f460c24feb139ca695ff506/extensions/typescript-language-features/src/languageFeatures/semanticTokens.ts#L85).

### DIFF
--- a/src/semanticTokenProvider.ts
+++ b/src/semanticTokenProvider.ts
@@ -26,7 +26,14 @@ export function registerSemanticTokensProvider(context: vscode.ExtensionContext)
 
 class SemanticTokensProvider implements vscode.DocumentSemanticTokensProvider {
     async provideDocumentSemanticTokens(document: vscode.TextDocument, token: vscode.CancellationToken): Promise<vscode.SemanticTokens> {
+        const versionBeforeRequest: number = document.version;
         const response = <any> await vscode.commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.PROVIDE_SEMANTIC_TOKENS, document.uri.toString());
+        const versionAfterRequest: number = document.version;
+
+        if (versionBeforeRequest !== versionAfterRequest) {
+            await waitForDocumentChangesToEnd(document);
+            throw new Error("busy");
+        }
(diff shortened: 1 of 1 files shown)


testforstephen (approved):
LGTM

fbricon:
travis chokes on

ERROR: /home/travis/build/redhat-developer/vscode-java/src/semanticTokenProvider.ts[50, 13]: Identifier 'iv' is never reassigned; use 'const' instead of 'let'.

0dinD:
Oh, sorry, I just copied the code from the TypeScript implementation. I'll fix it later today.

0dinD:
@fbricon I've fixed up the code now.

fbricon:
Thanks @0dinD, this seems to make a big difference, even without eclipse/eclipse.jdt.ls#1552

0dinD:
No problem!
I think the reason why this PR also mitigates the issue fixed in eclipse/eclipse.jdt.ls#1552 is because the extension now waits until the document changes have settled before actually applying any semantic tokens from the server. But still, there could be an edge case where we do need eclipse/eclipse.jdt.ls#1552, so as long as there is no problem with that PR I think it's a good idea to also apply that fix. Better safe than sorry, right?

When the document and compilation unit become out-of-sync on the server side without the client knowing, the token offset will become "stuck" until you change the document again, which is very annoying.

PR #1636, opened 2020-09-28
Declare semantic token type for modifier keywords

0dinD:
Part of eclipse/eclipse.jdt.ls#1539.

Declares semantic token type for `modifierKeyword`, and supertypes it to `keyword`. Also adds [scope mapping](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#semantic-token-scope-map) for the custom token types, since that was missing. The scope mapping is required in order for the default themes to look correct with the new `modifierKeyword`, and also means that themes still using TextMate scopes can style modifier keywords, annotations and annotation members.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,28 @@
         "id": "annotationMember",
         "superType": "function",
         "description": "Style for annotation members."
+      },
+      {
+        "id": "modifier",
+        "superType": "keyword",
+        "description": "Style for modifier keywords."
+      }
+    ],
+    "semanticTokenScopes": [
+      {
+        "language": "java",
+        "scopes": {
+          "annotation": [
+            "storage.type.annotation.java"
+          ],
+          "annotationMember": [
+            "entity.name.annotationMember.java",
+            "constant.other.key.java"
+          ],
+          "modifier": [
+            "storage.modifier.java"
+          ]
+        }
       }
     ],
     "languages": [


fbricon:
@Eskibear please review when you can

package.json
```diff
@@ -45,6 +45,27 @@
...
+    ],
+    "semanticTokenScopes": [
+      {
+        "language": "java",
+        "scopes": {
+          "annotation": [
+            "storage.type.annotation.java"
+          ],
+          "annotationMember": [
+            "constant.other.key.java"
```
Eskibear, line 63: Taking `@SuppressWarnings(value="all")` for example. "value" is `annotationMember` and its definition is:
```java
public @interface SuppressWarnings {
   /*....*/
    String[] value();  //<---HERE
}

```

I wonder "constant" is not proper for it.

Eskibear (commented):
Agree that "modifier" inherits from "keyword", otherwise vscode doesn't know the token (althought it's defined in LSP3.16)

package.json
0dinD: I wasn't entirely sure what TextMate scope made the most sense, and just used the same scope as in [the default Java TextMate grammar](https://github.com/redhat-developer/vscode-java/blob/6f08b0437dd7e1b8f8b0be8185d75cb33a1697d3/syntaxes/java.tmLanguage.json#L154). If we decide to use some other scope it might break existing color themes, although I guess you could work around that by adding multiple TextMate scopes.

I do see another issue (maybe this is what you meant), which is that annotation member **declarations** are now given the same scope as annotation members inside of an annotation like `@SuppressWarnings(value = "all")`. It's definitely possible to add scope a different scope mapping for `annotationMember.declaration`, but the question is again which TextMate scope it should have?

In the default Java grammar, annotation member declarations are treated like regular methods, and therefore get the `entity.name.function.java` and `meta.function-call.java` scopes. We could add those scopes to annotation member declarations for backwards compatibility, but we could also add a more specific annotation member scope. Again, these alternatives doesn't have to be mutually exclusive as you can declare an array of TextMate scopes in the contribution point.

So what do you think? Should we mirror the behaviour of the default TextMate grammar (and tell users to style using semantic token rules) or should we also add more specific/correct scopes?

fbricon:
@Eskibear @0dinD does it make sense to merge eclipse/eclipse.jdt.ls#1539 independently from this? Or should we expect more changes upstream?

package.json
Eskibear: In current behavior, "value" is highlighted as function name, because you define `annotationMember` as a subtype of `function`. And it makes some sense because of the consistency, i.e., when you go to definition of "value", it's also highligted as a function.

My point is, although default textMate grammar use "constant.other.key.java", it is not proper for the scope. `constant.*` is for constant values like string literals, numbers, etc. See https://macromates.com/manual/en/language_grammars#naming_conventions

And it's not picked up by the default theme.
image


So I prefer here to use a more specific/correct TextMate scope, maybe something looks like `entity.name.annotationMember`?

fbricon:
@Eskibear can we merge this one?

Eskibear:
@fbricon LGTM, safe to merge.

fbricon:
Thanks @0dinD and @Eskibear

PR #1668, opened 2020-10-23
Pass the cancellationtoken to the client request registered by provider

testforstephen:


fbricon:
Thanks @testforstephen!

PR #1676, opened 2020-10-27
Change scope of 'java.home' to machine-overridable

### LINKED ISSUE #1667: Change the scope of 'java.home' to machine-overridable?
Currently the setting `java.home`'s scope is `window`. This may have some problems when it comes to remote scenario.

Here is an example:

WeChat Screenshot_20201022155632

My user(local) setting point the `java.home` to my local installation path, once my vscode linked to the remote container, it will be consumed and get this error.

After change to `machine-overridable`, it becomes fine

Local setting will not be taken by remote:
WeChat Screenshot_20201022155117

WeChat Screenshot_20201022155138

The next question is, if change to `machine-overridable`, will it still be configurable in the workspace level? The answer is yes:

WeChat Screenshot_20201022155329

This prevents to make breakings since currently the `java.home` is `window` scope and can be set to workspace as well.

jdneo:
resolve #1667 

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,11 +63,13 @@
         "path": "./snippets/java.json"
       }
     ],
-    "grammars": [{
-			"language": "java",
-			"scopeName": "source.java",
-			"path": "./syntaxes/java.tmLanguage.json"
-		}],
+    "grammars": [
+      {
+        "language": "java",
+        "scopeName": "source.java",
+        "path": "./syntaxes/java.tmLanguage.json"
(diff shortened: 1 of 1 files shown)


testforstephen (approved):
LGTM.

Since `java.home` points to an absolute path in current machine, naturally it's not shared across machines. `machine-overridable` looks fair, because it keeps the capability to override it with workspace settings, but stops it sharing to other machines.

Issue #1633, opened 2020-09-22
Problem messages can get offset by frequent changes to the document


0dinD:
When editing a Java file with problems in it, the underlines (reported problem positions) can sometimes become offset. Usually, this fixes itself after a second or so of delay, but in some rare cases the problem underline gets stuck in the wrong position until further changes are made to the document.

##### Environment
- Visual Studio Code version: 1.49.1
- Java extension version: 0.67.0

##### Steps To Reproduce
1. Open a Java class with problems in it (unused variable warnings, for example).
2. Make some changes in quick succession.
3. You should see the error messages jumping around and getting offset to incorrect positions.
4. In some rare cases, the error messages get stuck on the wrong location until a new character is typed.

##### Current Result
![problem-offset-bug-1](https://user-images.githubusercontent.com/52179873/93882362-6b391b80-fce0-11ea-9b33-df6f21509f70.png)
![problem-offset-bug-2](https://user-images.githubusercontent.com/52179873/93882376-6f653900-fce0-11ea-8499-56e39b87df1c.png)

##### Expected Result
The error messages (underlines) should always appear on the correct positions, and never get stuck on the wrong ones, like in the screenshot above.

##### Additional Informations
As you can see in the screenshots above, not only are the error messages offset, the semantic tokens are also offset in the exact same way (see #1597). I've submitted two PRs (#1632, eclipse/eclipse.jdt.ls#1552) to fix the semantic highlighting issue, but this issue with problem messages still persists since I only changed the semantic highlighting implementation. From what I can tell in the screenshots above, the two issues seem highly related, so a fix for this PR can probably be made by looking at how I fixed #1597.

This is how the semantic tokens look with #1632 and eclipse/eclipse.jdt.ls#1552 applied, I would expect the problem underlines to behave in the same way:
![problem-offset-bug-3](https://user-images.githubusercontent.com/52179873/93885574-932a7e00-fce4-11ea-8cb7-6e2d223b3170.png)

0dinD:
If anyone wants a more detailed showcase on how to reproduce the issue, [see my comment here](https://github.com/eclipse/eclipse.jdt.ls/pull/1552#issuecomment-698345369). Those instructions apply to this issue as well.

0dinD:
[...]

PR #1679, opened 2020-10-28
Refactor semanticTokenProvider.ts

### LINKED ISSUE #1678: Changing java.semanticHighlighting.enabled does not consistently update semantic tokens
Most java settings take immediate effect without requiring a reload, which is a great user experience. However, the `java.semanticHighlighting.enabled` setting does not always update the semantic tokens in the editor when changed. This is very confusing, especially when you aren't even notified that a reload is required to see the change.

##### Environment
- Java extension version: `0.69.0`

##### Steps To Reproduce, case 1
1. Make sure that semantic highlighting is enabled.
2. Open any Java file where semantic highlighting can be seen.
3. Set `java.semanticHighlighting.enabled` to `false`.
4. Notice how the opened Java file still has semantic tokens, until completely reloaded.

##### Current Result
semantic-highlighting-setting-1

##### Steps To Reproduce, case 2
1. Open any Java file where semantic highlighting can be seen.
2. Change any setting other than `java.semanticHighlighting.enabled`.
3. Change `java.semanticHighlighting.enabled` to `true` or `false`.
4. Notice how the setting does not take effect.

##### Current Result
semantic-highlighting-setting-2

##### Expected Result
semantic-highlighting-setting-3

##### Additional Information
I will submit a PR with a fix for this shortly.

0dinD:
Fixes #1678 

I don't think there's any way to do this without implementing `onDidChangeSemanticTokens`, as that seems to be the only way to force a refresh of semantic tokens. Part of #1678 was caused by `onceSemanticTokenEnabledChange` not registering itself after a change to a setting other than `java.semanticHighlighting.enabled`, so I refactored that code as well.

While I was at it, I decided to change the indentation to tabs as per the `.editorconfig`, and move `waitForDocumentChangesToEnd` to utils.ts, since it feels like more of a utility than something specific to the semantic tokens provider. It might have other use cases, such as fixing #1633, although I haven't looked closely at that particular case.

### DIFF
--- a/src/semanticTokenProvider.ts
+++ b/src/semanticTokenProvider.ts
@@ -1,90 +1,111 @@
 import * as vscode from 'vscode';
 import { Commands } from './commands';
-import { getJavaConfiguration } from './utils';
+import { getJavaConfiguration, waitForDocumentChangesToEnd } from './utils';
 
-const semanticHighlightingKey = 'java.semanticHighlighting.enabled';
+export function registerSemanticTokensProvider(context: vscode.ExtensionContext): void {
+	if (!vscode.languages.registerDocumentSemanticTokensProvider) { // in case Theia doesn't support this API
+		return;
+	}
 
-export function registerSemanticTokensProvider(context: vscode.ExtensionContext) {
-    if (!vscode.languages.registerDocumentSemanticTokensProvider) { // in case Theia doesn't support this API
-        return;
-    }
+	context.subscriptions.push(semanticTokensProvider);
+	context.subscriptions.push(vscode.workspace.onDidChangeConfiguration((e) => {
+		if (e.affectsConfiguration('java.semanticHighlighting.enabled')) {
+			if (isSemanticHighlightingEnabled()) {
+				semanticTokensProvider.enable();
+			}
+			else {
+				semanticTokensProvider.disable();
+			}
+		}
+	}));
 
-    if (isSemanticHighlightingEnabled()) {
-        getSemanticTokensLegend().then(legend => {
-            const documentSelector = [
-                { scheme: 'file', language: 'java' },
-                { scheme: 'jdt', language: 'java' }
-            ];
-            const semanticTokensProviderDisposable = vscode.languages.registerDocumentSemanticTokensProvider(documentSelector, semanticTokensProvider, legend);
-            context.subscriptions.push(semanticTokensProviderDisposable);
-            onceSemanticTokenEnabledChange(context, semanticTokensProviderDisposable);
-        });
-    } else {
-        onceSemanticTokenEnabledChange(context, undefined);
-    }
-}
-
-class SemanticTokensProvider implements vscode.DocumentSemanticTokensProvider {
-    async provideDocumentSemanticTokens(document: vscode.TextDocument, token: vscode.CancellationToken): Promise<vscode.SemanticTokens> {
-        const versionBeforeRequest: number = document.version;
-        const response = <any> await vscode.commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.PROVIDE_SEMANTIC_TOKENS, document.uri.toString());
-        const versionAfterRequest: number = document.version;
-
-        if (versionBeforeRequest !== versionAfterRequest) {
-            await waitForDocumentChangesToEnd(document);
-            throw new Error("busy");
-        }
-        if (token.isCancellationRequested) {
-            return undefined;
-        }
-        if (!response || !response.data) {
-            return undefined;
-        }
-        return new vscode.SemanticTokens(new Uint32Array(response.data), response.resultId);
-    }
-}
-
-function waitForDocumentChangesToEnd(document: vscode.TextDocument): Promise<void> {
-    let version = document.version;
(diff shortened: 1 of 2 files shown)


Eskibear:
Thanks for reporting this behavior and providing this PR.
Before looking deep into it, I want to share some information, and I don't think it's necessary to "fix" this for the moment. 
* When you disable semantic highlighting, AFAIK previous calculated tokens (or more precisely, color from the tokens) are still there before you close and reopen the file. It's more like a vscode's stuff, and I don't think it's a bad thing, thus no need to fix.
* Once LSP 3.16 is ready, we'll remove these providers on client side.

0dinD:
@Eskibear I see your point, but don't quite agree. 

> When you disable semantic highlighting, AFAIK previous calculated tokens (or more precisely, color from the tokens) are still there before you close and reopen the file. It's more like a vscode's stuff, and I don't think it's a bad thing, thus no need to fix.

Yes, this is the current behavior, tokens are not removed when you disable the `java.semanticHighlighting.enabled` setting. However, I do think this is bad user experience, seeing as nearly all other VS Code settings are responsively updated right as you change them. Additionally, the `editor.semanticHighlighting.enabled` **does** remove semantic tokens when disabled, meaning we are inconsistent with how that setting is handled. I think that setting is handled by internal VS Code APIs, but I came up with an alternative solution in this PR to make `java.semanticHighlighting.enabled` behavior consistent with `editor.semanticHighlighting.enabled`. 

> Once LSP 3.16 is ready, we'll remove these providers on client side.

Yes, but as far as I've seen it could be another while before LSP 3.16 is actually released and implemented in VS Code. I don't see a reason why we should wait around for LSP 3.16 before applying a fix, unless you know it's coming very soon? Also, I'm not sure why LSP 3.16 would actually help in fixing this issue? Unless there's any drawback to applying this PR, I think it's a good change in order to achieve a more consistent and thus better user experience.

Eskibear:
> Additionally, the editor.semanticHighlighting.enabled does remove semantic tokens when disabled, meaning we are inconsistent with how that setting is handled.

Agree that we should also remove tokens for consistency.


> I don't see a reason why we should wait around for LSP 3.16 before applying a fix.

Don't get me wrong. My point is, we don't need the "provider" stuff after adopting LSP 3.16. I value your work, so it might be a waste of effort to tweak something will soon be deleted. 

For LSP 3.16, corresponding client library `vscode-languageclient@7.0.0-next.12` is already there, we can give it a try. If it can perfectly fix the issue, we can simply switch to that.

0dinD:
I had a look at `vscode-languageclient@7.0.0-next.12`, but it seems like adopting it will require a little more work than I've got time for right now, seeing as there were some breaking changes to APIs since the version currently used (`6.0.0-next.9`). From my understanding it seems like `vscode-languageclient` is just a wrapper around the currently used APIs in `semanticTokenProvider.ts`, so I don't really think it will fix this problem for us.

Given the above, I propose we just apply this PR for now and wait until LSP 3.16 is finalized before converting the provider into one based on `vscode-languageclient` / LSP 3.16. If my work ends up being deleted, that's fine as long as the same functionality is achieved after LSP 3.16 is adopted. If anything, this PR can serve as a reference when converting the provider to something using `vscode-languageclient`, since I don't think clearing the tokens is currently addressed in `vscode-languageclient` or the LSP 3.16 spec.

If you want to fix this by adopting `vscode-languageclient@7.0.0-next.12` instead, that's also fine by me, I'm just saying that I don't have time right now, and it seems like the LSP 3.16 spec is still not quite finalized.

Eskibear:
OK, I'll take a looks at this PR before we switch to the latest client lib.

To talk more about the topic but unrelated to this PR, since vscode provides `editor.semanticHighlighting.enabled`, I'm thinking whether we still need `java.semanticHighlighting.enabled`. Now the only goal for this setting is, we can disable semantic highlighting for Java while keeping it enabled for other languages... And I don't see similar settings for other languages like TS/JS, python.

Eskibear (approved):
LGTM. It works as expected.

0dinD:
> I'm thinking whether we still need `java.semanticHighlighting.enabled`. Now the only goal for this setting is, we can disable semantic highlighting for Java while keeping it enabled for other languages...

Well, I guess it was initially justified because the semantic highlighting was experimental and could cause some issues. I'm personally not against removing it, but there could always be someone who has a use case for it, and I don't think there's any reason to remove the feature unless it causes some problem.

> And I don't see similar settings for other languages like TS/JS, python.

True, not all languages have their own setting for semantic highlighting. But I know at least the C/C++ and C# extensions have a setting for it, and there could be more that I haven't heard of.

testforstephen:
thanks @0dinD

0dinD:
FYI @Eskibear [the workaround](https://github.com/0dinD/vscode-java/blob/43c75027c54ec32a3fe1533815d3925bb3eceb12/src/semanticTokenProvider.ts#L86) I added in this PR where semantic tokens are manually cleared before disposing the provider is no longer necessary as of VS Code `1.52.0`, since microsoft/vscode#97196 was fixed. So you can probably remove the workaround whenever you adopt LSP 3.16 and simply dispose the provider when the setting is disabled.

PR #1680, opened 2020-10-28
Infer expressions if there is no selection range when extracting method

CsCherrYY:
Requires eclipse/eclipse.jdt.ls#1585

Will infer different expressions if there is no selection range when extracting method.
extract

### DIFF
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -332,17 +332,16 @@ export namespace GetRefactorEditRequest {
 export interface InferSelection {
     name: string;
     length: number;
-    startPosition: number;
+    offset: number;
 }
 
-export interface GetInferSelectionParams {
+export interface InferSelectionParams {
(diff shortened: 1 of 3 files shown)


src/protocol.ts
```diff
@@ -329,6 +329,21 @@ export namespace GetRefactorEditRequest {
     export const type = new RequestType<GetRefactorEditParams, RefactorWorkspaceEdit, void, void>('java/getRefactorEdit');
 }
 
+export interface InferSelection {
```
testforstephen, line 332: To keep the name consistent with the server code.
InferSelection -> SelectionInfo?

CsCherrYY: Sure.

testforstephen (approved):

testforstephen:
thanks @CsCherrYY

PR #1683, opened 2020-10-28
Add the java.configuration.maven.globalSettings property

### LINKED ISSUE #1365: Add the java.configuration.maven.globalSettings property
I'm very close to make the switch to VSCode from Eclipse, this is the last hurdle, but one I can't walk around. Please help.

**The Problem:** A working Maven project that builds fine with command line mvn will have many errors in pom once loaded in VS code. POM errors result in classes / types in dependency libraries not resolved in the editor.

**Note:** This seems to happen only within a corporate network using our own internal central repo server (like "https://internal.myfirm.com/repository/maven/"). Same code works at home, though I don't use any customized maven setting files like those at work.

Screenshot of the project error here:

vscode-err-mytest1

My guess is as a result of the error, no .classpath file is generated, and no dependencies are shown on the project tree on the left panel.

**What I don't understand is why would VSCode complain about connecting to the public Maven central repo while my maven settings point to our own repo server?**

If I replace both settings.xml with standard plain setting files, it pulls dependencies from the public maven central https://repo.maven.apache.org/maven2 and works just fine. Obviously I cannot do that in practice as I'm required to use corporate maven repo.

**Note**:
- mvn command works fine standalone, recognizing the customized maven setting files, downloading dependencies from the internal repo.
- No override of any maven settings is put into VScode
- Environment variables are set on the machine:
M2=C:\dev\maven\3.3.9\bin
M2_HOME=C:\dev\maven\3.3.9

**Environment**
OS: Windows10 build 1809, 64-bit
VSCode Version: 1.43.2
Electron: 7.1.11
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0

**Reference Files**
- See attached project
[mytest.zip](https://github.com/redhat-developer/vscode-java/files/4430596/mytest.zip)

- Maven setting files attached (masked)
[settings(global).txt](https://github.com/redhat-developer/vscode-java/files/4430597/settings.global.txt)
[settings(user).txt](https://github.com/redhat-developer/vscode-java/files/4430598/settings.user.txt)

- JDT log file:
[mytest.jdt.log](https://github.com/redhat-developer/vscode-java/files/4430603/mytest.jdt.log)

snjeza:
Fixes #1365 
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/1586

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ The following settings are supported:
 * `java.errors.incompleteClasspath.severity` : Specifies the severity of the message when the classpath is incomplete for a Java file. Supported values are `ignore`, `info`, `warning`, `error`.
 * `java.trace.server` : Traces the communication between VS Code and the Java language server.
 * `java.configuration.updateBuildConfiguration` : Specifies how modifications on build files update the Java classpath/configuration. Supported values are `disabled` (nothing happens), `interactive` (asks about updating on every modification), `automatic` (updating is automatically triggered).
-* `java.configuration.maven.userSettings` : Path to Maven's settings.xml.
+* `java.configuration.maven.userSettings` : Path to Maven's user settings.xml.
 * `java.configuration.checkProjectSettingsExclusions`: Checks if the extension-generated project settings files (`.project`, `.classpath`, `.factorypath`, `.settings/`) should be excluded from the file explorer. Defaults to `true`.
 * `java.referencesCodeLens.enabled` : Enable/disable the references code lenses.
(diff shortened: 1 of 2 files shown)


rgrunber:
I would probably update the description for "java.configuration.maven.userSettings" (in package.json and README.md) to mention that it is for "user" settings, now that we also have global settings.

snjeza:
> I would probably update the description for "java.configuration.maven.userSettings" (in package.json and README.md) to mention that it is for "user" settings, now that we also have global settings.

It's fixed.

Issue #1700, opened 2020-11-17
Error in every java file after updating to macOS Big Sur


shucshin:
I started getting this message at the beginning of every java file that i open in vscode saying:

¨The type java.lang.Object cannot be resolved. It is indirectly referenced from required .class filesJava(16777540)¨

The extension that causes this error is Language Support for Java™ by Red Hat

The issue appeared after the software update to macOS Big Sur and installing Windows Bootcamp. These two actions are the only things I did that i can think of since before committing those actions there was no problem. Please help.

- Operating System: macOS Big Sur Version 11.0.1 (20B29)
- JDK version: 
java 15.0.1 2020-10-20
Java(TM) SE Runtime Environment (build 15.0.1+9-18)
Java HotSpot(TM) 64-Bit Server VM (build 15.0.1+9-18, mixed mode, sharing)

- Visual Studio Code version: 1.51.1
- Java extension version: v0.70.0

<img width="716" alt="Screen Shot 2020-11-14 at 6 26 08 PM" src="https://user-images.githubusercontent.com/70228900/99324536-1f1adb00-283a-11eb-986d-5597c99e5815.png">
<img width="1027" alt="Screen Shot 2020-11-16 at 6 33 14 PM" src="https://user-images.githubusercontent.com/70228900/99324562-30fc7e00-283a-11eb-9401-a5f53d26b375.png">
<img width="1034" alt="Screen Shot 2020-11-16 at 6 33 27 PM" src="https://user-images.githubusercontent.com/70228900/99324580-38bc2280-283a-11eb-9a3b-ca84dfd33e02.png">

jnalbert:
I am having the same problem and I don't know how to fix it.

Eskibear:
I'm updating to Big Sur now, and I'll let you know if I can reproduce it or not. 
BTW, is your workspace a 1) maven project or 2) gradle project or 3) simple folder containing .java files?

shucshin:
In my case it's been both simple folders containing .java files and maven projects,(don't know about gradle projects since I don't have one) for maven projects it would create a .settings folder containing files with names starting with org.eclipse.something; for simple folders containing only .java files it wouldn't create a .settings file but still with the same errors as posted.

summerfang:
I have the same problem and not able to fix it yet.

Eskibear:
I've updated to Big Sur, but I cannot reproduce it. 
Maybe you can try command `Java: Clean the Java language server workspace`  from command palette.

[...]

PR #1701, opened 2020-11-18
search all installed JDKs when java.home not specified

### LINKED ISSUE #1621: potential failure of activation when JDK <11 is also installed
### Context
It simply finds the first available JDK in below order:
* java.home
* JDK_HOME
* JAVA_HOME
* fallback to call [node-find-java-home](https://github.com/jsdevel/node-find-java-home#algorithm)

### Steps
Now since it requires JDK 11+, a broken case is:
* both JDK 8 and JDK 11 installed 
* `java.home` not set (for most new users)
* JAVA_HOME points to JDK 8

It will pick JDK 8, and request users to install JDK 11+.

### Possible solution

Add new options to find-java-home, like `minVersion` (requires the library to check version), or request the library to return all possible java homes.

Eskibear:
In macOS Big Sur, `/usr` is wronly recongnized as JDK homedir. See https://github.com/redhat-developer/vscode-java/issues/1700#issuecomment-729478810 

It's caused by defect of upstream lib node-find-java-home. As a hotfix, here I copied a file from https://github.com/microsoft/vscode-java-pack/blob/fc6cb66f496fbbe5e8b2d9971c430e081f5b5460/src/java-runtime/utils/findJavaRuntime.ts

By validating all installed JDKs, this PR can also fix #1621 

Now the JDK detection follows below order: 
* if `java.home` is specified, we validate it, and if it's invalid, we prompt error message as before 
* if `java.home` is not specified, we search all possible places(env.JAVA_HOME, env.PATH, Windows Registry, Common installation directories for popular distributions) for installed JDKs, pick a valid one (major version >= 11) if exists

### DIFF
--- a/src/findJavaRuntimes.ts
+++ b/src/findJavaRuntimes.ts
@@ -26,8 +26,9 @@ export async function findJavaHomes(): Promise<JavaRuntime[]> {
     const ret: JavaRuntime[] = [];
     const jdkMap = new Map<string, string[]>();
 
-    updateJDKs(jdkMap, await fromJavaHomeEnv(), "env.JAVA_HOME");
-    updateJDKs(jdkMap, await fromPathEnv(), "env.PATH");
+    updateJDKs(jdkMap, await fromEnv("JDK_HOME"), "env.JDK_HOME");
+    updateJDKs(jdkMap, await fromEnv("JAVA_HOME"), "env.JAVA_HOME");
+    updateJDKs(jdkMap, await fromPath(), "env.PATH");
     updateJDKs(jdkMap, await fromWindowsRegistry(), "WindowsRegistry");
     updateJDKs(jdkMap, await fromCommonPlaces(), "DefaultLocation");
 
@@ -59,10 +60,10 @@ function updateJDKs(map: Map<string, string[]>, newJdks: string[], source: strin
     }
 }
 
-async function fromJavaHomeEnv(): Promise<string[]> {
+async function fromEnv(name: string): Promise<string[]> {
     const ret: string[] = [];
-    if (process.env.JAVA_HOME) {
-        const javaHome = await verifyJavaHome(process.env.JAVA_HOME, JAVAC_FILENAME);
+    if (process.env[name]) {
+        const javaHome = await verifyJavaHome(process.env[name], JAVAC_FILENAME);
         if (javaHome) {
             ret.push(javaHome);
         }
@@ -70,7 +71,7 @@ async function fromJavaHomeEnv(): Promise<string[]> {
     return ret;
 }
 
-async function fromPathEnv(): Promise<string[]> {
+async function fromPath(): Promise<string[]> {
     const ret: string[] = [];
 
     const paths = process.env.PATH ? process.env.PATH.split(path.delimiter).filter(Boolean) : [];
--- a/src/findJavaRuntimes.ts
+++ b/src/findJavaRuntimes.ts
@@ -1,0 +1,334 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as cp from "child_process";
+import * as fse from "fs-extra";
+import * as _ from "lodash";
+import * as os from "os";
+import * as path from "path";
+const expandHomeDir = require("expand-home-dir");
+const WinReg = require("winreg-utf8");
+const isWindows: boolean = process.platform.indexOf("win") === 0;
+const isMac: boolean = process.platform.indexOf("darwin") === 0;
+const JAVAC_FILENAME = "javac" + (isWindows ? ".exe" : "");
+const JAVA_FILENAME = "java" + (isWindows ? ".exe" : "");
+
+export interface JavaRuntime {
+    home: string;
+    version: number;
+    sources: string[];
+}
+
+/**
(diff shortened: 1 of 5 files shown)


Eskibear:
https://github.com/redhat-developer/vscode-java#setting-the-jdk should also be updated in this PR. I'll do that after you review it.

src/findJavaRuntimes.ts
```diff
@@ -0,0 +1,334 @@
...
+ * return metadata for all installed JDKs.
+ */
+export async function findJavaHomes(): Promise<JavaRuntime[]> {
+    const ret: JavaRuntime[] = [];
+    const jdkMap = new Map<string, string[]>();
+
+    updateJDKs(jdkMap, await fromJavaHomeEnv(), "env.JAVA_HOME");
+    updateJDKs(jdkMap, await fromPathEnv(), "env.PATH");
+    updateJDKs(jdkMap, await fromWindowsRegistry(), "WindowsRegistry");
+    updateJDKs(jdkMap, await fromCommonPlaces(), "DefaultLocation");
```
testforstephen, line 32: >The path to the Java Development Kit is searched in the following order:
>- the java.home setting in VS Code settings (workspace then user settings)
>- the JDK_HOME environment variable
>- the JAVA_HOME environment variable
>- on the current system path

This is from our wiki. 

If you could add `updateJDKs(jdkMap, await fromJavaHomeEnv(), "env.JDK_HOME");`, that would be more compatible with the old behavior.

src/findJavaRuntimes.ts
```diff
@@ -0,0 +1,334 @@
...
+        } else {
+            map.set(newJdk, [source]);
+        }
+    }
+}
+
+async function fromJavaHomeEnv(): Promise<string[]> {
+    const ret: string[] = [];
+    if (process.env.JAVA_HOME) {
+        const javaHome = await verifyJavaHome(process.env.JAVA_HOME, JAVAC_FILENAME);
```
testforstephen, line 65: should support process.env.JDK_HOME too.

src/requirements.ts
```diff
@@ -33,9 +34,42 @@ interface ErrorData {
...
+                } else {
+                    msg = `The ${source} (${javaHome}) does not point to a JDK.`;
+                }
+                invalidJavaHome(reject, msg);
+            }
+            javaVersion = await getJavaVersion(javaHome);
+        } else {
+            // java.home not specified, search valid JDKs from env.JAVA_HOME, env.PATH, Registry(Window), Common directories
+            const javaRuntimes = await findJavaHomes();
+            const requiredJdk = javaRuntimes.find(r => r.version >= REQUIRED_JDK_VERSION);
```
testforstephen, line 60: Currently it may return the first JDK randomly, I would prefer to return the first qualified JDK with the source order such as JDK_HOME > JAVA_HOME > System Path > Default Location.

src/findJavaRuntimes.ts
Eskibear: resolved in https://github.com/redhat-developer/vscode-java/pull/1701/commits/ce0652c541f96fd91e6d3114f5149b1512459bc0

testforstephen (approved):
LGTM.

@rgrunber @fbricon Since multiple users are reporting Java errors due to wrong JDK detection after upgrading to macOS Big Sur, can we merge this PR into Nov Middle release? WDYT.

rgrunber:
Yes, feel free to merge. We can update release documents afterwards.

rgrunber:
I think there might be an issue with this change that ends up affecting Linux distros.

Have a look at https://github.com/redhat-developer/vscode-java/blob/master/src/findJavaRuntimes.ts#L247
In my case, 'proposed' was '/usr/lib/jvm/java-11-openjdk' . Clearly 'bin/javac' existed to get to this point. What failed for me was the '.isDirectory()' check. On my machine '/usr/lib/jvm/java-11-openjdk' is a symbolic link, so I'd assume the directory check fails.

It's fairly common for symbolic links in /usr/lib/jvm of the form `java-11-openjdk -> java-11-openjdk-11.0.9.11-0.fc32.x86_64`

Eskibear:
Exactly. Let me check if it's safe to simply remove the `isDirectory()` check.

Eskibear:
@rgrunber Is it a convention to install JDKs into `/usr/lib/jvm` ? If so, we can fallback to check this directory on Linux, if no JAVA_HOME specified..

Eskibear:
The `isDirectory()` check was originally added from this commit: https://github.com/jsdevel/node-find-java-home/commit/9bd6daf0461a604d67a4fddc0ae7c36a1e2aae86#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R129

It was not intended to fix any edge case, and I just tried after removing it, no problems found. So I just created a PR to remove it.

rgrunber:
> @rgrunber Is it a convention to install JDKs into `/usr/lib/jvm` ? If so, we can fallback to check this directory on Linux, if no JAVA_HOME specified..

This can vary between distros, so I don't think it holds in call cases, but it could be a good fallback. It's certainly the case for Fedora and Ubuntu.

Eskibear:
@rgrunber ok, I can add `/usr/lib/jvm` as fallback, to increase coverage.

PR #1719, opened 2020-11-26
Update language client to adopt the new capabilities such as resolveCodeAction

testforstephen:

This is used to adopt the feature https://github.com/eclipse/eclipse.jdt.ls/pull/1607


fbricon:
@rgrunber please check this is compatible with Che/Theia

PR #1760, opened 2021-01-10
Declare semantic token modifiers

0dinD:
Declares new semantic token types* for eclipse/eclipse.jdt.ls#1641, and adds scope mapping for them so that themes using TextMate rules won't break. I also added all the custom modifiers provided by jdt.ls to the contribution point, which means that theme authors can now get intellisense for those modifiers when editing `color-theme.json` files or the `editor.semanticTokenColorCustomizations` setting.

*after [discussion with Eskibear](https://github.com/eclipse/eclipse.jdt.ls/pull/1641#issuecomment-763603693), I decided not to add any new token types yet. However, scope mapping for Javadoc tags is still required to prevent them from being styled as generic control keywords (regressing the default themes).

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,36 @@
         "description": "Style for modifier keywords."
       }
     ],
+    "semanticTokenModifiers": [
+      {
+        "id": "public",
+        "description": "Style for symbols with the public access modifier."
+      },
+      {
+        "id": "private",
+        "description": "Style for symbols with the private access modifier."
+      },
+      {
+        "id": "protected",
+        "description": "Style for symbols with the protected access modifier."
+      },
+      {
+        "id": "native",
+        "description": "Style for symbols with the native modifier."
+      },
+      {
+        "id": "generic",
+        "description": "Style for symbols that are generic (have type parameters)."
+      },
+      {
+        "id": "typeArgument",
+        "description": "Style for symbols that are type arguments for a generic symbol."
+      },
+      {
+        "id": "importDeclaration",
+        "description": "Style for symbols that are part of an import declaration."
+      }
+    ],
     "semanticTokenScopes": [
       {
         "language": "java",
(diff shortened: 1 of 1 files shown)


0dinD:
Updated to match changes in eclipse/eclipse.jdt.ls#1641 made in response to [discussion with Eskibear](https://github.com/eclipse/eclipse.jdt.ls/pull/1641#issuecomment-763603693).

0dinD:
@Eskibear seeing as eclipse/eclipse.jdt.ls#1641 was merged, you should take a look at this. Without it, javadoc tags will be treated as generic keywords since the semantic token overrides the TextMate scope:
javadoc-tags

This PR scope maps `keyword.documentation` to the correct TextMate scope, and while we're at it I also added contributions for the custom jdt.ls token modifiers, so that theme authors will get intellisense for them:
intellisense

Eskibear:
According to https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide :
* It's fair that we declare the non-standard modifiers in package.json
* It's fair that we add scope mapping, keep consistency with previous behaviors. (but since `modifier` type is now one of standard LSP token types, I think it would be better if vscode handles its scope mapping)

LGTM, Thanks!

0dinD:
> but since `modifier` type is now one of standard LSP token types, I think it would be better if vscode handles its scope mapping

Agreed, I was going to submit an issue about it last time we discussed it, but forgot to. I've added [a comment about this](https://github.com/microsoft/vscode/issues/97063#issuecomment-767852061) to the microsoft/vscode#97063 discussion issue.

Eskibear:
@testforstephen @rgrunber please help to merge this PR before next release, otherwise in default theme `@param` would be wrongly highlighted as keywords (pink) like `new` `while` `if` etc, as shown in https://github.com/redhat-developer/vscode-java/pull/1760#issuecomment-766871642

testforstephen:
Sure. Since this is a paired PR against eclipse/eclipse.jdt.ls#1641, it's fair to merge it in one release together.

PR #1780, opened 2021-01-27
use lightweight mode by default when vscode is running from web browsers

Eskibear:
The purpose is to reduce pop-up notifications. 
Here we remove the notification, and by default only launches LS in lightweight mode if vscode is running from a web browser. It's based on the hypothesis that when users open vscode on a web browser they mainly want to browse code or do simple editing.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -4,7 +4,7 @@ import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
 import * as fse from 'fs-extra';
-import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, Selection, Position, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration } from 'vscode';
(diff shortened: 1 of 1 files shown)


fbricon:
@rgrunber make sure this change doesn't break the Che/Theia integration

rgrunber (commented):
This definitely affects Theia. `env.uiKind` is '2' (UIKind.Web) so the various prompts are skipped. Perhaps `env.appName` may need to be used to further narrow this down ?

Eskibear:
I just tested some scenarios on VS Code, below are possible values I found:
* env.uiKind
   * `1` for desktop
   * `2` for web
* env.appName
   * `Visual Studio Code`
   * `Visual Studio Code - Insiders`
* env.remoteName
  * `wsl` when using Remote-WSL
  * `ssh-remote` when using Remote-SSH
  * `codespaces` when using GitHub Codespaces 

So yes, we can check `env.appName` if you want to keep this notification in Che/Theia. Furthermore, with `env.remoteName` we actually have the potential to optimize for different scenarios.

akaroml:
I may suggest `env.uiKind === "web" && env.appName.contains("Visual Studio Code")`

Eskibear:
OK, I've updated the PR to check both `env.uiKind` and `env.appName`

rgrunber (approved):
Looks fine to me. I can confirm Theia is excluded.

testforstephen (approved):

PR #1797, opened 2021-02-07
[FIX] Fix java settings causes LSP issues in theia

veredcon:


fbricon:
What problem is it fixing in Theia? (while certainly breaking jdt.ls)

Sum1cares (approved):

PR #1815, opened 2021-02-24
Rework rename/move refactoring support

testforstephen:

It requires eclipse/eclipse.jdt.ls#1445

- Should use **onWillRenameFiles** listener to handle the rename refactoring because it supports returning an edit via `e.waitUntil` API. Also another benefit is that it will composite the edit with the undo stack of the file operation, that means clicking `ctrl + z` once will undo the file rename operation and the corresponding refactoring together.
- Remove the old preview feature. We're going to delegate the preview capability to the client itself, no need to handle it in the extension side any more.


testforstephen:
it's more stable in 1.53.

https://user-images.githubusercontent.com/14052197/108954720-358dc980-76a8-11eb-8981-c49b97cf4f9c.mov

snjeza (approved):

snjeza (approved):

jdneo (approved):

PR #1819, opened 2021-02-25
Configure whether to show recommendations of external installations

### LINKED ISSUE #1816: Configure whether to show recommendations of external installations
When opening a pom.xml file there is a pop up showing recommendation of dependency analytics:
image

related PR: https://github.com/redhat-developer/vscode-java/pull/1771/files

Is it possible to add configuration in settings.json so it won't show such recommendations at all? 
We know when choosing "Never" the choice is saved in the extension's global state but we don't want to show this info message at all in the first place. 

The reason is that in our landscape we have predefined/close set of installed extensions and not all external extensions can work out of the box. 

##### Environment
- Operating System:
- JDK version: 11
- Visual Studio Code version: 
- Java extension version:

##### Steps To Reproduce
1. Open pom.xml file and see recommendation notification to install dependency analytics

snjeza:
Fixes #1816 

You can use
```
"java.dependency.analytics.show": false
```

### DIFF
--- a/src/recommendation/dependencyAnalytics.ts
+++ b/src/recommendation/dependencyAnalytics.ts
@@ -8,12 +8,17 @@ import { IHandler } from "./handler";
 const EXTENSION_NAME = "redhat.fabric8-analytics";
 const GH_ORG_URL = `https://github.com/fabric8-analytics`;
 const RECOMMENDATION_MESSAGE = `Do you want to install the [Dependency Analytics](${GH_ORG_URL}) extension to stay informed about vulnerable dependencies in pom.xml files?`;
+const JAVA_DEPENDENCY_ANALYTICS_SHOW = "java.recommendations.dependency.analytics.show";
 
 function isPomDotXml(uri: vscode.Uri) {
 	return !!uri.path && uri.path.toLowerCase().endsWith("pom.xml");
 }
 
 export function initialize (context: vscode.ExtensionContext, handler: IHandler): void {
+	const show = vscode.workspace.getConfiguration().get(JAVA_DEPENDENCY_ANALYTICS_SHOW);
+	if (!show) {
+		return;
+	}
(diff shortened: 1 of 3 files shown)


src/recommendation/dependencyAnalytics.ts
```diff
@@ -8,12 +8,17 @@ import { IHandler } from "./handler";
 const EXTENSION_NAME = "redhat.fabric8-analytics";
 const GH_ORG_URL = `https://github.com/fabric8-analytics`;
 const RECOMMENDATION_MESSAGE = `Do you want to install the [Dependency Analytics](${GH_ORG_URL}) extension to stay informed about vulnerable dependencies in pom.xml files?`;
+const JAVA_DEPENDENCY_ANALYTICS_SHOW = "java.dependency.analytics.show"
```
rgrunber, line 11: ;

snjeza: Fixed

rgrunber: (though it build just fine for me)

rgrunber:
:+1: This works for me. It won't respond to changes in the setting after activation (so if you set it to false in settings but don't restart you'll still get the prompts) but I guess we don't need to go that far in this case.

@fbricon any issues with this option ?

src/recommendation/dependencyAnalytics.ts
fbricon: i'd prefer java.recommendations.dependency.analytics.show

fbricon:
workflow is fine

src/recommendation/dependencyAnalytics.ts
snjeza: Fixed.

veredcon:
Thanks for this enhancement!

PR #1824, opened 2021-03-04
Disable console.log for lsp trace due to performance issue

testforstephen:
See https://github.com/microsoft/vscode/issues/117327, when enabling `"java.trace.server": "verbose"`, opening a Java project in vscode@1.53 for a while, the whole UI process is frozen. The reason is the LSP will print lots of trace to the console of VS Code Dev Tools. To mitigate that, will disable console.log for LSP trace.

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -18,6 +18,8 @@ export const logger = createLogger({
 		format.prettyPrint()
 	),
 	transports: [
-		new transports.Console()
+		// See https://github.com/microsoft/vscode/issues/117327
+		// Disable console.log for lsp trace because intense logging will freeze the code render process.
+		// new transports.Console()
 	]
 });


snjeza (approved):

rgrunber (approved):
I'm able to reproduce and this fixes the issue.

rgrunber:
Out of curiosity, what console is this disabling, because from what I can tell the workspace client logs, and output console itself seem to be about the same size before and after this change ?

testforstephen:
> Out of curiosity, what console is this disabling, because from what I can tell the workspace client logs, and output console itself seem to be about the same size before and after this change ?

We printed the LSP log to three places: Output Channel, Client log files, Console of VS Code Dev Tools (**Help -> Toggle Developer Tools**). The performance culprit is from the last one.

PR #1836, opened 2021-03-17
Add support for Java 16

snjeza:
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/1691

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Features
 =========
 ![ screencast ](https://raw.githubusercontent.com/redhat-developer/vscode-java/master/images/vscode-java.0.0.1.gif)
 
-* Supports code from Java 1.5 to Java 15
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -78,6 +78,12 @@ If you need to compile your projects against a different JDK version, it's recom
     "path": "/path/to/jdk-15",
     "default": true
   },
+  ,
+  {
+    "name": "JavaSE-16",
```
testforstephen, line 83: Since this is only a sample configuration, how about replacing `JavaSE-15` with the new `JavaSE-16` instead of appending a new configuration to the sample.

testforstephen (approved):
Looks good. Only a minor feedback about the sample configuration part.

README.md
snjeza: Fixed.

Issue #1825, opened 2021-03-04
How can I add bazel support?


dbushenko:
Hello,

I'm trying to add bazel support to vascode-java using the lsp server. Unfortunately some hardcode in vscode-java prevents me from doing this. For instance, in standardLanguageClient.ts there is a function:

function isJavaConfigFile(path: String) {
	return path.endsWith('pom.xml') || path.endsWith('.gradle');
}

When the project file is updated, this function prevents from sending the command to eclipse.jdt.ls to update the project. More or less correct function should look like this:

function isJavaConfigFile(path: String) {
	return path.endsWith('pom.xml') || path.endsWith('.gradle')
		|| path.endsWith('BUILD') || path.endsWith('WORKSPACE');
}

But if the bazel support is not added to vscode, then this function will contain invalid behavior.

So my question is the following: what is the correct way to add bazel support in general and fix function isJavaConfigFile() in particular?

jdneo:
Seems that we need to have a new contribution point that allow other extension to register the build file pattern. (Something like what we have for `javaExtensions` in package.json)

### Proposal

Maybe a new entry named as `javaBuildFilePatterns`? For example:

```
"javaBuildFilePatterns": ["BUILD", "WORKSPACE"]
```

Any other ideas are open for discussion.

PR #1845, opened 2021-03-24
Add new contribution point that allows other extension to register the build file pattern.

LeonovecSergey:
Add new contribution point that allows other extension to register the build file pattern.

Close https://github.com/redhat-developer/vscode-java/issues/1825

Signed-off-by: Siarhei Leanavets siarhei_leanavets1@epam.com

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -5,6 +5,7 @@ import * as path from 'path';
 import { Commands } from './commands';
 
 let existingExtensions: Array<string>;
+let existingBuildFilePatterns: Array<string>;
 
 export function collectJavaExtensions(extensions: readonly vscode.Extension<any>[]): string[] {
 	const result = [];
@@ -26,7 +27,31 @@ export function collectJavaExtensions(extensions: readonly vscode.Extension<any>
 	return result;
 }
 
+export function collectBuildFilePattern(extensions: readonly vscode.Extension<any>[]) {
+	const result = [];
+	if (extensions && extensions.length) {
+		for (const extension of extensions) {
+			const contributesSection = extension.packageJSON['contributes'];
+			if (contributesSection) {
+				const buildFilePatterns = contributesSection['javaBuildFilePatterns'];
+				if (Array.isArray(buildFilePatterns) && buildFilePatterns.length) {
+					result.push(...buildFilePatterns);
+				}
+			}
+		}
+	}
+	// Make a copy of build file patterns:
+	existingBuildFilePatterns = result.slice();
+	return result;
+}
+
 export function onExtensionChange(extensions: readonly vscode.Extension<any>[]) {
+	updateExistingExtensions(extensions);
+	updateExistinguildFilePatterns(extensions);
+
+}
+
+function updateExistingExtensions(extensions: readonly vscode.Extension<any>[]) {
 	if (!existingExtensions) {
 		return;
 	}
@@ -52,4 +77,34 @@ export function onExtensionChange(extensions: readonly vscode.Extension<any>[]) 
 			}
 		});
 	}
+
+}
+
+function updateExistinguildFilePatterns(extensions: readonly vscode.Extension<any>[]) {
+	if (!existingBuildFilePatterns) {
+		return;
+	}
+	const oldPatterns = new Set(existingBuildFilePatterns.slice());
+	const newPatterns = collectBuildFilePattern(extensions);
+	let hasChanged = ( oldPatterns.size !== newPatterns.length);
+	if (!hasChanged) {
+		for (const newPattern of newPatterns) {
+			if (!oldPatterns.has(newPattern)) {
+				hasChanged = true;
+				break;
+			}
+		}
+	}
+
+	if (hasChanged) {
+		const msg = `Java build file patterns changed, reloading ${vscode.env.appName} is required for the changes to take effect.`;
+		const action = 'Reload';
+		const restartId = Commands.RELOAD_WINDOW;
+		vscode.window.showWarningMessage(msg, action).then((selection) => {
+			if (action === selection) {
+				vscode.commands.executeCommand(restartId);
+			}
+		});
+	}
+
 }
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -3,9 +3,9 @@
 import * as vscode from 'vscode';
 import * as path from 'path';
 import { Commands } from './commands';
+import { buildFilePatterns } from './standardLanguageClient';
 
 let existingExtensions: Array<string>;
-let existingBuildFilePatterns: Array<string>;
 
 export function collectJavaExtensions(extensions: readonly vscode.Extension<any>[]): string[] {
 	const result = [];
@@ -40,8 +40,6 @@ export function collectBuildFilePattern(extensions: readonly vscode.Extension<an
 			}
 		}
 	}
-	// Make a copy of build file patterns:
-	existingBuildFilePatterns = result.slice();
 	return result;
 }
 
@@ -72,13 +70,14 @@ function isExistingExtensionsUpdated(extensions: readonly vscode.Extension<any>[
 			}
 		}
 	}
+	return hasChanged;
 }
 
 function isExistingBuildFilePatternsUpdated(extensions: readonly vscode.Extension<any>[]) {
-	if (!existingBuildFilePatterns) {
+	if (!buildFilePatterns) {
 		return false;
 	}
-	const oldPatterns = new Set(existingBuildFilePatterns.slice());
+	const oldPatterns = new Set(buildFilePatterns.slice());
 	const newPatterns = collectBuildFilePattern(extensions);
 	const hasChanged = ( oldPatterns.size !== newPatterns.length);
 	if (!hasChanged) {
@@ -88,4 +87,5 @@ function isExistingBuildFilePatternsUpdated(extensions: readonly vscode.Extensio
 			}
 		}
 	}
+	return hasChanged;
 }
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -45,7 +45,7 @@ export function collectBuildFilePattern(extensions: readonly vscode.Extension<an
 }
 
 export function onExtensionChange(extensions: readonly vscode.Extension<any>[]) {
-	if (isContributedPartUpdated(collectJavaExtensions(extensions), existingExtensions) || isContributedPartUpdated(collectBuildFilePattern(extensions), buildFilePatterns)) {
+	if (isContributedPartUpdated(existingExtensions, collectJavaExtensions(extensions)) || isContributedPartUpdated(buildFilePatterns, collectBuildFilePattern(extensions))) {
 		const msg = `Java Extension Contributions changed, reloading ${vscode.env.appName} is required for the changes to take effect.`;
 		const action = 'Reload';
 		const restartId = Commands.RELOAD_WINDOW;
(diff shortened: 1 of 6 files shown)


jdneo:
Hi @LeonovecSergey,

I have two general comments about this PR.

1. `isBuildFilePattern()` is called every time `isJavaConfigFile()` is called. At least we can cache all the available patterns to avoid iterating all the extension metadata every time.
2. The `pattern` here should be defined more clearly, from the implementation it's a `suffix`. Maybe there is more space for improvement here. VS Code uses [regular expression](https://code.visualstudio.com/api/references/when-clause-contexts#conditional-operators) for its when clause. Can we also use that for our patterns?

src/plugin.ts
```diff
@@ -52,4 +77,34 @@ export function onExtensionChange(extensions: readonly vscode.Extension<any>[])
...
+	if (!hasChanged) {
+		for (const newPattern of newPatterns) {
+			if (!oldPatterns.has(newPattern)) {
+				hasChanged = true;
+				break;
+			}
+		}
+	}
+
+	if (hasChanged) {
```
jdneo, line 99: If both Java extension and file patterns are changed. Users will see two pop-up dialog asking for reload, which I think is a little bit annoying.

How about migrate the two function into one named something like `updateJavaExtensionContributions()`

src/standardLanguageClient.ts
```diff
@@ -466,19 +468,8 @@ function projectConfigurationUpdate(languageClient: LanguageClient, uri?: Uri) {
...
-function isBuildFilePattern(fileName: String): boolean {
-    return vscode
-            .extensions
-            .all
-            .filter(extension => extension.packageJSON['contributes'] && Array.isArray(extension.packageJSON['contributes']['javaBuildFilePatterns']))
-            .map(extension => extension.packageJSON['contributes']['javaBuildFilePatterns'])
-            .reduce((acc, val) => acc.concat(val), [])
-            .filter((pattern: String) => fileName.toLowerCase().endsWith(pattern.toLowerCase()))
-            .length > 0;
+	return buildFilePatterns.filter((pattern: String) => path.toLowerCase().endsWith(pattern.toLowerCase()))
```
jdneo, line 471: Here we need to figure out the rules for this new contribution point.

1. If we call it `javaBuildFilePatterns`, we need to use a pattern mechanism to check it, for example, regular expression.
2. If we just check it via `xxx.endWith(...)`, the field should be called `javaBuildFileNames`.

Personally I prefer the first one, because it has scalability for the future. And regular expression is well used in VS Code when clause.

jdneo (changes requested):

src/plugin.ts
LeonovecSergey: @jdneo done

src/standardLanguageClient.ts
LeonovecSergey: @jdneo done

src/plugin.ts
```diff
@@ -5,6 +5,7 @@ import * as path from 'path';
 import { Commands } from './commands';
 
 let existingExtensions: Array<string>;
+let existingBuildFilePatterns: Array<string>;
```
jdneo, line 8: I find there are two copies for the build file patterns (one here and another in src/standardLanguageClient.ts). Can we just have one? Maybe just use the one from standardLanguageClient.ts.

src/standardLanguageClient.ts
```diff
@@ -464,8 +467,10 @@ function projectConfigurationUpdate(languageClient: LanguageClient, uri?: Uri) {
 	}
 }
 
-function isJavaConfigFile(path: String) {
-	return path.endsWith('pom.xml') || path.endsWith('.gradle');
+function isJavaConfigFile(path: string) {
+	const fileName = require("path").basename(path);
```
jdneo, line 471: No need to use `require` here. The `path` module is already imported in this file (line 15.)

src/plugin.ts
```diff
@@ -26,30 +27,65 @@ export function collectJavaExtensions(extensions: readonly vscode.Extension<any>
...
-		vscode.window.showWarningMessage(msg, action).then((selection) => {
-			if (action === selection) {
-				vscode.commands.executeCommand(restartId);
+function isExistingBuildFilePatternsUpdated(extensions: readonly vscode.Extension<any>[]) {
+	if (!buildFilePatterns) {
+		return false;
+	}
+	const oldPatterns = new Set(buildFilePatterns.slice());
+	const newPatterns = collectBuildFilePattern(extensions);
+	const hasChanged = ( oldPatterns.size !== newPatterns.length);
```
jdneo, line 82: Just curious: what will happen when the sizes are equal but the contents are different? For example, update an extension that the build patterns registered by the extension changes.

jdneo:
Have no idea why CI fails. BTW, @LeonovecSergey do you know how we can verify this change manually?

src/plugin.ts
LeonovecSergey: For this case we have "if (!hasChanged)" where the content of the oldPatterns and the newPatterns is compared

LeonovecSergey:
> Have no idea why CI fails. BTW, @LeonovecSergey do you know how we can verify this change manually(like the steps using the Bazel extension to play around)?

I use debug breakpoints to track java extension and build file patterns changes 

1. Open vscode-java fork in VSCode
2. Launch as Extension
3. In the opened window open some maven or gradle project (e.g. https://github.com/gradle/gradle-build-scan-quickstart)
4. Update project (e.g shift+alt+U) for watching current state of java extensions and build file patterns
5. Install new VSCode Extension that will add java extension or build file patterns from contributes in package.json (You can find vsix file in attached zip document)
6. Click on the pop-up Reload button
7. Update project (e.g shift+alt+U) for watching current state of java extensions and build file patterns 
[bazel-ls-vscode-0.0.1.zip](https://github.com/redhat-developer/vscode-java/files/6274135/bazel-ls-vscode-0.0.1.zip)

src/plugin.ts
```diff
@@ -53,3 +55,20 @@ export function onExtensionChange(extensions: readonly vscode.Extension<any>[])
 		});
 	}
 }
+
+function isContributedPartUpdated(newContributedPart: Array<string>, oldContributedPart: Array<string>) {
+	if (!oldContributedPart) {
+		return false;
+	}
+	const oldExtensions = new Set(oldContributedPart.slice());
```
jdneo, line 63: Like the way we merge two functions into one, then I guess the variable name here should also be updated? Something like `oldContribution` `newContribution`?

jdneo:
> > Have no idea why CI fails. BTW, @LeonovecSergey do you know how we can verify this change manually(like the steps using the Bazel extension to play around)?
> 
> I use debug breakpoints to track java extension and build file patterns changes
> 
> 1. Open vscode-java fork in VSCode
> 2. Launch as Extension
> 3. In the opened window open some maven or gradle project (e.g. https://github.com/gradle/gradle-build-scan-quickstart)
> 4. Update project (e.g shift+alt+U) for watching current state of java extensions and build file patterns
> 5. Install new VSCode Extension that will add java extension or build file patterns from contributes in package.json (You can find vsix file in attached zip document)
> 6. Click on the pop-up Reload button
> 7. Update project (e.g shift+alt+U) for watching current state of java extensions and build file patterns
>    [bazel-ls-vscode-0.0.1.zip](https://github.com/redhat-developer/vscode-java/files/6274135/bazel-ls-vscode-0.0.1.zip)

Thanks for the detailed steps, will have a try!

jdneo:
It works fine, when I installed the private build, the dialog pops up to ask for reloading.

Just one more question, I used this project for testing https://github.com/bmuschko/bazel-examples/tree/master/java/junit4-test, after reload, the classpath seems still not correct(red squiggles in the code), any steps I missed?

LeonovecSergey:
> It works fine, when I installed the private build, the dialog pops up to ask for reloading.
> 
> Just one more question, I used this project for testing https://github.com/bmuschko/bazel-examples/tree/master/java/junit4-test, after reload, the classpath seems still not correct(red squiggles in the code), any steps I missed?

It is normal that the plugin did not resolve the classpath. There're two reasons for this:
1. Our plugin supports a specific project structure only
    e.g. https://github.com/salesforce/bazel-ls-demo-project/tree/master/small
   And also you need to add some settings to the setting.json file: 
   "java.import.bazel.enabled": true,
   "java.import.bazel.src.path": "/java/src",
    "java.import.bazel.test.path": "/java/test"
2. The plugin used new features of eclipse.jdt.ls, which are not currently available in the release version. 
    https://github.com/eclipse/eclipse.jdt.ls/pull/1694

src/plugin.ts
```diff
@@ -26,24 +27,26 @@ export function collectJavaExtensions(extensions: readonly vscode.Extension<any>
...
+			const contributesSection = extension.packageJSON['contributes'];
+			if (contributesSection) {
+				const buildFilePatterns = contributesSection['javaBuildFilePatterns'];
+				if (Array.isArray(buildFilePatterns) && buildFilePatterns.length) {
+					result.push(...buildFilePatterns);
+				}
 			}
 		}
 	}
+	buildFilePatterns = result.slice();
```
jdneo, line 43: There is a potential bug here. the cache `buildFilePatterns` is updated during the collect function, then passed as the `oldContributedPart` to `isContributedPartUpdated()`, that will cause `isContributedPartUpdated()` always return false.

Same to `existingExtensions`

jdneo:
The problem fixed. It can pop-up the dialog for reloading when contribution changes.

Though I still cannot import the sample project, there is an error saying:

```
[Info  - 10:59:47 AM] Apr 17, 2021, 10:59:47 AM Tried to link a non-existant file [/Users/neo/Downloads/bazel-ls-demo-project-master/small/BUILD] for project [Bazel Workspace (small)]
[Error - 10:59:47 AM] Apr 17, 2021, 10:59:47 AM Initialization failed 
java.lang.NullPointerException
java.lang.RuntimeException: java.lang.NullPointerException
	at com.salesforce.b2eclipse.config.BazelEclipseProjectFactory.precomputeBazelAspectsForWorkspace(BazelEclipseProjectFactory.java:601)
	at com.salesforce.b2eclipse.config.BazelEclipseProjectFactory.importWorkspace(BazelEclipseProjectFactory.java:175)
	at com.salesforce.b2eclipse.managers.BazelProjectImporter.importToWorkspace(BazelProjectImporter.java:81)
	at org.eclipse.jdt.ls.core.internal.managers.ProjectsManager.importProjects(ProjectsManager.java:113)
	at org.eclipse.jdt.ls.core.internal.managers.ProjectsManager.initializeProjects(ProjectsManager.java:102)
	at org.eclipse.jdt.ls.core.internal.handlers.InitHandler$1.runInWorkspace(InitHandler.java:187)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:42)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
Caused by: java.lang.NullPointerException
	at com.salesforce.b2eclipse.config.BazelEclipseProjectFactory.precomputeBazelAspectsForWorkspace(BazelEclipseProjectFactory.java:597)
	... 7 more
```

Anyway, this is another problem beyond the scope of this PR. Overall the PR LGTM now.

+ @testforstephen @rgrunber @fbricon for awareness

jdneo (approved):

fbricon:
Thanks @LeonovecSergey!

PR #1847, opened 2021-03-24
Add the java.settings.url property

### LINKED ISSUE #1657: Writing Java in VS CODE shows way too much warnings!
Java Code in vs code shows way too much warnings. I mean these are really unnecessary warnings ! Even IDEs like IntelliJ don't show these unnecessary warnings. I want to disable such warnings, but not all the warnings. Is there a way to do so, or can you please do something about it ?

##### **Environment**
- Operating System: Windows 10 Pro
- JDK version:
java 14.0.2 2020-07-14
Java(TM) SE Runtime Environment (build 14.0.2+12-46)
Java HotSpot(TM) 64-Bit Server VM (build 14.0.2+12-46, mixed mode, sharing)
- Visual Studio Code version:  1.50.1
- Java extension version: 0.69.0

##### **Steps To Reproduce**
Download the folder below and open in vs code

[assignmentSix.zip](https://github.com/redhat-developer/vscode-java/files/5386879/assignmentSix.zip)

Here we see that under MonthException we see a squiggle line with warning: 

a

Warning Given: 
_The serializable class MonthException does not declare a static final serialVersionUID field of type longJava(536871008)_

Even IDE like IntelliJ **dont** show these useless warnings...
b

So is there any way to disable these warnings **_**(NOT ALL WARNINGS)**_** I mean can I only select what warnings I want to see and what I don't?

snjeza:
Fixes #1657 
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/1579

### DIFF
--- a/package.json
+++ b/package.json
@@ -754,6 +754,12 @@
           "default": false,
           "description": "Enable/disable lazy loading the content in type hierarchy. Lazy loading could save a lot of loading time but every type should be expanded manually to load its content.",
           "scope": "window"
+        },
+        "java.settings.url": {
+          "type": "string",
+          "markdownDescription": "Specifies the url or file path to the workspace Java settings. See [Setting Global Preferences](https://github.com/redhat-developer/vscode-java/wiki/Settings-Global-Preferences)",
+          "default": null,
+          "scope": "window"
         }
       }
     },
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ New in 0.77.0:
   - `afterCursor`: Insert the generated code after the member where the cursor is located.
   - `beforeCursor`: Insert the generated code before the member where the cursor is located.
   - `lastMember`: Insert the generated code as the last member of the target type.
+* `java.settings.url` : Specifies the url or file path to the workspace Java settings. See [Setting Global Preferences](https://github.com/redhat-developer/vscode-java/wiki/Settings-Global-Preferences)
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -736,6 +736,12 @@
           "default": false,
           "description": "Enable/disable lazy loading the content in type hierarchy. Lazy loading could save a lot of loading time but every type should be expanded manually to load its content.",
           "scope": "window"
+        },
+        "java.settings.url": {
+          "type": "string",
+          "markdownDescription": "Specifies the url or file path to the workspace Java settings.",
```
fbricon, line 742: please provide more infos about that settings file. How do you create it, what's in it? You might want to link to a wiki page with some details about it

snjeza: Fixed. See [Settings Global Preferences](https://github.com/redhat-developer/vscode-java/wiki/Settings-Global-Preferences)

rgrunber (approved):
Looks good to me.

package.json
rgrunber: Is it worth mentioning on that wiki page that for those unable to use Eclipse itself, the preference keys, descriptions and values are listed at https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java ? At least until we find a way to convert that page to something we can display somewhere here.

PR #1858, opened 2021-03-31
Fix vulnerabilities with 'npm audit fix'

rgrunber:
- y18n 4.0.0 -> 4.0.1
- Update webpack & webpack-cli to update ssri

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ before_install:
 - if [ $TRAVIS_OS_NAME == "linux" ]; then
       export CXX="g++-4.9" CC="gcc-4.9";
   fi
+- nvm install 12.22;
 - npm install -g typescript
 
 env:
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,22 +6,10 @@
         {
             "type": "npm",
             "script": "watch",
-            "problemMatcher": {
-                "owner": "typescript",
-                "pattern":[
-                    {
-                        "regexp": "\\[tsl\\] ERROR",
-                        "file": 1,
-                        "location": 2,
-                        "message": 3
-                    }
-                ],
-                "background": {
-                    "activeOnStart": true,
-                    "beginsPattern": "Compilation \\w+ starting…",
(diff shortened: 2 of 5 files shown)


.vscode/tasks.json
```diff
@@ -6,22 +6,10 @@
...
-                ],
-                "background": {
-                    "activeOnStart": true,
-                    "beginsPattern": "Compilation \\w+ starting…",
-                    "endsPattern": "Compilation\\s+finished"
-                }
-            },
+            "problemMatcher": [
+                "$ts-webpack-watch",
+                "$tslint-webpack-watch"
```
fbricon, line 11: this requires the eamodio.tsl-problem-matcher extension. See https://github.com/redhat-developer/vscode-commons/blob/master/.vscode/extensions.json#L6

rgrunber:
I think the issue is that webpack-cli update introduced the colorette 1.2.2 requirements, which apparently [doesn't work on node v13](https://github.com/jorgebucaran/colorette/issues/48), which is what our macos builder is using.

.vscode/tasks.json
rgrunber: Done.

.travis.yml
```diff
@@ -19,6 +19,9 @@ before_install:
 - if [ $TRAVIS_OS_NAME == "linux" ]; then
       export CXX="g++-4.9" CC="gcc-4.9";
   fi
+- if [ $TRAVIS_OS_NAME == "osx" ]; then
+      nvm install 12.12;
```
fbricon, line 23: I'd use the same node version consistently across platform builds

rgrunber: I noticed [vscode-microprofile](https://github.com/redhat-developer/vscode-microprofile/blob/master/.travis.yml#L25) and [vscode-quarkus](https://github.com/redhat-developer/vscode-quarkus/blob/master/.travis.yml#L25) were doing the same thing for osx, so went with that. I'll eliminate the conditional here and likely those other places as well.

rgrunber: Done.

PR #1868, opened 2021-04-07
Set minimum for threshold settings

jdneo:
Fix https://github.com/microsoft/vscode-java-pack/issues/590

### DIFF
--- a/package.json
+++ b/package.json
@@ -641,12 +641,15 @@
           "type": "integer",
           "description": "Specifies the number of imports added before a star-import declaration is used.",
           "default": 99,
-          "scope": "window"
+          "scope": "window",
+          "minimum": 1
         },
         "java.sources.organizeImports.staticStarThreshold": {
           "type": "integer",
           "description": "Specifies the number of static imports added before a star-import declaration is used.",
-          "default": 99
+          "default": 99,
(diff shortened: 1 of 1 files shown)


jdneo:
Do we need to fallback the value to default (99) when it's non-positive at JDT.LS? Since some users may manually edit settings in `settings.json`

package.json
```diff
@@ -641,12 +641,15 @@
           "type": "integer",
           "description": "Specifies the number of imports added before a star-import declaration is used.",
           "default": 99,
-          "scope": "window"
+          "scope": "window",
+          "minimum": 1
```
testforstephen, line 645: https://github.com/eclipse/eclipse.jdt.ls/blob/b647c553ed817c1c70b278f1d3a7e047d27e54b8/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/Preferences.java#L1544-L1549

We have guarded the negative case in server side. If zero is not legal, i think we need update the condition check in server side as well.

jdneo: PR raised: https://github.com/eclipse/eclipse.jdt.ls/pull/1715

fbricon (approved):

PR #1885, opened 2021-04-16
Upversion to 0.78.0

rgrunber:


rgrunber:
No idea why it's consistently failing on Mac only. The last commit only touched the changelog after all (and passed), so either way we need to upversion to address the issue.

PR #1888, opened 2021-04-16
Remove 'container environments' term.

rgrunber:
- The default memory options passed to the JVM help for desktop
  environments as well, so let's make the description less confusing


fbricon:
thx @rgrunber

PR #1894, opened 2021-04-23
Update language client to 7.0.0 to adopt LSP 3.16 features

testforstephen:

> See [language client 7.0](https://github.com/microsoft/vscode-languageserver-node#3160-protocol-600-json-rpc-700-client-and-700-server) changelog.
> split code into common, node and browser to allow using the LSP client and server npm modules in a Web browser via webpack. This is a breaking change and might lead to compile / runtime errors if not adopted.

LanguageClient@7.0.0 has a breaking change that splits the "LanguageClient" model to "vscode-languageclient/node" subfolder. So we have to update all imports of LanguageClient.


src/protocol.ts
```diff
@@ -95,39 +95,39 @@ export interface EventNotification {
 }
 
 export namespace StatusNotification {
-	export const type = new NotificationType<StatusReport, void >('language/status');
+	export const type = new NotificationType<StatusReport>('language/status');
```
testforstephen, line 98:
```typescript
export declare class NotificationType<P> extends AbstractMessageSignature {
    private _parameterStructures;
    /**
     * Clients must not use this property. It is here to ensure correct typing.
     */
    readonly _: [P, _EM] | undefined;
    constructor(method: string, _parameterStructures?: ParameterStructures);
    get parameterStructures(): ParameterStructures;
}
```

See the definition of NotificationType, it only accepts one Generic Type Parameter. No idea why it doesn't report compilation errors before.

src/protocol.ts
```diff
@@ -95,39 +95,39 @@ export interface EventNotification {
...
 }
 
 export namespace ProgressReportNotification {
-	export const type = new NotificationType<ProgressReport, void >('language/progressReport');
+	export const type = new NotificationType<ProgressReport>('language/progressReport');
 }
 
 export namespace ClassFileContentsRequest {
-    export const type = new RequestType<TextDocumentIdentifier, string, void, void> ('java/classFileContents');
+    export const type = new RequestType<TextDocumentIdentifier, string, void> ('java/classFileContents');
```
testforstephen, line 106:
```typescript
export declare class RequestType<P, R, E> extends AbstractMessageSignature {
    private _parameterStructures;
    /**
     * Clients must not use this property. It is here to ensure correct typing.
     */
    readonly _: [P, R, E, _EM] | undefined;
    constructor(method: string, _parameterStructures?: ParameterStructures);
    get parameterStructures(): ParameterStructures;
}
```

Same as NotificationType, RequestType just accepts 3 generic parameter types.

fbricon:
We have already seen issues with vscode-xml on Theia, because this new dependency requires vscode 1.52, while Theia is only at 1.50 for now

testforstephen:
https://github.com/redhat-developer/vscode-java/blob/f911f00b87253447d34aefe2e4c1bb99863c800c/package.json#L13-L15

Since the latest vscode-java already requires a minimum 1.53 engine, i assume this won't break Theia.

fbricon:
I prefer certainty over assumptions ;-) @snjeza / @rgrunber please check in Theia/CRW

rgrunber:
I do get a failure to install 0.77.0 ( which includes bcd7dca94b3f19cd10adf847a380a09d4825716c ) on VSCode 1.50.1 and the marketplace will install a lower version. Is there something particular about Type Hierarchy that forced 1.53 ?

testforstephen:
> I do get a failure to install 0.77.0 ( which includes bcd7dca#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R14 ) on VSCode 1.50.1 and the marketplace will install a lower version.

If you're in VS Code 1.50.1, VS Code Extensions page only lists those compatible vscode-java versions for you. That means vscode-java@0.77 (which requires a minimum 1.53 engine) is not visible to those VS Code 1.50.1 users.

> Update 2 : Theia will begin to fail with this patch introduced. Looks like the engines/vscode 1.53 requirement doesn't fail Theia. It's the language client moving to 7.0. It's why vscode-java still works, and why vscode-xml started failing.

In Theia Extensions page, what's the highest vscode-java version you can see? Is it 0.77 or a lower compatible version?

rgrunber:
> If you're in VS Code 1.50.1, VS Code Extensions page only lists those compatible vscode-java versions for you. That means vscode-java@0.77 (which requires a minimum 1.53 engine) is not visible to those VS Code 1.50.1 users.

Yup. VS Code 1.50.1 will not permit the extension to be installed, and installs an earlier version if attempted from the marketplace (eg. 0.76.0 instead of 0.77.0).
 
> In Theia Extensions page, what's the highest vscode-java version you can see? Is it 0.77 or a lower compatible version?

If installing from the extension page for vscode-java, it will install 0.76.0. So on Theia the marketplace correctly installs the compatible version but if done by "Install from vsix" it doesn't complain, which might be a separate bug.

testforstephen:
> If installing from the extension page for vscode-java, it will install 0.76.0. So on Theia the marketplace also correctly installs the compatible version.

Thank you for the verification. The marketplace is the only official channel for installing Java extension. If it ensures the compatibility check, then it's safe for us to use the current mechanism to adopt new capabilities.

fbricon:
we'll release  this in 0.79.0, right after 0.78.0, so we can use Java 16 support in Theia

fbricon:
Now requires #1914

rgrunber:
This looks fine to me. I've tested out the code action / resolve code action calls and seems to be working as expected. Additional functionality also seems undisturbed with these changes.

PR #1908, opened 2021-04-28
Continue to support VS Code 1.50

rgrunber:
- Undo our update of the VS Code engine requirement to ^1.53.2, moving
  it down to ^1.50.0
- Allow those basing their language client off of VS Code 1.50.0 to
  continue to use this version and not be forced to install an earlier
  version
- Guard the 'api' field (which holds the reference view) if using a
  version that does not support it

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "preview": true,
   "enableProposedApi": false,
   "engines": {
-    "vscode": "^1.53.2"
+    "vscode": "^1.50.0"
   },
   "repository": {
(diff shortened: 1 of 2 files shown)


fbricon:
need to commit package-lock.json as well

rgrunber:
> need to commit package-lock.json as well

I think because the version in package-lock.json for "@types/vscode" is 1.53.0, ^1.50.0 will still accept it. If I do force an update to 1.55.0.

PR #1909, opened 2021-04-28
Fix TreeItem API usage for VS Code 1.50.0

rgrunber:
- The TreeItemLabel constructor for vscode.TreeItem was introduced in
  1.52.0, so use the string constructor for compatibility. The only
  downside is tree items that are the basis of the search, will not be
  highlighted
- Update package-lock.json

The commit in vscode that introduced the functionality (in 1.52.0) that we were using : https://github.com/microsoft/vscode/commit/0286c4f793a508837176e1ee3b76a0431d22bb6a

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -95,9 +95,9 @@
       "dev": true
     },
     "@types/vscode": {
-      "version": "1.53.0",
-      "resolved": "https://registry.npmjs.org/@types/vscode/-/vscode-1.53.0.tgz",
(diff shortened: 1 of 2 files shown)


rgrunber:
tree-item-highlight

The highlighting on "AssembleBundleP2RepositoryMojo" (the basis for the search) would no longer be there until we adopted the 1.52.0 API.

rgrunber:
NVM, this change is pointless because on 1.50.0 we have no type hierarchy to worry about. Sorry for the noise.

I think the important part is that the vscode engine is set to 1.50.0 so the marketplace should work now with a 1.50.0 client.

PR #1943, opened 2021-05-12
disable extension in untrusted workspace

Eskibear:
See #1926

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,11 @@
     "onCommand:_java.templateVariables"
   ],
   "main": "./dist/extension",
+  "capabilities": {
+    "untrustedWorkspaces": {
+      "supported": false
+    }
+  },
   "contributes": {
     "javaBuildFilePatterns": [
       "^pom.xml$",


rgrunber:
From https://github.com/redhat-developer/vscode-java/issues/1926#issuecomment-839764371, I'm wondering if it's at all possible to declare `limited` support and define a `restrictedConfigurations` option that when not present (workspace not trusted), puts the server in lightweight mode (syntax server only). This is basically working around the fact that we're using an older version of VS Code so can't yet use the API for workspace trust.

Eskibear:
I agree that we only start lightweight server when the workspace is untrusted. It's possible and I can create a PR to implement the basic logic. But the overall UX should be polished for different use cases.

Cases to cover:
- on activation, determine which server(standard/lightweight) to start. Can tell from flag `workspace.isTrusted`.
- untrusted -> trusted, switch to standard server if necessary. We can listen to event `workspace.onDidGrantWorkspaceTrust`.
- trusted -> untrusted, VS Code will reload extension host automatically. Possibly no extra effort from our side.

`restrictedConfigurations` defines settings that we don't read from **untrusted workspace**. Values of these settings are simply ignored if it's defined in workspace level and the workspace is untrusted. For example, previously a repository can include workspace level settings `java.home` pointing to a malicious executable file, and we prompt to ask for users trust by our own. Now we can simply add `java.home` into `restrictedConfigurations`.

List of `restrictedConfigurations`:
- `java.home`
- ...

please comment to add more settings you think should be restricted, i.e. the scope should be workspace level for flexibility, but with potential security risk like injection.

rgrunber:
Sounds good. I would even by fine with just supporting the trust based on startup, but if we can also support switching the trust at runtime, that would be great. I think `java.home` is the main one. I don't know if we want to restrict all of `java.jdt.ls.vmargs` just for the javaagent flag. Currently we handle this by asking the user to explicitly allow this.

Looks like trust will be [enabled by default](https://github.com/microsoft/vscode/commit/dd61b26f33a91f700d2e2835d8a009bd251c76b8) in the next release.

Eskibear:
Based on my experiment, it's important to also listen to `workspace.onDidGrantWorkspaceTrust`, and start standard server when trust is granted. 

Below is a demo where I simply check `workspace.isTrusted` before starting standard server. The modal dialog asking for trust doesn't block java extension from activation, thus for a newly opened folder, java extension would always be activated when workspace is untrusted, and you don't have chance to start standard server automatically.

https://user-images.githubusercontent.com/2351748/121144056-3ec62f00-c870-11eb-825d-ebd628fb92d3.mp4

Eskibear:
The simplest way is:
* if workspace is untrusted, we force `serverMode` to be lightweight.
* listen to events, switch server mode when trust is granted.

I can create a corresponding PR.

PR #1975, opened 2021-06-08
launch lightweight LS in untrusted workspaces

Eskibear:
alternative to #1943 , providing limited language support for untrusted workspaces.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1266,7 +1266,7 @@
     "decompress-tar": {
       "version": "4.1.1",
       "resolved": "https://registry.npmjs.org/decompress-tar/-/decompress-tar-4.1.1.tgz",
-      "integrity": "sha1-cYy9P8sWIJcW5womuE57pFkuWvE=",
+      "integrity": "sha512-JdJMaCrGpB5fESVyxwpCx4Jdj2AagLmv3y58Qy4GE6HMVjWz1FeVQk1Ct4Kye7PftcdOo/7U7UKzYBJgqnGeUQ==",
       "dev": true,
       "requires": {
         "file-type": "^5.2.0",
@@ -1285,7 +1285,7 @@
     "decompress-tarbz2": {
       "version": "4.1.1",
       "resolved": "https://registry.npmjs.org/decompress-tarbz2/-/decompress-tarbz2-4.1.1.tgz",
-      "integrity": "sha1-MIKluIDqQEOBY0nzeLVsUWvho5s=",
+      "integrity": "sha512-s88xLzf1r81ICXLAVQVzaN6ZmX4A6U4z2nMbOwobxkLoIIfjVMBg7TeguTUXkKeXni795B6y5rnvDw7rxhAq9A==",
       "dev": true,
       "requires": {
         "decompress-tar": "^4.1.0",
@@ -1298,7 +1298,7 @@
         "file-type": {
           "version": "6.2.0",
           "resolved": "https://registry.npmjs.org/file-type/-/file-type-6.2.0.tgz",
-          "integrity": "sha1-5QzXXTVv/tTjBtxPW89Sp5kDqRk=",
+          "integrity": "sha512-YPcTBDV+2Tm0VqjybVd32MHdlEGAtuxS3VAYsumFokDSMG+ROT5wawGlnHDoz7bfMcMDt9hxuXvXwoKUx2fkOg==",
           "dev": true
         }
       }
@@ -1306,7 +1306,7 @@
     "decompress-targz": {
       "version": "4.1.1",
       "resolved": "https://registry.npmjs.org/decompress-targz/-/decompress-targz-4.1.1.tgz",
-      "integrity": "sha1-wJvDXE0R894J8tLaU+neI+fOHu4=",
+      "integrity": "sha512-4z81Znfr6chWnRDNfFNqLwPvm4db3WuZkqV+UgXQzSngG3CEKdBkw5jrv3axjjL96glyiiKjsxJG3X6WBZwX3w==",
       "dev": true,
       "requires": {
         "decompress-tar": "^4.1.1",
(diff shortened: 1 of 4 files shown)


Eskibear:
A demo:

https://user-images.githubusercontent.com/2351748/121150331-0295cd00-c876-11eb-8ad5-3c3a25b85c5c.mp4

package.json
```diff
@@ -10,8 +10,16 @@
...
+    "untrustedWorkspaces": {
+      "supported": "limited",
+      "restrictedConfigurations": [
+        "java.home"
+      ]
+    }
+  },
   "engines": {
-    "vscode": "^1.53.2"
+    "vscode": "^1.56.0"
```
rgrunber, line 22: @fbricon , Currently Theia doesn't even support 1.52, so we'd be leaving them further behind, but it seems there's no way to use the API without this.

package-lock.json
```diff
@@ -1460,7 +1460,7 @@
         },
         "readable-stream": {
           "version": "1.1.14",
-          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-1.1.14.tgz",
+          "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-1.1.14.tgz",
```
rgrunber, line 1463: Is it possible to keep these as https ?

rgrunber (commented):
It seems to be working as expected. Starts in lightweight mode if not trusted. If the workspace is in hybrid and trusted, then the switch is made to standard.

I noticed that if I switch from trusted to untrusted while in standard mode then the language server itself is immediately killed but no way to bring it back up after that in any mode. I guess this is just an unsupported part o the API currently ?

package-lock.json
Eskibear: It's modified by `npm install`, maybe I should upgrade my local npm...  I'll change it back.

Eskibear:
> I noticed that if I switch from trusted to untrusted while in standard mode then the language server itself is immediately killed but no way to bring it back up after that in any mode.

Leaving `java.server.mode` to be its default value "hybrid", everything works fine. After I clicked "Don't trust", it reloaded the whole extension host, vscode-java extension was re-activated, and lightweight server started. 

I can reproduce it after changing `java.server.mode` to "Standard". The culprit is how we switch server mode, instead of the trusted workspace API. The event of granting trust is received, and it returns early when switching to standard mode.

https://github.com/redhat-developer/vscode-java/blob/9e208a272e5dfb922b067560c24b604d7b737d33/src/extension.ts#L282-L285

Seems like value of `api.serverMode` is directly read from setting instead of reflecting current mode. @jdneo Can you help to take a look, see if I was using it correctly.

Eskibear:
>  ERROR  @types/vscode ^1.56.0 greater than engines.vscode ^1.53.2. Consider upgrade engines.vscode or use an older @types/vscode version

It failed during `vsce package`. seems we cannot simply leave engine version unchanged.

rgrunber:
> > ERROR  @types/vscode ^1.56.0 greater than engines.vscode ^1.53.2. Consider upgrade engines.vscode or use an older @types/vscode version
> 
> It failed during `vsce package`. seems we cannot simply leave engine version unchanged.

Would keeping @types/vscode at 1.53.0 and using `(workspace as any)` work for those cases where we call isTrusted/onDidGrantWorkspaceTrust until we can properly adopt the API ?

Eskibear:
That should work, let me update and squash the commits.

rgrunber (commented):
Overall, this change seems to be working well when tested with a locally built vsix.

However, there is some strange behaviour when testing in an Extension Development Host depending on whether the main window (with the vscode-java sources) has vscode-java enabled or not.

- If the vscode-java extension is left enabled in the main workspace, then the extension development host will start the installed vscode-java (eg. ~/.vscode/extensions/redhat.java-0.79.2/..) language server instead of the one from the workspace (eg. ~/git/vscode-java/server/..).

The above is probably a bit confusing so I'll try to attach a demo of these cases. I'm fine with merging but it would be good to know if these are issues in how the extension development host handles trust.

rgrunber:
The 2 initial java-0.79.2 java processes are from the pain VSCode window. There are 2 because the syntax server doesn't automatically shut down in the released version. This will be fixed in the next release though, and on the development branch.

trust_starts_wrong_jdt_ls

package.json
```diff
@@ -10,6 +10,14 @@
   "bugs": "https://github.com/redhat-developer/vscode-java/issues",
   "preview": true,
   "enableProposedApi": false,
+  "capabilities": {
+    "untrustedWorkspaces": {
+      "supported": "limited",
+      "restrictedConfigurations": [
+        "java.home"
```
rgrunber, line 17: It might make sense to also add `java.jdt.ls.vmargs` (though we do protect javaagent with a warning popup when we detect it in the workspace).

Eskibear:
> ... but it would be good to know if these are issues in how the extension development host handles trust.

I can reproduce it with vscode-java 0.79.2 enabled. Then I modified `~/.vscode/redhat.java-0.79.2/package.json` to add `untrustedWorkspaces` section, it works again. I think it's caused by the inconsistency declaration of untrusted workspace support between the enabled java extension (0.79.2) and the one we are developing.

A reasonable **guess** is: when you click "Don't Trust", vscode reloads extension development host (but not completely the same with reloading window). When determine whether to activate each extension. When processing then enabled redhat.java-0.79.2, vscode decides not to activate it because there's no untrustedWorkspaces support declared. And for some reason it fails to activate/recognize/load the java extension we are debugging now. **The evidence is, in your screenshot, there's no rocket icon in status bar after you click "Don't Trust".** And you can execute "Developer: Show running extensions" command, and you'll find java extension is not activated at this moment. Then you click "Trust" again, vscode activates redhat.java-0.79.2 that has already been recognized, and `jps -l` logs make sense.

package.json
Eskibear: added in commit 563af52.

rgrunber:
I think you're right. The released extension lacking the support for workspace trust combined with slight difference in how reloading is done probably causes this. I guess developers just have to be extra careful when testing in an extension development host (or test with a vsix).

I'll have a final look for `jdt.ls.vmargs` but I think the above behaviour is ok.

rgrunber (approved):
I think this is ready to be merged now.

PR #1995, opened 2021-06-23
Recollect Java bundles when granting workspace trust

### LINKED ISSUE #1994: 3rd party Java bundles are not loaded when clicking trust workspace button to switch server mode
Reproduce steps:
- Install java-0.80.0 nightly build from https://download.jboss.org/jbosstools/jdt.ls/staging/ and Java debugger extension.
- If it exists, remove the 0.80.0 global cache from VS Code global storage cache. In macOS, it's `/Users/username/Library/Application Support/Code/User/globalStorage/redhat.java/0.80.0`.
- Open a new Java project which has not been opened in VS Code before, it's untrusted by default.
- Click "Trust" button to trust current workspace and Java language server will be switched to standard mode.
- Once Java LS is started and tries to run a Java file, Java debugger failed because of the exception about "workspace command not found".

image

testforstephen:

Fixes #1994

In untrusted workspaces, those extensions that do not declare support for workspace trust are not visible to the `extensions.all API`. Since the variable `clientOptions` is preconfigured when activating the Java extension, it becomes outdated after we trust the workspace. The list of Java bundles needs to be recollected before switching to standard mode.

It seems that `extensions.all` still does not include the new available Java extensions during the callback `onDidGrantWorkspaceTrust`. This looks like a VS code bug. A workaround would be to add a timeout to allow Java extensions to get the latest available Java bundles.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -7,7 +7,7 @@ import * as fse from 'fs-extra';
 import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, Selection, Position, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration, env, UIKind } from 'vscode';
 import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClientOptions, RevealOutputChannelOn, ErrorHandler, Message, ErrorAction, CloseAction, DidChangeConfigurationNotification, CancellationToken } from 'vscode-languageclient';
 import { LanguageClient } from 'vscode-languageclient/node';
-import { collectJavaExtensions } from './plugin';
+import { collectJavaExtensions, isContributedPartUpdated } from './plugin';
 import { prepareExecutable } from './javaServerStarter';
 import * as requirements from './requirements';
 import { initialize as initializeRecommendation } from './recommendation';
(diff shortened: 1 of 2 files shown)


testforstephen:
// @Eskibear

rgrunber:
If there's some way to use extensions.onDidChange() and something like https://github.com/redhat-developer/vscode-microprofile/blob/master/src/languageServer/plugin.ts#L45-L68 to avoid the timeout that would be great. Otherwise I'd just confirm that the timeout can be made no smaller.

testforstephen:
Update: `extensions.onDidChange()` is not called when granting workspace trust.

We also created an issue https://github.com/microsoft/vscode/issues/127055 to ask `extension.all` API to returns latest result when onDidGrantWorkspaceTrust. Before that's addressed, timeout could be a temporary workaround for us now.

rgrunber:
If the `extensions.onDidChange()` route produces unwanted behaviour in some cases, I think both @fbricon and I agreed that the current solution of waiting 1s could work with a change :

- Poll every ~100ms (timeout after 1s) and check whether contributing extensions have changed (eg. https://github.com/redhat-developer/vscode-java/blob/master/src/plugin.ts#L60 ).

This way at least we never wait the maximum amount of time. I know we've discussed that we can't tell whether the change came from the workspace trust action, but how likely is it for a user to be able to perform an installation action near the "trust action" and have this still fail ?

testforstephen:
I have changed to polling every 100ms and timeout after 1s. I tried it in two machines, both them could get the extension changes at 400ms ~500ms.
**Machine 1 - Windows 10**
Intel(R) Xeon(R) CPU E5-1620 v4 @ 3.50GHz   3.50 GHz
32.0 GB

**Machine 2 - macOS**
MacBook Pro (13-inch, 2016, Four Thunderbolt 3 Ports)
Processor: 2.9 GHz Dual-Core Intel Core i5
Memory: 8 GB 2133 MHz LPDDR3

> how likely is it for a user to be able to perform an installation action near the "trust action" and have this still fail?

Once the standard server is initialized, we have extra logic to cover the case of installing new extension.
https://github.com/redhat-developer/vscode-java/blob/7ac2be5fba4b436e9a49f7b205cbace275bf26f4/src/standardLanguageClient.ts#L402-L406

If you're talking about the time window between the action of starting standard server and registering `extensions.onDidChange` callback, yes, it is possible to miss new extensions if they are coincidentally installed at that time. This should also exist in current stable version 0.79.2, we might need a new PR to systematically solve that case.

rgrunber (approved):
Works for me.

PR #2003, opened 2021-06-25
replace sync icon to refresh

shuyaqian:
Signef-off-by: shuyaqian 717749594@qq.com

The original icon is too small, and the rotation track is incorrect.
before:

after:

### DIFF
--- a/src/serverStatusBarProvider.ts
+++ b/src/serverStatusBarProvider.ts
@@ -57,7 +57,7 @@ class ServerStatusBarProvider implements Disposable {
 
 enum StatusIcon {
 	LightWeight = "$(rocket)",
-	Busy = "$(sync~spin)",
+	Busy = "$(refresh~spin)",
 	Ready = "$(thumbsup)",
 	Error = "$(thumbsdown)"
 }


fbricon:
Thanks @shuyaqian!

rgrunber:
What OS did you test this ?

I ask because this is the behaviour I am seeing before and after this change on Linux with VS Code 1.57.1 :

Before :
before_spin

After :
after_spin

It may be a bit trickier to fix this nicely for everyone.

testforstephen:
The new icon doesn't work in my macOS and windows 10 either. It just shows a static refresh icon and doesn't rotate. I think we need to revert this commit.

See https://github.com/microsoft/vscode/search?q=sync~spin&type=code, VS Code also uses `$(sync~spin)` in its own code to display a loading status. @shuyaqian, If there is a problem with its display in your OS, i suggest you to open a bug in VS Code to fix this.

fbricon:
weird, it worked for me on macos, that's why I merged it.

rgrunber:
Ok, so I'll revert this, but we'll need to know under what conditions the icon renders like the first comment (rotation slightly off centre). The only place I have seen this that I can recall, is in Theia, so perhaps we're dealing with an older version of VSCode.

PR #2004, opened 2021-06-25
Revert .vscodeignore changes

joaomoreno:
More info: https://github.com/microsoft/vscode-vsce/issues/588

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -12,8 +12,9 @@ vsc-extension-quickstart.md
 undefined/**
 CONTRIBUTING.md
 .vscode-test/**
-**/*.vsix
-**/*.tar.gz
+**/**.vsix
+**/**.tar.gz
+dist/**/*.map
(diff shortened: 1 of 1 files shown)


rgrunber:
Thanks for the heads up.

According to https://github.com/redhat-developer/vscode-java/pull/1983#pullrequestreview-684686936 , our changes should be able to support both, which would be better, but I'll verify.

joaomoreno:
@rgrunber Yeah at a glance I believe your solution does support both implementations.

rgrunber:
I've confirmed vsce ls output is identical for local builds before and after.
Our builders also switched over to 1.95 today and comparing the packaging :

DONE  Packaged: java-0.80.0-255.vsix (113 files, 45.83MB)
DONE  Packaged: java-0.80.0-256.vsix (113 files, 45.83MB)

I would probably close this out for now.

Issue #2014, opened 2021-07-07
Retire the asking importing dialog


jdneo:
Long time ago we introduced a feature that asking user whether to import the projects when opening the workspace for the first time:

![image](https://user-images.githubusercontent.com/6193897/124707685-06853100-deae-11eb-932e-6778a3a5d9cb.png)

According to the user's selection, we will determine which server mode will be used.

Now, since VS Code has introduced the workspace trust feature as a platform solution, I think it's a good time to retire this dialog. By doing so, we can also reduce the popping-ups and get a better UX.

Any thoughts? @rgrunber @fbricon @testforstephen 

_// I'm willing to propose a PR if we agree on this._

rgrunber:
Looks like we're in favour of changing the default option to "always", but would like to keep the settings itself around for now so users still have the option to change it.

If we determine at a later point that this isn't a deal breaker for anyone, then we can remove it completely.

jdneo:
Sounds good.

PR #2016, opened 2021-07-08
Change the default value of the setting 'java.project.importOnFirstTimeStartup' to 'automatic'

jdneo:
related with #2014


rgrunber (approved):
LGTM. I'm fine with incrementally adding to the CHANGELOG, though can be optional. We'll see if it's any trickier to sort them out at release time.

Issue #2021, opened 2021-07-12
Tune the import UX.


jdneo:
Importing experience is the very important part for our getting start experience. If we go through our issues list, we can find there are several parts that we can improve:

### 1. Provide more transparency
- How to get the import status?
  - We have a spinning icon but that is not straightforward/easy to find.
- What's the result after import finished?

Proposal: #2022 

https://user-images.githubusercontent.com/6193897/126450147-266d9659-6b13-4c59-aaba-37c8c8fc8973.mp4

### 2. Provide more controls
- 2.1 Let user control the active build tool when there are conflicts

Proposal: #2037

https://user-images.githubusercontent.com/6193897/126450128-0b9c968a-e18c-4947-8d53-059bbeb49977.mp4

- 2.2 Let user control the project configuration files that will be used to discover projects to import. (WIP)

### Eliminate the side-effect
- Importing project will generate `.project`, `.classpath`, etc... into users workspace. This is discussed in #618, so won't be discussed here.

Eskibear:
> And maybe an action to switch between the build tools once the project got imported?
_from @martinlippert_

Eskibear:
IMO ideally we need a “imported project manager”, let users explicit choose which projects to import and how to import them. It’s very useful when I open a giant repo containing hundreds of projects and I only want to import a few projects I’m interested in. 

E.g. https://github.com/Azure/azure-sdk-for-java/tree/main/sdk
I'm working on only some of the SDKs but all are in a mono repo.

PR #2022, opened 2021-07-13
Explicitly show import progress

jdneo:
This is the first part of #2021.

In this PR:

- we show the job status more explicitly during the import stage. (Like Eclipse or IntelliJ, at the right-bottom corner, users can see what jobs are running)

- If user click the hyper link in the notification, the build status terminal will pop up and the notification get dismissed.
- After importing jobs are finished. A dialog shows up to let the user know.

https://user-images.githubusercontent.com/6193897/125410372-6a987f80-e3ef-11eb-9e40-6d3d344cd1ee.mp4

### DIFF
--- a/src/serverTaskPresenter.ts
+++ b/src/serverTaskPresenter.ts
@@ -122,8 +122,16 @@ export class ActivationProgressNotification {
 	private lastJobId: string;
 
 	public showProgress() {
-		const isProgressReportEnabled: boolean = getJavaConfiguration().get('progressReports.enabled');
-		const title = isProgressReportEnabled ? `[Importing Projects](command:${Commands.SHOW_SERVER_TASK_STATUS})` : "Importing Projects";
+		const showBuildStatusEnabled = getJavaConfiguration().get("showBuildStatusOnStart.enabled");
+		if (typeof showBuildStatusEnabled === "string" || showBuildStatusEnabled instanceof String) {
+			if (showBuildStatusEnabled !== "notification") {
+				return;
+			}
+		} else if (!showBuildStatusEnabled) {
+			return;
+		}
+		const isProgressReportEnabled: boolean = getJavaConfiguration().get("progressReports.enabled");
+		const title = isProgressReportEnabled ? `[Importing Projects](command:${Commands.SHOW_SERVER_TASK_STATUS})` : "Importing Projects...";
 		window.withProgress({
 			location: ProgressLocation.Notification,
 			title,
--- a/src/serverTaskPresenter.ts
+++ b/src/serverTaskPresenter.ts
@@ -131,7 +131,7 @@ export class ActivationProgressNotification {
 			return;
 		}
 		const isProgressReportEnabled: boolean = getJavaConfiguration().get("progressReports.enabled");
-		const title = isProgressReportEnabled ? `[Importing Projects](command:${Commands.SHOW_SERVER_TASK_STATUS})` : "Importing Projects...";
+		const title = isProgressReportEnabled ? "Importing Projects" : "Importing Projects...";
 		window.withProgress({
 			location: ProgressLocation.Notification,
 			title,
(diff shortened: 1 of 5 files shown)


fbricon:
need a preference to not show the progress

jdneo:
> need a preference to not show the progress

Ok

@rgrunber @snjeza @testforstephen @akaroml @Eskibear @CsCherrYY Since this change related to UX, feel free to let me know if you have any comments

fbricon:
nitpick : can you please fix the typo in the commit message? (s/improt/import)

jdneo:
> nitpick : can you please fix the typo in the commit message? (s/improt/import)

Done, will have another commit to add the preference

src/serverTaskPresenter.ts
```diff
@@ -112,3 +113,63 @@ class ServerTaskTerminal implements Pseudoterminal {
...
+			cancellable: false,
+		}, (progress: Progress<{ message?: string; increment?: number }>) => {
+			return new Promise<void>((resolve) => {
+				this.disposables.push(
+					serverTasks.onDidUpdateServerTask(tasks => {
+						let taskToShow: ProgressReport | undefined = tasks.find((task) => {
+							if (task.id === this.lastJobId) {
+								return task;
+							}
+						});
```
rgrunber, line 136: Is this ever triggered ? this.lastJobId is never set so this always results in an undefined taskToShow here.

Not sure if keeping track of the last incomplete task is required if the tasks.find(..) logic directly below here will always find the next incomplete one.

rgrunber (commented):
Works well for me. Just a minor thing to address so far.

src/serverTaskPresenter.ts
jdneo: My bad. I should cache the `lastJobId` once the a uncomplete task is found.

jdneo:
About adding a setting to control the appearance of the progress notification.

I'm thinking that adding a setting called: `java.showJobStatusOnStart`, which has 3 options:

- `notification`: show the job status via notification no start
- `terminal`: Show the job status via terminal on start
- `off`: Do not show the job status on start

Meanwhile, deprecated `java.showBuildStatusOnStart.enabled`. (Still working but consider to remove it after several iterations). What do you think?

jdneo:
PR update, finally I use the `anyOf` mechanism in `package.json` to reuse the setting `java.showBuildStatusOnStart.enabled` meanwhile provides backwards compatibility. User can set it to:

- `notification`: Show the build status via progress notification.
- `terminal`: Show the build status via terminal.
- `off`: Do not show any build status.

  > For backward compatibility, this setting also accepts boolean value, where `true` has the same meaning as `notification` and `false` has the same meaning as `off`.

Besides, the default value changes to `notification`

@fbricon @rgrunber

rgrunber (commented):
Change works well. Just one thing I noticed that may be annoying (given how much space the progress bar takes), although the same behaviour occurs with the status bar provider.

If you kill the syntax/standard server while it's importing/building the project (or it crashes), the progress notification will remain indefinitely on the screen. The restarted language server ultimately finishes the importing, but the progress bar remains. I think the same thing can happen for the status bar provider. I wonder if there's a way to deal with this.

jdneo:
I can imagine that some users might not like these kind of pop-ups.

Or, maybe we can still use `off` as the default value and then in the terminal, always show some action links somewhere in it. One of the action is open the related setting to let user set the preference.

Anyway, we can discuss in our next sync meeting.

jdneo:
PR updated to simplify the text.

Meanwhile, I created a [new issue](https://github.com/redhat-developer/vscode-java/issues/2036) to track the item that dealing with the situation when server crashes

rgrunber (approved):
LGTM. Feel free to squash and merge when ready.

jdneo:
Hi @rgrunber, Jinbo(@testforstephen) is the only person who has the permission to merge PRs. He is now on vacation until next week.

If you plan to make a release in this week, maybe we need your help to merge. :)

PR #2031, opened 2021-07-15
Move pull request verification job to GitHub Actions

rgrunber:
- Migrate away from Travis CI and to GitHub Actions

GitHub verification builds haven't run on PRs for about a month since the changes to travis-ci, so I've created a basic PR to migrate to GitHub actions.

It's working pretty well, except this seems to fail on the `macos-latest` environment :

```
1) Public APIs - Standard
       onDidClasspathUpdate should work:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/vscode-java/vscode-java/out/test/standard-mode-suite/publicApi.test.js)
  	at listOnTimeout (internal/timers.js:554:17)
  	at processTimers (internal/timers.js:497:7)
```

Any ideas here ?


testforstephen:
Add @jdneo as a reviewer as well, since he has experience migrating to GitHub Actions.

jdneo:
@rgrunber Is this case always failed after re-run?

rgrunber:
> @rgrunber Is this case always failed after re-run?

Yes, it seems to fail consistently. The only thing that stood out to me was that various tasks do take much longer to complete on the macos instance, so I did try doubling the timeout but that seemed to have no effect. Only other difference I noticed is I believe on travis-ci we had macos 10.13, whereas here we have 10.15.

jdneo:
Interesting... I didn't observe any suspicious part in the yaml file. I'll use my Mac to test it locally first when I go to the office next week.

jdneo:
I cannot repro this on my local machine. @rgrunber How about adding the following task at the end of the MacOS job definition to see if we can find any clues from the server log?

```
- name: Print language server Log if job failed
      if: ${{ failure() }}
      run: find $HOME/Library/Application\ Support/Code/User/workspaceStorage/*/redhat.java/jdt_ws/.metadata/.log -print -exec cat '{}' \;;
```

champpp96 (approved):

jdneo:
There are a lot of `!MESSAGE >> workspace/didChangeConfiguration` before timeout, that's quite strange. 

On my local machine,

It will show sth like when executing the classpath update test
```
Starting Maven update for salut
...
Updated salut in xxx ms
```

jdneo:
Maybe just skip that one test when the platform is darwin and mark it as a debt to resolve in the future if we have time?

Issue #1008, opened 2019-07-26
Allow user to pick Gradle or Maven as build tool if it is ambiguous


dsyer:
I noticed that if you open VSCode in a directory that has both Maven and Gradle build config it always starts up the Gradle daemon. I know it's a bit unusual, but we do it with all the Guides in Spring, for instance: https://github.com/spring-guides/gs-rest-service/tree/master/complete. I would like an option to pick either Maven or Gradle in such cases.

snjeza:
You can try the following property:
```
"java.import.gradle.enabled": false,
```

dsyer:
That helps thanks. Possibly related: the Gradle daemon never seems to die when the workspace exits. The LS process doe go away but Gradle is still running (and multiple instances build up over time).

dsyer:
Actually, it doesn't help. The Gradle daemon is still there whatever I try to do:

```
$ jps
22481 org.eclipse.equinox.launcher_1.5.500.v20190715-1310.jar
22645 GradleDaemon
22745 BootLanguagServerBootApp
22766 Jps
```

snjeza:
Can you kill the Gradle daemon?

dsyer:
I did. but it always comes back when I open this folder in code.

fbricon:
Try to clean the workspace directory

dsyer:
I vaguely remember you asked me to try that once before. Where is it? It’s something buried under a uuid-names directory somewhere in my home dir?

fbricon:
- Open the command palette (F1)
- select Java: Clean the Java language server workspace
- click Restart and delete

fbricon:
Deleting the eclipse .project files to remove the Gradle builder might be useful too

dsyer:
That did it. Thanks. I had to kill the LS and the Gradle daemon again as well, but once all that was done it started to behave.

hanct:
> You can try the following property:
> 
> ```
> "java.import.gradle.enabled": false,
> ```

Is this setting enabled by default? Cos the guide does not state the default

dsyer:
The default is `true` (so both Gradle and Maven try to build the project if both are available).

Issue #600, opened 2018-07-30
New setting entry to let user choose project type


akaroml:
We have projects that have multiple project info inside. Like this one:
https://github.com/spring-guides/gs-spring-boot

Both Gradle and Maven project info exists. In this case, Gradle wins by default.

The project configurations come from the source repo, but when users actually work on it, it's better to have an option to let users choose their preferred project type.

The new setting could look like:
preferredProjectType: "Gradle" // or "Maven"

And the setting could be global or per project.

fbricon:
You can use `java.import.gradle.enabled:false` or `java.import.maven.enabled:false`, for similar results

Eskibear:
@fbricon I tried the both settings, they work in most cases. For some reason I such a repository, in the `Utils` folder, there are Maven based utils, and the main project is Gradle based. I want to import all Maven and Gradle projects, but only Gradle projects are recognized.

Eskibear:
I think the root cause is [Here](https://github.com/eclipse/eclipse.jdt.ls/blob/b49b83a2c30928578fd3085b0a36d918b3bcf20e/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/managers/ProjectsManager.java#L323-L333), When gradle importer applies, it stops checking for other importers. So whenever there is a `build.gradle` file in the workspace, there would be no maven projects recognized I think. 


```java
	private IProjectImporter getImporter(File rootFolder, IProgressMonitor monitor) throws OperationCanceledException, CoreException {
		Collection<IProjectImporter> importers = importers();
		SubMonitor subMonitor = SubMonitor.convert(monitor, importers.size());
		for (IProjectImporter importer : importers) {
			importer.initialize(rootFolder);
			if (importer.applies(subMonitor.split(1))) {
				return importer;
			}
		}
		return null;
	}
```

PR #2037, opened 2021-07-22
Can select build tool when there is a conflict

jdneo:
fix https://github.com/redhat-developer/vscode-java/issues/600
fix https://github.com/redhat-developer/vscode-java/issues/1008

The notification will only be pop up when the following conditions meet:
- Both Maven importer and Gradle importer are enabled
- At least one folder contains both `pom.xml` & `*.gradle`, meanwhile, it does not contain `.project` (not being imported before)

So this will only affect a new workspace which has build tool conflicts

--------

During implementation, I also find some buggy things in the Java Language Server. For example,

1. User select Gradle when first opening a fresh project
2. User update the setting to enable Maven importer and disable Gradle Importer.
3. Clean workspace cache and reload

The project is still Gradle, this is because the .classpath/.project file is still in user's project and Buildship listens some broadcast events(when refreshLocal is called) for these files. Buildship will then schedule a sync job by itself.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -4,7 +4,7 @@ import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
 import * as fse from 'fs-extra';
-import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, Selection, Position, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration, env, UIKind } from 'vscode';
+import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration, env, UIKind } from 'vscode';
 import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClientOptions, RevealOutputChannelOn, ErrorHandler, Message, ErrorAction, CloseAction, DidChangeConfigurationNotification, CancellationToken } from 'vscode-languageclient';
 import { LanguageClient } from 'vscode-languageclient/node';
 import { collectJavaExtensions, isContributedPartUpdated } from './plugin';
@@ -14,7 +14,7 @@ import { initialize as initializeRecommendation } from './recommendation';
 import { Commands } from './commands';
 import { ExtensionAPI, ClientStatus } from './extension.api';
 import { getJavaConfiguration, deleteDirectory, getBuildFilePatterns, getInclusionPatternsFromNegatedExclusion, convertToGlob, getExclusionBlob } from './utils';
-import { onConfigurationChange, getJavaServerMode, ServerMode } from './settings';
+import { onConfigurationChange, getJavaServerMode, ServerMode, ACTIVE_BUILD_TOOL_STATE } from './settings';
 import { logger, initializeLogFile } from './log';
 import glob = require('glob');
 import { SyntaxLanguageClient } from './syntaxLanguageClient';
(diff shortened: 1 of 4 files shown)


snjeza (approved):

snjeza:
> The project is still Gradle, this is because the .classpath/.project file is still in user's project and Buildship listens some broadcast events(when refreshLocal is called) for these files. Buildship will then schedule a sync job by itself.

We could remove .classpath, .project, .settings before importing a gradle or maven project if the project doesn't exist in the workspace.

jdneo:
> We could remove .classpath, .project, .settings before importing a gradle or maven project if the project doesn't exist in the workspace.

@snjeza Yes I also thought of that. One question about this: I'm not sure how many users have their own customized .project/.classpath files in their project. (I'm afraid that removing those file will surprise them.)

snjeza:
>  Yes I also thought of that. One question about this: I'm not sure how many users have their own customized .project/.classpath files in their project. (I'm afraid that removing those file will surprise them.)

The gradle/maven importer updates these files anyway.

jdneo:
> The gradle/maven importer updates these files anyway.

They just do a append-updated instead of a 'entire' update right?

But anyway, I agree that given the current situation, we have to remove those files to make sure the projects can be imported correctly. Either we help users to remove them or let the users remove them by themselves.

rgrunber:
The change seems to work for me. I'll have a closer look at the code itself and then I think we can merge. We should probably file a separate bug for the use-case you described.

src/extension.ts
```diff
@@ -409,6 +415,74 @@ async function workspaceContainsBuildFiles(): Promise<boolean> {
...
+	});
+}
+
+async function getBuildFilesInWorkspace(): Promise<Uri[]> {
+	const buildFiles: Uri[] = [];
+	const inclusionPatterns: string[] = getBuildFilePatterns();
+	inclusionPatterns.push("**/.project");
+	const inclusionPatternsFromNegatedExclusion: string[] = getInclusionPatternsFromNegatedExclusion();
+	if (inclusionPatterns.length > 0 && inclusionPatternsFromNegatedExclusion.length > 0) {
+		buildFiles.push(...await workspace.findFiles(convertToGlob(inclusionPatterns, inclusionPatternsFromNegatedExclusion), null /*force not use default exclusion*/));
```
rgrunber, line 466: Suppose this part of the code runs, because the user has some negated term they've placed in exclusions. The `null` ensures no exclusions are used. However, directly below we look for the normal inclusions (without the negated exclusions) and the normal exclusions. Haven't we already allowed content meant to be excluded to end up in buildFiles due to the initial findFiles (with no exclusions) ?

rgrunber (commented):
Overall, the change works and looks good to me. Just some small changes and question on a special case.

src/extension.ts
jdneo: Here we have two invocations of `workspace.findFiles()`.

- As you said, the first one is to search for the negated terms users placed in the setting `java.import.exclusions`. If the user explicitly set a negated term in `java.import.exclusions`, that means he wants to include that path, that's why a null (no exclusion when finding files) is used here.

- For the second one, since we've already dealt with the negated terms, here we can just do a normal search - set the exclusion pattern as what the user has set in `java.import.exclusions` (those positive terms).

The logic here is tricky, maybe I need to write some comments to explain it?

Not sure I explained it clearly, but let me know if you have more questions :)

rgrunber: My mistake. I assumed that `convertToGlob(inclusionPatterns, inclusionPatternsFromNegatedExclusion)` was simply combining each argument to form a union of patterns. As a result I thought setting exclusions to null would have unwanted patterns leaking in.

In fact the 1st argument is for file patterns included only under the base path of the 2nd, so there's no issue.

rgrunber:
Can you address the other minor issues and then I think we can merge.

src/extension.ts
jdneo: Yes correct, maybe the variable name is not good enough, I'll rename it.

jdneo:
Sure, PR updated.

PR #2042, opened 2021-07-26
Make commands wait for applyEdit 

fvclaus:
I stumbled over this trying to find out why `organizeImports` in combination with Prettier and `formatDocument` would always alternate its output. My configuration looked like this:
```
        "editor.codeActionsOnSave": [ 
            "source.organizeImports",
            "source.formatDocument",
        ]
```
import_order_alternating

The list is executed in order and the commands are awaited for completion. I must organize imports before saving to get rid of unused imports and because I cannot recreate Prettier's import order using vscode-java. (More details here: https://gitter.im/redhat-developer/vscode-java?at=60e84814f1274d5e503bf7d8)
Trying to recreate this issue with a simple extension, I found out that this occurs, when the command returns before all its asynchronous work is done.

After my changes, the behavior changes to the following:
import_order_stable

I applied the changes to all commands that use `applyWorkspaceEdit`. I used the code for about a week and did not find any (new) problems.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -718,10 +718,12 @@ async function addFormatter(extensionPath, formatterUrl, defaultFormatter, relat
 	});
 }
 
-export function applyWorkspaceEdit(obj, languageClient) {
+export function applyWorkspaceEdit(obj, languageClient): Thenable<boolean> {
 	const edit = languageClient.protocol2CodeConverter.asWorkspaceEdit(obj);
 	if (edit) {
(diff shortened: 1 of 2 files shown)


testforstephen (approved):
Since I don't know how to use Prettier Java extension in VS Code, I cannot reproduce the conflict issue between organizeImport and prettier.

But regarding this change, i don't see any bad effects of returning a promise in the applyEdit command. So i'd like to merge it.

PR #2056, opened 2021-08-10
Do not show the import notification when no projects available

jdneo:
@rgrunber @testforstephen  This fix some missed corner cases for #2022, which are:

1. Do not show the progress notification when there is no workspace folders available. (no root means no project)
   - For example, in a vs code window without workspace, create a new untitled file and set the language to Java
2. After import finishes, if no projects are imported, or only have the default project, do not pop-up the finishing dialog.


rgrunber (approved):
LGTM, and bypasses the import progress + notification when a simple file is opened and associated with the Java language.

PR #2065, opened 2021-08-17
Only show the import notification for the first time

jdneo:
Since our message is called `Projects are imported into workspace.` So it should only be shown for the first time.

To know whether the projects have been imported or not, I check the existence of `.metadata/.plugins`, which is used before: https://github.com/redhat-developer/vscode-java/blob/master/src/extension.ts#L334

@rgrunber Sorry for the late patch. Please consider if this should be contained in 0.81.0

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -390,7 +390,7 @@ async function startStandardServer(context: ExtensionContext, requirements: requ
 		apiManager.getApiInstance().serverMode = ServerMode.HYBRID;
 		apiManager.fireDidServerModeChange(ServerMode.HYBRID);
 	}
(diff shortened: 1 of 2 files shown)


testforstephen (approved):
LGTM.

Since this commit is related with other importing enhancements in 0.81.0, i prefer to carry it in the upcoming 0.81.0 release. @rgrunber what do you think?

rgrunber:
Sounds good. I'm assuming we need not update the changelogs as the relevant issues are mentioned.

testforstephen:
> Sounds good. I'm assuming we need not update the changelogs as the relevant issues are mentioned.

yes, no need to update changelog.

PR #2073, opened 2021-08-19
Add "shortmonth" date option and fix existing month/date variables

### LINKED ISSUE #1987: Add additional variables for java.template snippets
Like most teams, I have a required set of content that belongs in the header of my Java documents. One of the required items is a creation date. The format of the creation date is _shortmonth day, year_. So something like Jan 24, 2025. There doesn't seem to be a way using the predefined variables to add this to my template. Using `${date}` gets me `1/24/2025`. Using `${month} ${day}, ${year}` gets me `01 24, 2025`.

I would like it if `shortmonth` was added as a predefined variable. There are likely additional predefined variables that other developers would want.

##### Environment
- JDK version: JDK 11
- Visual Studio Code version: 1.57.1
- Java extension version: 0.79.2

##### Steps To Reproduce
1. Configure template in the settings file.
2. Create a new file.

##### Current Result
It is not possible to generate a date in the format required by my team.

##### Expected Result
It is possible to generate a date in the format required by my team.

##### Additional Informations
Documentation for existing predefined variables: https://github.com/redhat-developer/vscode-java/wiki/Predefined-Variables-for-Java-Template-Snippets

Example template configuration
```
    "java.templates.fileHeader": [
        "/**",
        " * Created: ${date}",
        " */"
    ],
```

rgrunber:
- Adjust 0-based month values & use day-of-month for date
- Fixes #1987


src/fileEventHandler.ts
```diff
@@ -67,11 +67,12 @@ async function handleNewJavaFiles(e: FileCreateEvent) {
...
             type_name: typeName,
             user: userInfo().username,
-            date: date.toLocaleDateString(),
+            date: date.toLocaleDateString(undefined, {month: "short", day: "2-digit", year: "numeric"}),
             time: date.toLocaleTimeString(),
             year: date.getFullYear(),
-            month: formatNumber(date.getMonth()),
-            day: formatNumber(date.getDay()),
+            month: formatNumber(date.getMonth() + 1),
+            shortmonth: date.toLocaleDateString(undefined, {month: "short"}),
```
testforstephen, line 74: Add the new variable `shortmonth` to the doc https://github.com/redhat-developer/vscode-java/blob/master/document/_java.templateVariables.md and the wiki https://github.com/redhat-developer/vscode-java/wiki/Predefined-Variables-for-Java-Template-Snippets

testforstephen: also `shortmonth` is missing in server side template, need add the variableResolver to eclipse.jdt.ls.

testforstephen (commented):
The new variables work as expected. Just some minor change to the doc and add the corresponding variable support to server side.

src/fileEventHandler.ts
rgrunber: PR updated. I'll create a separate PR on jdt.ls to add shortmonth. Realized it was probably a good idea to add it after I already merged :|

testforstephen (approved):

Issue #632, opened 2018-09-08
Support for Gradle .kts Files


dresch86:
The Language Server keeps saying my .java files aren't on the classpath. I am using Gradle 4.10. Initially I started with the standard Groovy based build files, but decided to give Kotlin (.kts) a go. The Groovy based Gradle build files were recognized by the language server, but it doesn't seem to be picking up any information from the build.gradle.kts files. Is this supported? Everything builds without error from command line when using .kts files.

##### Environment
- Operating System: Windows 10
- JDK version: 1.8.181
- Visual Studio Code version: 1.27.1
- Java extension version: 0.30.0

##### Steps To Reproduce
1. Create a gradle.build.kts file
2. Open it
3. Open a .java file

primesoftware:
I have some time to try and tackle this, however I am not clear on where the implementation details lay here. I see the references in extension.ts to the Gradle build files, but I don't see where these files get handled. How does vscode know where to look for external jars downloaded via Gradle? I can't seem to find any other references to Gradle that deal with where vscode-java looks for classpath.

fbricon:
Most of the work requires the jdt.ls server to support kotlin-based gradle files, that means buildship needs to [provide proper kotlin support](https://github.com/eclipse/buildship/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+kotlin) in the 1st place.
But I'm not sure we want to add gradle kotlin support to jdt.ls OOTB (for technical and legal reasons), so it might be enabled through an extension.
I suggest you 1st make sure gradle+kotlin is something viable in Eclipse. Then we can provide further pointers on how to try to make it work in jdt.ls+vscode.

fbricon:
FYI, on the jdt.ls side, gradle support starts with https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/managers/GradleBuildSupport.java

xenoterracide:
need, this plugin is useless without it, imho.

xenoterracide:
and honestly, all I need is for it to get the classpath appropriately, I couldn't care less about it parsing the kotlin file right now...

dsyer:
FWIW .kts build files seem to work OK in Eclipse now. Is it time to take another look at this maybe?

fbricon:
@akaroml @testforstephen @Eskibear @jdneo is this something you guys can look at?

[...]

PR #2077, opened 2021-08-20
Add *.gradle.kts support

snjeza:
Fixes https://github.com/redhat-developer/vscode-java/issues/632
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/1849

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,9 @@
     "onLanguage:java",
     "workspaceContains:pom.xml",
     "workspaceContains:build.gradle",
+    "workspaceContains:settings.gradle",
+    "workspaceContains:build.gradle.kts",
+    "workspaceContains:settings.gradle.kts",
     "workspaceContains:.classpath",
     "onCommand:_java.templateVariables"
   ],
@@ -46,7 +49,7 @@
   "contributes": {
     "javaBuildFilePatterns": [
       "^pom.xml$",
-      ".*\\.gradle$"
+      ".*\\.gradle(.kts)?$"
     ],
     "semanticTokenTypes": [
       {
@@ -942,7 +945,7 @@
       "explorer/context": [
         {
           "command": "java.projectConfiguration.update",
-          "when": "resourceFilename =~ /(.*\\.gradle)|(pom.xml)$/",
+          "when": "resourceFilename =~ /(.*\\.gradle.*)|(pom.xml)$/",
           "group": "1_javaactions"
         },
         {
@@ -964,7 +967,7 @@
         },
         {
           "command": "java.projectConfiguration.update",
-          "when": "resourceFilename =~ /(.*\\.gradle)|(pom.xml)$/",
+          "when": "resourceFilename =~ /(.*\\.gradle.*)|(pom.xml)$/",
           "group": "1_javaactions"
         },
         {
--- a/package.json
+++ b/package.json
@@ -48,8 +48,8 @@
   "main": "./dist/extension",
   "contributes": {
     "javaBuildFilePatterns": [
-      "^pom.xml$",
-      ".*\\.gradle(.kts)?$"
+      "^pom\\.xml$",
+      ".*\\.gradle(\\.kts)?$"
     ],
     "semanticTokenTypes": [
       {
@@ -945,7 +945,7 @@
       "explorer/context": [
         {
           "command": "java.projectConfiguration.update",
-          "when": "resourceFilename =~ /(.*\\.gradle.*)|(pom.xml)$/",
+          "when": "resourceFilename =~ /(.*\\.gradle)|(.*\\.gradle\\.kts)|(pom\\.xml)$/",
           "group": "1_javaactions"
         },
         {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -39,14 +39,17 @@
...
+    "workspaceContains:settings.gradle.kts",
     "workspaceContains:.classpath",
     "onCommand:_java.templateVariables"
   ],
   "main": "./dist/extension",
   "contributes": {
     "javaBuildFilePatterns": [
       "^pom.xml$",
-      ".*\\.gradle$"
+      ".*\\.gradle(.kts)?$"
```
Eskibear, line 52: should use `\\.kts`?

package.json
```diff
@@ -942,7 +945,7 @@
       "explorer/context": [
         {
           "command": "java.projectConfiguration.update",
-          "when": "resourceFilename =~ /(.*\\.gradle)|(pom.xml)$/",
+          "when": "resourceFilename =~ /(.*\\.gradle.*)|(pom.xml)$/",
```
Eskibear, line 948: better to be consistent with what in `javaBuildFilePatterns`, otherwise the command will also be registered on something like a `gradle.txt` file if exists

jdneo:
There are some other places that might need to check .kts:

https://github.com/redhat-developer/vscode-java/blob/076455e9f9d4392886e5fb858a88d7b1a2353bb8/src/utils.ts#L58
https://github.com/redhat-developer/vscode-java/blob/c3f18c34949a9cbfdb3604fa1e3a1a43e9eebe35/src/extension.ts#L453
  - I think here the parameter `fileName` of `getDirectoriesByBuildFile()` might need to change from `string` to `string[]`, and we check if the file ends with `gradle` or `gradle.kts`?
https://github.com/redhat-developer/vscode-java/blob/a8fef5089e61f4840672931d72b50b1664418098/test/standard-mode-suite/extension.test.ts#L124

package.json
snjeza: Fixed.

package.json
snjeza: Fixed.

snjeza:
@jdneo I have updated the PR.

package.json
```diff
@@ -964,7 +967,7 @@
         },
         {
           "command": "java.projectConfiguration.update",
-          "when": "resourceFilename =~ /(.*\\.gradle)|(pom.xml)$/",
+          "when": "resourceFilename =~ /(.*\\.gradle.*)|(pom.xml)$/",
```
testforstephen, line 970: Suggest the  "when clause" is consistent as other places.

`"when": "resourceFilename =~ /(.*\\.gradle)|(.*\\.gradle.kts)|(pom.xml)$/",`

fbricon: shouldn't .kts and .xml be escaped too? `/(.*\\.gradle)|(.*\\.gradle\\.kts)|(pom\\.xml)$/`

liuchong (approved):

package.json
snjeza: > 
> 
> Suggest the "when clause" is consistent as other places.
> 
> `"when": "resourceFilename =~ /(.*\\.gradle)|(.*\\.gradle.kts)|(pom.xml)$/",`

https://github.com/redhat-developer/vscode-java/pull/2077/commits/b16c65bfec64dd44c1b8f6cea8595ae7bce95621#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R948

snjeza: Fixed.

snjeza:
@fbricon @testforstephen I have updated the PR.

package.json
```diff
@@ -942,7 +945,7 @@
       "explorer/context": [
         {
           "command": "java.projectConfiguration.update",
-          "when": "resourceFilename =~ /(.*\\.gradle)|(pom.xml)$/",
+          "when": "resourceFilename =~ /(.*\\.gradle)|(.*\\.gradle.kts)|(pom\\.xml)$/",
```
testforstephen, line 948: `/(.*\\.gradle)|(.*\\.gradle\\.kts)|(pom\\.xml)$/`

package.json
```diff
@@ -964,7 +967,7 @@
         },
         {
           "command": "java.projectConfiguration.update",
-          "when": "resourceFilename =~ /(.*\\.gradle)|(pom.xml)$/",
+          "when": "resourceFilename =~ /(.*\\.gradle.*)|(.*\\.gradle.kts)|(pom\\.xml)$/",
```
testforstephen, line 970: `/(.*\\.gradle)|(.*\\.gradle\\.kts)|(pom\\.xml)$/`

package.json
snjeza: Fixed.

package.json
snjeza: Fixed.

snjeza:
@testforstephen I have updated the PR.

jdneo (approved):

PR #2083, opened 2021-08-24
Adopt new createStatusBarItem API

### LINKED ISSUE #1982: Consider to adopt new `createStatusBarItem` API for `id` and `name` properties
👋  Hi there, Ben here, from the VS Code team.

We just released VSCode 1.57 that landed a new stable API for status bar items to provide `id` and `name` properties: https://code.visualstudio.com/updates/v1_57#_status-bar-item-id-and-name-properties

I noticed that your extension is a consumer of the `createStatusBarItem` API. 

Here is an example how using the new API will improve the user experience around managing status bar items:

Without `id` / `name` right clicking on the status bar item from your extension will show a generic label that is just the name of the extension:

statusbar

As you can see from the example above, the GH Pull Request extension did not yet adopt this new API and as such, the menu only shows a generic entry (as opposed to the other entries that are more specific).

Another motivation to adopt this new API is that it will enable a user to selectively hide/show status bar entries from your extension in case you register more than one status bar item.

An adoption of the new API involves:
* configuring `"engines": { "vscode": "^1.57.0" }` in `package.json`
* where `createStatusBarItem` is called, add a unique `id: string` property as first parameter
* in the returned `StatusBarItem` set `name` to something that describes the intent of the status bar entry (if you provide multiple status bar entries, make sure the name is different for each entry)

Thanks for the support and the great work! ❤️

jdneo:
resolve #1982 

Here is the appearance in standard mode
WeChat Screenshot_20210824163918

Here is the appearance in Lightweight mode
WeChat Screenshot_20210824163941


Eskibear (approved):

Eskibear:
LGTM.  

Current it's showing extension name instead of a meaningful name of the item, as shown below: 
image

jdneo:
The new [`createStatusBarItem()`](https://github.com/microsoft/vscode/blob/2d6b50d6962ed3a507feeb4c9d26157a923b2ba9/src/vs/vscode.d.ts#L9089) is an overload method, which I guess we need to use the [version](https://github.com/microsoft/vscode/blob/2d6b50d6962ed3a507feeb4c9d26157a923b2ba9/src/vs/vscode.d.ts#L11) field to check which version of the API should be used. Do you think this is accpetable? (we may need to introduce a new npm module: [semver](https://www.npmjs.com/package/semver) to do the version comparing)

rgrunber:
> The new [`createStatusBarItem()`](https://github.com/microsoft/vscode/blob/2d6b50d6962ed3a507feeb4c9d26157a923b2ba9/src/vs/vscode.d.ts#L9089) is an overload method, which I guess we need to use the [version](https://github.com/microsoft/vscode/blob/2d6b50d6962ed3a507feeb4c9d26157a923b2ba9/src/vs/vscode.d.ts#L11) field to check which version of the API should be used. Do you think this is accpetable? (we may need to introduce a new npm module: [semver](https://www.npmjs.com/package/semver) to do the version comparing)

I guess if this is the only way to approach this, we'd have to do it. It seems to already be a transitive dependency of other requirements we have.

jdneo:
Cool table for API tracking, thanks!

jdneo:
@rgrunber Thanks it works for me as well. I just updated the PR.

PR #2085, opened 2021-08-25
Gracefully handle OutOfMemory errors

### LINKED ISSUE #1959: Explicitly report OutOfMemory error to user
When opening some large Java projects (e.g. https://github.com/Azure/azure-sdk-for-java) in VS Code, the language server may fail silently due to OutOfMemory error. The setting `"java.jdt.ls.vmargs"` of Java extension specifies a default `"-Xmx1G"` memory during starting language server, that's not enough for some large project. We need some way to explicitly tell user to increase the memory args in this case.

rgrunber:
- Detect and report java.lang.OutOfMemory errors
- Fixes #1959
- Use -XX:+ExitOnOutOfMemoryError to ensure Java language server exits
  when an OutOfMemory error occurs (rather than staying up)
- Use -XX:+HeapDumpOnOutOfMemoryError & -XX:HeapDumpPath to generate a
  heap dump whose existence can notify the client that an OutOfMemory
  error has occured
- Once OutOfMemory error is detected, clean up the heap dumps

Taken from https://github.com/redhat-developer/vscode-xml/pull/548

A few limitations of this approach :
- We rely particularly on setting -XX:HeapDumpPath (we set it to global storage path of the extension) in order to detect if OOM occured. If the user overrides this value, we can't really be sure.
- The status bar icon continues to spin even after we handle the issue, but I believe this has always been a problem when a crash occurs
- We emit the error message for the 2 servers (syntax, standard) though I haven't noticed any obvious downside to this

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -4,11 +4,11 @@ import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
 import * as fse from 'fs-extra';
-import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration, env, UIKind, Selection, CodeActionContext, Diagnostic } from 'vscode';
+import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration, env, UIKind, CodeActionContext, Diagnostic } from 'vscode';
 import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClientOptions, RevealOutputChannelOn, ErrorHandler, Message, ErrorAction, CloseAction, DidChangeConfigurationNotification, CancellationToken, CodeActionRequest, CodeActionParams, Command } from 'vscode-languageclient';
 import { LanguageClient } from 'vscode-languageclient/node';
 import { collectJavaExtensions, isContributedPartUpdated } from './plugin';
-import { prepareExecutable } from './javaServerStarter';
+import { HEAP_DUMP_LOCATION, prepareExecutable } from './javaServerStarter';
 import * as requirements from './requirements';
 import { initialize as initializeRecommendation } from './recommendation';
 import { Commands } from './commands';
@@ -26,6 +26,7 @@ import { SnippetCompletionProvider } from './snippetCompletionProvider';
 import { runtimeStatusBarProvider } from './runtimeStatusBarProvider';
 import { serverStatusBarProvider } from './serverStatusBarProvider';
 import { markdownPreviewProvider } from "./markdownPreviewProvider";
+import * as chokidar from 'chokidar';
 
 const syntaxClient: SyntaxLanguageClient = new SyntaxLanguageClient();
 const standardClient: StandardLanguageClient = new StandardLanguageClient();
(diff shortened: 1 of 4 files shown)


testforstephen:
> Use -XX:+ExitOnOutOfMemoryError to ensure Java language server exits
when an OutOfMemory error occurs (rather than staying up)

I'm thinking if it's too strict to shutdown JVM if any OOM error is detected.

With the Java debugger (which is on top of JLS), there are cases where it may cause an OOM, but it just fails the operation but not breaking the language server. See the issue https://github.com/microsoft/vscode-java-debug/issues/1044

src/extension.ts
```diff
@@ -219,12 +277,13 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			};
 
 			apiManager.initialize(requirements, serverMode);
+			const globalStoragePath = context.globalStorageUri.fsPath;
```
testforstephen, line 280: Saving the dump file to `globalStorage` is not a good idea, since it's shared by all VS Code windows. Using `storagePath` per workspace would be better.

src/extension.ts
```diff
@@ -219,12 +277,13 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
 
 			if (requireSyntaxServer) {
 				if (process.env['SYNTAXLS_CLIENT_PORT']) {
-					syntaxClient.initialize(requirements, clientOptions, resolve);
+					syntaxClient.initialize(requirements, clientOptions, resolve, globalStoragePath);
 				} else {
-					syntaxClient.initialize(requirements, clientOptions, resolve, prepareExecutable(requirements, syntaxServerWorkspacePath, getJavaConfig(requirements.java_home), context, true));
+					syntaxClient.initialize(requirements, clientOptions, resolve, globalStoragePath, prepareExecutable(requirements, syntaxServerWorkspacePath, getJavaConfig(requirements.java_home), context, true));
```
testforstephen, line 286: We can handle OOM errors for the standard server only. Because the standard server has more memory consumption than the syntax server, if the syntax server fails due to OOM, then the standard server will definitely have the same error.

src/extension.ts
```diff
@@ -78,6 +93,46 @@ export class ClientErrorHandler implements ErrorHandler {
...
+		await fse.remove(heapProfile);
+	}
+}
+
+/**
+ * Shows a message about the server crashing due to an out of memory issue
+ */
+async function showOOMMessage(): Promise<void> {
+	const DOCS = 'More info...';
+	const result = await window.showErrorMessage('The Java Language Server crashed due to an Out Of Memory Error, and will not be restarted. ', //
```
testforstephen, line 113: I would suggest providing a "**Configure**" button here, which might be more intuitive than "**More info...** ".

Click on the "Configure" button and then open the doc and settings view with the filter "java.jdt.ls.vmargs" lined up.

testforstephen (commented):
> - Avoid using -XX:+ExitOnOutOfMemoryError. We can have a file watcher on the client side that simply detects if a heapdump was created, and if so, shows the same notifications, and if an OOM did happen, it doesn't restart if the LS ever happens to close.
> - Only show the OOM notification once per extension lifecycle. Any more might just be annoying.

Agree, using file watcher to monitor the creation of the heapdump file is a sensible way to detect OOM error. Once an OOM error is detected, pop up a notification and provide suggestion to mitigate it.

src/extension.ts
testforstephen, line 113: An alternative message could be:

"An OutOfMemoryError has occurred, some language features may not work due to limited memory. Please consider increasing available heap."

testforstephen: A new feedback is we could even simplify the workflow more, for example, just make the button to be "**Increase A new feedback is that we can even simplify the workflow more. For example, make the button to be "**Increase and Restart**". Once the user clicks it, then automatically double `-Xmx` or crank up 512M each time and restart the language server.

src/extension.ts
rgrunber: Done. Re-using `storagePath` now.

src/extension.ts
rgrunber: I've added a "Configure" button but I'll expand it further to a button that just increases `-Xmx` and restarts.

rgrunber: Done.

src/extension.ts
rgrunber: I've removed the logic that required anything of the syntax server. We now just listen for the heap dump files using chokidar, and show the error message if detected.

src/extension.ts
```diff
@@ -925,3 +965,16 @@ function isPrefix(parentPath: string, childPath: string): boolean {
...
+
+function registerOutOfMemoryDetection(storagePath: string) {
+	const heapDumpFolder = getHeapDumpFolderFromSettings() || storagePath;
+	chokidar.watch(`${heapDumpFolder}/java_*.hprof`).on('add', path => {
+		// Only clean heap dumps that are generated in the default location.
+		// The default location is the extension global storage
+		// This means that if users change the folder where the heap dumps are placed,
+		// then they will be able to read the heap dumps,
+		// since they aren't immediately deleted.
+		fse.remove(path);
```
testforstephen, line 977: The comments declare that we only auto delete the heap dumps of the default location. If that's what you want, then you need to check if the path is from `storagePath` before the line `fse.remove(path)`.

src/extension.ts
```diff
@@ -4,11 +4,11 @@ import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
 import * as fse from 'fs-extra';
-import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration, env, UIKind, Selection, CodeActionContext, Diagnostic } from 'vscode';
+import { workspace, extensions, ExtensionContext, window, commands, ViewColumn, Uri, languages, IndentAction, InputBoxOptions, EventEmitter, OutputChannel, TextDocument, RelativePattern, ConfigurationTarget, WorkspaceConfiguration, env, UIKind, Selection, CodeActionContext, Diagnostic, FileSystemWatcher } from 'vscode';
```
testforstephen, line 7: unused import `FileSystemWatcher`.

testforstephen (commented):
It works well with the latest commit. Only a minor comment about auto deletion of dump file, once it is addressed, feel free to merge it.

src/extension.ts
rgrunber: Done. Also noticed Selection so removed that too.

src/extension.ts
rgrunber: Done. I compare `storagePath` to `heapDumpFolder` to figure out if we delete or not.

rgrunber:
Just a note about an additional change I made. I removed the notification that warns we won't delete heap dump files when the user sets their own `-XX:HeapDumpPath=...`. The reason for that warning was because the file system api could only tell us of the existence of files on some path. If a user pointed to a path that already had heap dump files, we would incorrectly detect them and assume a crash had occurred. With chokidar, we can actually be notified on file creation, so this problem goes away. We do need `ignoreInitial: true` to avoid the initial addition.

All that's left is whether we want to inform the user that we won't delete their heap dumps but will delete ours. I think this is the right thing to do anyways, and since we should probably avoid too many notifications, I've removed it. I've also added a check to ensure we don't add any of the heap dump options to the syntax server.

testforstephen (approved):

PR #2087, opened 2021-08-27
Permit usage of javaagent (no prompt) if resource not in workspace

### LINKED ISSUE #1965: Define in the user preferences a list of file hashes to be trusted when jar files are added to the classpath of jdt.ls

##### Environment
- Operating System:
- JDK version:
- Visual Studio Code version: Che-theia
- Java extension version:

##### Steps To Reproduce
When we add a jar file in the classpath like 
```
      java.jdt.ls.vmargs: '-javaagent:/lombok.jar'
```
a Security Warning is popping up and this is an issue in Che the option is not persisted in the user preferences (but in the global state which is gone when starting a new workspace).

image

For security reasons, we would like to provide a list of jar files that can be trusted in the preferences.
Would be something similar to https://github.com/redhat-developer/vscode-xml/blob/985705ebd603bd08ee8ca149484d3ea48841007a/docs/Preferences.md#trusted-binary-hashes

rgrunber:
- Fixes #1965

### DIFF
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -6,7 +6,7 @@ import * as os from 'os';
 import * as fs from 'fs';
 import { StreamInfo, Executable, ExecutableOptions } from 'vscode-languageclient/node';
 import { RequirementsData } from './requirements';
-import { getJavaEncoding, IS_WORKSPACE_VMARGS_ALLOWED, getKey, getJavaagentFlag } from './settings';
+import { getJavaEncoding, IS_WORKSPACE_VMARGS_ALLOWED, getKey, getJavaagentFlag, isInWorkspaceFolder } from './settings';
 import { logger } from './log';
 import { getJavaConfiguration, deleteDirectory, ensureExists, getTimestamp } from './utils';
 import { workspace, ExtensionContext } from 'vscode';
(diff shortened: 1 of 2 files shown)


src/javaServerStarter.ts
```diff
@@ -69,7 +69,7 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
 		if (agentFlag !== null) {
 			const keyVmargs = getKey(IS_WORKSPACE_VMARGS_ALLOWED, context.storagePath, vmargsCheck);
 			const key = context.globalState.get(keyVmargs);
-			if (key !== true) {
+			if (key !== true || (workspace.workspaceFolders && isInWorkspaceFolder(agentFlag, workspace.workspaceFolders[0].uri.fsPath))) {
```
sunix, line 72: what if we have several folders in the workspace ? https://code.visualstudio.com/docs/editor/multi-root-workspaces

rgrunber: Done. One more thing I need to do, after discussing with others. We should also bypass the security warning if the user has already trusted the workspace. This is more for clients that have supported the feature.

If we get to the point where we can safely assume all clients support workspace trust, then there really should be no need for such a prompt.

rgrunber:
Ugh, silly addition on my part. At the point where we parse the agent flag, we've already determined the workspace is trusted (because we only start the standard JDT-LS for a trusted workspace) so there's no need for that workspace trust code at all.

rgrunber:
I'll just need to test this in Theia to confirm it'd work as expected.

sunix:
ok let me know if you need some help to test it.

rgrunber:
Ok, so I've verified this should work in Theia.

In Theia, workspace trust is undefined, so in that case we check whether the agentFlag value is in the workspace. If so, we show the security prompt, and otherwise we allow it. If a client supports workspace trust, the entire security prompt should be skipped.

PR #2106, opened 2021-09-11
Fix watch_server & dev_server tasks

### LINKED ISSUE #230: npm run watch fails with unresolved dependencies
Ubuntu 14.04

- Follow the instructions in https://github.com/redhat-developer/vscode-java/blob/master/CONTRIBUTING.md#building-eclipse--jdt-language-server
`npm run build-server` works but `npm run watch-server` fails with

Cannot resolve dependencies of product languageServer.product:
[ERROR]   eclipse-plugin artifact with ID "org.eclipse.jdt.launching.macosx" and version matching "0.0.0" was not found in the target platform
[ERROR]   eclipse-plugin artifact with ID "org.eclipse.core.net.linux.x86" and version matching "0.0.0" was not found in the target platform
[ERROR]   eclipse-plugin artifact with ID "org.eclipse.core.net.win32.x86" and version matching "0.0.0" was not found in the target platform
[ERROR]   eclipse-plugin artifact with ID "org.eclipse.core.net.win32.x86_64" and version matching "0.0.0" was not found in the target platform
[ERROR]   eclipse-plugin artifact with ID "org.eclipse.equinox.security.macosx" and version matching "0.0.0" was not found in the target platform
[ERROR]   eclipse-plugin artifact with ID "org.eclipse.equinox.security.win32.x86" and version matching "0.0.0" was not found in the target platform
[ERROR]   eclipse-plugin artifact with ID "org.eclipse.equinox.security.win32.x86_64" and version matching "0.0.0" was not found in the target platform

rgrunber:
- Fixes #230
- Update watch_server task to comply with Gulp v4

This isn't ideal, because the buid isn't as fast as it could be, but it does fix watch_server & dev_server.

Unfortunately I'm not sure how they were ever built for a single OS only. Maybe earlier Tycho versions allowed this ? Basically, the JDT-LS tarball includes bundle fragments for all platforms (so one tarball can be used everywhere). If we try to restrict the target platform to just a single OS then the product build will fail since each of the platform specific bundles is mandatory.

If there's a way to make the platform specific bundles optional then maybe this could work.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -3,6 +3,7 @@ const gulp = require('gulp');
 const cp = require('child_process');
 const decompress = require('gulp-decompress');
 const download = require('gulp-download');
+const glob = require('glob');
 const fse = require('fs-extra');
 const path = require('path');
 const server_dir = '../eclipse.jdt.ls';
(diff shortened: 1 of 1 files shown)


rgrunber:
The new approach basically packages just `org.eclipse.jdt.ls.core` (from target/) and copies it into the location of `server/plugins/org.eclipse.jdt.ls.core_*.jar` . By copying it and overwriting the old location, we can keep all the references in the product file the same.

This approach should take less time to build (and even less if we can ignore the gradle version checksum logic), in the common use-case of just making a change to jdt.ls.core and needing a quick rebuild.

PR #2118, opened 2021-09-17
Upversion to 1.0.0

rgrunber:

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "author": "Red Hat",
   "icon": "icons/icon128.png",
   "license": "EPL-2.0",
-  "version": "0.82.0",
+  "version": "0.83.0",
   "publisher": "redhat",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "author": "Red Hat",
   "icon": "icons/icon128.png",
   "license": "EPL-2.0",
-  "version": "0.82.0",
+  "version": "1.0.0",
   "publisher": "redhat",
   "bugs": "https://github.com/redhat-developer/vscode-java/issues",
   "preview": true,
```
rgrunber, line 11: Should we be setting this to `false`, or should we wait closer to the release for this.

fbricon:
I'd rather have 0.83.0 for now and set 1.0.0 for the actual release (and then we can remove the preview property)

fbricon (approved):

Issue #1061, opened 2019-09-15
Java Code Formatter Inconsistencies (Again...)


TheBrenny:
It seems that the extension is ignoring either my formatting profile, doing its own thing, or a combination of both.

Having Eclipse and VSCode open at the same time, using the same profile (afai can tell), both programs format the code differently: Eclipse does it the way I expect, VSCode literally does its own thing...

I thought it was because of the formatter profile version, so I've played around there with no success. Tried editing the XML file, changes occur, but no the goal wasn't seen.

I've searched open and closed issues for "formatter" and found #863 open and dormant, I also found #957 which is actually a different issue than expected.

##### Version Info
```
Operating System: Windows 10.0.17134
JDK version: 1.8
Visual Studio Code version: 1.38.0
Java extension version: 0.48.0
```
##### Steps To Reproduce
1. Grab one of the code files from [this gist](https://gist.github.com/TheBrenny/e2035b8dea9e9a2914ebf52b9f2d8f6a). They have the same code but have different formatting.
2. Open the code in Eclipse and format using the profile found in the same gist.
3. Open VSCode and format the same file. Notice that the code changes. It shouldn't (or be documented somewhere that it should).
4. `confusion *= 1000;`.

##### Results
[Current Result](https://gist.github.com/TheBrenny/e2035b8dea9e9a2914ebf52b9f2d8f6a#file-snowpanel-vscode-java)

[Expected Result](https://gist.github.com/TheBrenny/e2035b8dea9e9a2914ebf52b9f2d8f6a#file-snowpanel-eclipse-java)

Take note of the difference on [line 44](https://gist.github.com/TheBrenny/e2035b8dea9e9a2914ebf52b9f2d8f6a#file-snowpanel-vscode-java-L44) and particularly [line 282](https://gist.github.com/TheBrenny/e2035b8dea9e9a2914ebf52b9f2d8f6a#file-snowpanel-vscode-java-L282).

[Here's the log file](https://gist.github.com/TheBrenny/e2035b8dea9e9a2914ebf52b9f2d8f6a#file-vscode-java-log-file-json-txt)

karansood:
One specific setting that it seems to ignore is `org.eclipse.jdt.core.formatter.continuation_indentation`

DanieleStrafile:
[...]

PR #2138, opened 2021-09-29
Fix Open Java Formatter Settings

snjeza:
Steps to reproduce:
- set the following property
```
"java.format.settings.url": "file:///<some_path>/format.xml"
```
- call Java: Open Formatter Settings

See https://github.com/redhat-developer/vscode-java/issues/1061#issuecomment-929557096

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -673,7 +673,7 @@ async function openFormatter(extensionPath) {
 	const formatterUrl: string = getJavaConfiguration().get('format.settings.url');
 	if (formatterUrl && formatterUrl.length > 0) {
 		if (isRemote(formatterUrl)) {
-			commands.executeCommand(Commands.OPEN_BROWSER, Uri.parse(formatterUrl));
+			return commands.executeCommand(Commands.OPEN_BROWSER, Uri.parse(formatterUrl));
 		} else {
 			const document = getPath(formatterUrl);
 			if (document && fs.existsSync(document)) {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -742,7 +742,7 @@ function openDocument(extensionPath, formatterUrl, defaultFormatter, relativePat
 }
 
 function isRemote(f) {
-	return f !== null && f.startsWith('http:/') || f.startsWith('https:/');
+	return f !== null && f.startsWith('http:/') || f.startsWith('https:/' || f.startsWith('file:/'));
```
rgrunber, line 745: This is a typo right ? : `f.startsWith('https:/'` . No closing round bracket ? Couldn't figure out why it wasn't returning true.

rgrunber: Ah, and the extra bracket at the end of the line.

rgrunber:
The other issue I'm seeing is that at https://github.com/redhat-developer/vscode-java/blob/master/src/extension.ts#L689, fileName takes the value of formatterUrl (eg. `file:///home/rgrunber/Scripts/all.xml`, which gets joined to the workspace folder becoming something like `'/home/rgrunber/sample-projects/json-example/file:/home/rgrunber/Scripts/all.xml'`. This doesn't exist, so I get prompted to create a proper formatter path, but the current one should be fine.

src/extension.ts

rgrunber:
@snjeza what about the prompt that comes up afterwards, from the above comment ?

snjeza:
> @snjeza what about the prompt that comes up afterwards, from the above comment ?

Checking...

Issue #2109, opened 2021-09-14
Support platform specific extension


testforstephen:
See https://github.com/microsoft/vscode/issues/23251#issuecomment-918989339, VS Code Marketplace now allows extensions to publish different VSIXs for each platform (Windows, Linux, macOS). We might consider taking advantage of this by packaging a platform-specific Java runtime in the extension itself.

This document has all the details:
https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions

rgrunber:
LGTM https://download.jboss.org/jbosstools/jdt.ls/staging/java-1.2.0-476/

PR #2183, opened 2021-10-27
Support platform specific extension

testforstephen:

This will address the issue https://github.com/redhat-developer/vscode-java/issues/2109

Below are the tasks added by PR:
- [x] Add a gulp task to download a platform specific JRE.
      The full version is `npx gulp download_jre --target darwin-x64 --javaVersion 17`.
      The short version is `npx gulp download_jre`, which will download latest JRE for your dev machine platform.

   This task will download and decompress the artifact to a jre folder under root directory of vscode-java.
image

- [x] Adjust the JDK search logic in Java extension to allow using embedded JRE to launch Java extension itself.
      **1)** We plan to embed `org.eclipse.justj.openjdk.hotspot.jre.full.stripped-17` instead of `org.eclipse.justj.openjdk.hotspot.jre.minimal.stripped-17`. The minimal stripped is about 30M, and the full jre stripped is about 55M (only plus 25M). But the full jre stripped includes a full set of Java modules, and can cover maven project import better.
https://download.eclipse.org/justj/jres/17/downloads/latest/

    **2)** For platform extensions, we will always use the embedded JRE to launch Java extension. "java.home" setting is useless and may cause confusion, we will mark it as deprecated. And it can only be used for universal versions without embedded JRE.

    **3)** For platform extensions, it will still detect the installed JDKs in local machine. If no one was found, then pop up to install a JDK to compile users' project. The difference is we don't limit the JDK version any more. The concern is because the embedded JRE doesn't include JDK source, GTD cannot find source. To avoid confusion later, we can ask the user to install a project JDK when starting the Java extension.

- [x] Update CI job (Jenkinsfile) to allow publishing platform specific extension to VS Code marketplace.
       Since JustJ only provides 5 JRE platforms, we're going to support embedded JRE for these platforms. For other platforms, we publish a universal one without embedded JRE.
        
  Below are the CI artifacts, it includes a generic version for not supported platforms, 5 platform specific versions.
image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,7 +46,8 @@ node('rhel8'){
 
 	stage "Package vscode-java"
 	def packageJson = readJSON file: 'package.json'
-	sh "vsce package -o java-${packageJson.version}-${env.BUILD_NUMBER}.vsix"
+	env.EXTENSION_VERSION = "${packageJson.version}"
+	sh "vsce package -o java-${env.EXTENSION_VERSION}-${env.BUILD_NUMBER}.vsix"
 
 	stage 'Test vscode-java for staging'
 	wrap([$class: 'Xvnc']) {
@@ -64,14 +65,18 @@ node('rhel8'){
 	def embeddedJRE = 17
 	for(platform in platforms){
 		sh "npx gulp download_jre --target ${platform} --javaVersion ${embeddedJRE}"
-		sh "vsce package --target ${platform} -o java-${platform}-${packageJson.version}-${env.BUILD_NUMBER}.vsix"
+		sh "vsce package --target ${platform} -o java-${platform}-${env.EXTENSION_VERSION}-${env.BUILD_NUMBER}.vsix"
 	}
 	stash name:'platformVsix', includes:'java-win32-*.vsix,java-linux-*.vsix,java-darwin-*.vsix'
 
 	stage 'Upload vscode-java to staging'
 	def artifacts = findFiles(glob: '**.vsix')
+	def artifactDir = "java-${env.EXTENSION_VERSION}-${env.BUILD_NUMBER}"
+	sh "mkdir ${artifactDir}"
+	sh "mv *.vsix ${artifactDir}"
+
 	for(artifact in artifacts){
-		sh "rsync -Pzrlt --rsh=ssh --protocol=28 ${artifact.path} ${UPLOAD_LOCATION}/jdt.ls/staging"
+		sh "rsync -Pzrlt --rsh=ssh --protocol=28 --relative ${artifactDir}/${artifact.path} ${UPLOAD_LOCATION}/jdt.ls/staging"
 	}
 }
 
@@ -104,13 +109,17 @@ node('rhel8'){
 			}
 		}
 
-		archive includes:"**.vsix"
-
 		stage "Publish to http://download.jboss.org/jbosstools/static/jdt.ls/stable/"
 		def artifacts = findFiles(glob: '**.vsix')
+    	def artifactDir = "java-${env.EXTENSION_VERSION}"
+	    sh "mkdir ${artifactDir}"
+	    sh "mv *.vsix ${artifactDir}"
+
+		archive includes:"${artifactDir}/**/*.*"
+
 		// copy this stable build to Akamai-mirrored /static/ URL, so staging can be cleaned out more easily
 		for(artifact in artifacts){
-			sh "rsync -Pzrlt --rsh=ssh --protocol=28 ${artifact.path} ${UPLOAD_LOCATION}/static/jdt.ls/stable/"
+			sh "rsync -Pzrlt --rsh=ssh --protocol=28 --relative ${artifactDir}/${artifact.path} ${UPLOAD_LOCATION}/static/jdt.ls/stable/"
 		}
 	}// if publishToMarketPlace
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -55,10 +55,24 @@ node('rhel8'){
 		sh "npm test --silent"
 	}
 
-	stage 'Upload vscode-java to staging'
 	def vsix = findFiles(glob: '**.vsix')
-	sh "rsync -Pzrlt --rsh=ssh --protocol=28 ${vsix[0].path} ${UPLOAD_LOCATION}/jdt.ls/staging"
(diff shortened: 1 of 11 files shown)


src/requirements.ts
```diff
@@ -34,12 +35,13 @@ interface ErrorData {
  *
  */
 export async function resolveRequirements(context: ExtensionContext): Promise<RequirementsData> {
+    let toolingJre: string = await findEmbeddedJRE(context);
```
testforstephen, line 38: @Eskibear for the latest JDK requirement change in vscode-java. pls check if vscode-java-pack need update as well.

Eskibear: sure

package.json
```diff
@@ -21,7 +21,7 @@
     "virtualWorkspaces": false
   },
   "engines": {
-    "vscode": "^1.53.2"
+    "vscode": "^1.61.0"
```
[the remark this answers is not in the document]
rgrunber: Yup. I was expecting this to be an issue in case Theia checked the engine version and rejected based on that, but it seems to be working, so we should be good.

src/requirements.ts
```diff
@@ -54,25 +57,102 @@ export async function resolveRequirements(context: ExtensionContext): Promise<Re
...
+        });
     });
 }
 
+async function findEmbeddedJRE(context: ExtensionContext): Promise<string | undefined> {
+    const jreHome = context.asAbsolutePath("jre");
+    if (fse.existsSync(jreHome) && fse.statSync(jreHome).isDirectory()) {
+        if (fse.existsSync(path.join(jreHome, "bin", JAVA_FILENAME))) {
+            return jreHome;
+        }
```
rgrunber, line 118: Is there ever a case where the path is `./jre/bin/java` . Looking at the packaging logic in download_jre of the gulpfile.js, the archive is extracted into `'./jre/' + jreVersionLabel` and `jreVersionLabel` is either the version-platform-arch or the name of the file.

rgrunber (changes requested):
Overall, works well for me.

- Under the Optional section of "Server Side Setup" in CONTRIBUTING.md, I think we should mention that a JRE can be embedded into the extension. I guess it could even go under the "Client Side Setup".

Just mentioning `npx gulp download_jre` will build a platform specific JRE but that it has options `target` and `javaVersion`, is probably enough.

- I found it useful to add something like this for debugging :

```diff
diff --git a/.vscode/launch.json b/.vscode/launch.json
index 2620f81..f502215 100644
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -110,6 +110,26 @@
                        "outFiles": ["${workspaceRoot}/out/**/*.js"],
                        "preLaunchTask": "prepareLightweightTest",
                        "postDebugTask": "cleanTestFolder"
+               },
+               {
+                       "args": [
+                               "${input:gulpTask}"
+                       ],
+                       "name": "Launch Gulp Task",
+                       "program": "${workspaceFolder}/node_modules/gulp/bin/gulp.js",
+                       "request": "launch",
+                       "skipFiles": [
+                               "<node_internals>/**"
+                       ],
+                       "type": "pwa-node"
                }
-       ]
+       ],
+       "inputs": [
+        {
+            "id": "gulpTask",
+            "type":"promptString",
+            "description": "Name of the Gulp task to execute",
+            "default": "download_jre"
+        }
+    ]
 }
```

src/requirements.ts
testforstephen: No, the gulp task always decompresses the embedded jre to `'./jre/' + jreVersionLabel`. I will remove the unused condition branch.

gulpfile.js
```diff
@@ -6,12 +6,128 @@ const download = require('gulp-download');
...
+		});
+
+		if (!fse.existsSync('./jre/justj.manifest')) {
+			cleanManifest();
+			done(new Error(`Failed to download justj.manifest, please check if the link ${mafinestUrl} is valid.`))
+			return;
+		}
+
+		const value = fse.readFileSync('./jre/justj.manifest').toString();
+		if (value.startsWith("<!DOCTYPE html>")) {
```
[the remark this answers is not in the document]
testforstephen: The lib `gulp-download` we use doesn't expose http response status to the callback, so I will use `request` lib to get justj.manifest, which supports error status check.

testforstephen:
> Under the Optional section of "Server Side Setup" in CONTRIBUTING.md, I think we should mention that a JRE can be embedded into the extension. I guess it could even go under the "Client Side Setup".
Just mentioning `npx gulp download_jre` will build a platform specific JRE but that it has options target and javaVersion, is probably enough.

I mentioned `npx gulp download_jre` under the part "Client Side Setup", since the embedded JRE is the client related stuff.

> I found it useful to add something like this for debugging :

Nice suggestion 👍

Jenkinsfile
```diff
@@ -55,10 +55,24 @@ node('rhel8'){
...
+	for(platform in platforms){
+		sh "npx gulp download_jre --target ${platform} --javaVersion ${embeddedJRE}"
+		sh "vsce package --target ${platform} -o java-${platform}-${packageJson.version}-${env.BUILD_NUMBER}.vsix"
+	}
+	stash name:'platformVsix', includes:'java-win32-*.vsix,java-linux-*.vsix,java-darwin-*.vsix'
+
+	stage 'Upload vscode-java to staging'
+	def artifacts = findFiles(glob: '**.vsix')
+	for(artifact in artifacts){
+		sh "rsync -Pzrlt --rsh=ssh --protocol=28 ${artifact.path} ${UPLOAD_LOCATION}/jdt.ls/staging"
```
rgrunber, line 74: @fbricon we currently place the vsix files under https://download.jboss.org/jbosstools/jdt.ls/staging/ & https://download.jboss.org/jbosstools/static/jdt.ls/stable/ . Do we want to create a separate directory for each version ? Any requirements to to clean staging in the future given how much faster it'll grow in size ?

rgrunber (approved):
Change looks good to me. Just a storage consideration and I think we're ready to merge.

Jenkinsfile
testforstephen: The platform extensions uploaded to staging will be placed at a folder like "*/jdt.ls/staging/java-1.2.0-<BUILD_NUMBER>".

The artifacts uploaded to stable will be placed at a folder like "*/jdt.ls/stable/java-1.2.0".

PR #2203, opened 2021-11-09
Clean unused Java workspace storage folders

### LINKED ISSUE #2110: Limit what you write to workspaceStorage
My ~/Library/Application Support/Code/User/workspaceStorage folder is 95GB
Pretty much all of this comes from the content of `redhat.java` folders

There are 802 such folders inside `workspaceStorage/*`

The biggest folder contains 6.7GB stuff. That contains 64505 files named `client.log.2021-08-17.<number>` with apparently some trace logging. Not sure if I at one point tried to debug something and maybe caused this. Still seems a bit excessive.

Most folders are around 100MB and all space is consumed by `redhat.java/jdt_ws/.metadata/.plugins/org.eclipse.jdt.core/*.index` files.

Are you indexing the same dependencies over and over again for each workspace or something similar?

For comparison, my whole ~/.m2 folder is 18GB so using 95GB for workspace storage does not seem reasonable.

##### Environment
- Operating System: macOs Big Sur
- JDK version: 16
- Visual Studio Code version: 1.60.0
- Java extension version: 0.81.0

##### Steps To Reproduce
1. Code with VS code
2. Code some more
3. Code even more

rgrunber:
- Clean unused workspace storage folders on startup
- Create settings "java.configuration.cleanWorkspaceCache", to allow
  users to configure maximum number of days to preserve cached data
- Fixes #2110

### DIFF
--- a/package.json
+++ b/package.json
@@ -343,6 +343,16 @@
           "description": "Specifies severity if the plugin execution is not covered by Maven build lifecycle.",
           "scope": "window"
         },
+        "java.configuration.workspaceCacheLimit": {
+          "type": [
+            "null",
+            "integer"
+          ],
+          "default": null,
+          "minimum": 1,
+          "description": "The number of days (if enabled) to keep unused workspace cache data. Beyond this limit, cached workspace data may be removed.",
+          "scope": "application"
+        },
         "java.format.enabled": {
           "type": "boolean",
           "default": true,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -343,6 +343,16 @@
           "description": "Specifies severity if the plugin execution is not covered by Maven build lifecycle.",
           "scope": "window"
         },
+        "java.configuration.workspaceCacheLimit": {
```
testforstephen, line 346: We could add the new setting to README as well.

testforstephen (approved):
LGTM.

package.json
rgrunber: Done. Added under `New in 1.1.0`

PR #2214, opened 2021-11-15
fix: Java build status spinning forever

Eskibear:
To fix https://github.com/microsoft/vscode-java-pack/issues/769

Signed-off-by: Yan Zhang <yanzh@microsoft.com>

### DIFF
--- a/src/serverTaskPresenter.ts
+++ b/src/serverTaskPresenter.ts
@@ -39,6 +39,7 @@ async function getPresenterTaskExecution(): Promise<TaskExecution> {
 		echo: false,
 		focus: false
 	};
+	presenterTask.isBackground = true;
 
 	return presenterTaskExecution = await tasks.executeTask(presenterTask);
 }


src/serverTaskPresenter.ts
```diff
@@ -39,6 +39,7 @@ async function getPresenterTaskExecution(): Promise<TaskExecution> {
 		echo: false,
 		focus: false
 	};
+	presenterTask.isBackground = true;
```
fbricon, line 42: When was this introduced in the API? Asking to know if this'll work in theia

rgrunber: I think `isBackground` has been around for a while, and supported by Theia ( https://che-incubator.github.io/vscode-theia-comparator/status.html ). However, I'm guessing the terminal only recently started using the property to determine whether to have a spinning icon beside the terminal label ?

Eskibear: `isBackground` was introduced four years ago, since this task is long-running, I believe it's correct to mark it as `isBackground`. 

> I'm guessing the terminal only recently started using the property to determine whether to have a spinning icon beside the terminal label ?

Yes, I tried some previous version e.g. 1.50 as below, there's no visualization of the state at that time, and terminals are in dropdown menu... It has been visualized since vscode refine the terminal UX to present all terminals together.
image

rgrunber:
Theia doesn't show a spinning icon for the terminal at all and when installing vscode-java with this change in Theia, I see no visible difference, so it should be safe. I think this is more because Theia doesn't have a view that displays all terminals at once, but instead allows a user to select them from the command palette. I think this should be fine to merge.

fbricon:
Thanks @Eskibear!

PR #2240, opened 2021-12-03
Provide client-side snippet until language server is ready

### LINKED ISSUE #684: Code snippet should be working before LS done resolving Maven dependencies
Maven dependencies take lots of time to resolve.
image

At that moment, LS is not responding to user key strokes. But some of the features like code snippet can acutally work. The idea is to enable some language servers before the project is completely loaded. Code snippet can be a good start because it require less project context. For example, the `public_method` does not require project context but only the surrounding class.

jdneo:
This PR fix #684 
- postpone the client side snippet provider dispose time when the server is ready.
- dispose the provider to make sure the items will be removed in the completion list when disposing happens.

Signed-off-by: Sheng Chen<sheche@microsoft.com>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -22,7 +22,7 @@ import { registerClientProviders } from './providerDispatcher';
 import * as fileEventHandler from './fileEventHandler';
 import { StandardLanguageClient } from './standardLanguageClient';
 import { apiManager } from './apiManager';
-import { SnippetCompletionProvider } from './snippetCompletionProvider';
+import { snippetCompletionProvider } from './snippetCompletionProvider';
 import { runtimeStatusBarProvider } from './runtimeStatusBarProvider';
 import { serverStatusBarProvider } from './serverStatusBarProvider';
 import { markdownPreviewProvider } from "./markdownPreviewProvider";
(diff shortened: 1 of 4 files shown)


jdneo:
This is a great finding! Currently we register the capabilities earlier than the server is able to dealing with those request. During this time window, if user starts to typing or doing other operations, a lot of requests sending from client to server and queued there, which makes the import time becomes 'longer'. Server needs to treat with those queued request but most of them are out-of-date.

testforstephen:
There is a feature request https://github.com/microsoft/vscode/issues/107343 about showing results from different completion providers incrementally, but it's still under discussion. 

if we want to mitigate the issue for now, we need to delay registering completion capability on the server side until the language server is actually ready.

testforstephen:
It's possible to show duplicated snippet in completion list when lightweight server is up but standard server is still loading. both lightweight server and client-side snippet provide same snippets.

image

testforstephen (approved):
It works fine with the latest commit. Feel free to merge this PR once https://github.com/eclipse/eclipse.jdt.ls/pull/1979 is merged.

PR #2244, opened 2021-12-06
Show the error status when imported projects having errors

jdneo:
requires https://github.com/eclipse/eclipse.jdt.ls/pull/1962

### DIFF
--- a/src/serverStatus.ts
+++ b/src/serverStatus.ts
@@ -5,6 +5,7 @@ import { serverTasks } from "./serverTasks";
 
 export enum ServerStatusKind {
 	Ready = "Ready",
+	Warning = "Warning",
 	Error = "Error",
 	Busy = "Busy",
 }
@@ -24,6 +25,8 @@ function fireEvent() {
 
 export namespace serverStatus {
 
+	let hasError: boolean = false;
+
 	export const onServerStatusChanged = _emitter.event;
 
 	export function initialize() {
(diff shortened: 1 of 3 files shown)


rgrunber:
Using https://github.com/spring-projects/spring-boot, this doesn't seem to work for me as I'd expect. I've built the JDT-LS change along with this change, but even with breakpoints on the [`Error`](https://github.com/redhat-developer/vscode-java/blob/f505f60d74abecbdd7bd290ff6549406c870c484/src/standardLanguageClient.ts#L117) case in the client, or at the [`GradleProjectImporter`](https://github.com/eclipse/eclipse.jdt.ls/blob/62f20895b67378327df9081742c2812c63651a47/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/managers/GradleProjectImporter.java#L201) in the server, nothing is triggered.

The import of does result in some errors reported in the problems tab, but the thumbs up (ServiceReady) is also shown.

jdneo:
@rgrunber did you also adopt the PR at the LS side? https://github.com/eclipse/eclipse.jdt.ls/pull/1962

rgrunber (approved):
It turns out my problem was the project likely built and all I was seeing were compilation errors. I introduced an issue in the build file and this allowed me to see this change works.

jdneo:
@rgrunber This PR is still use the stable API to show the status when project configuration has errors.

It use 👎(same as when server crash) when project configuration has errors. Or maybe a ⚠️ icon will be better?

testforstephen (approved):
LGTM

PR #2246, opened 2021-12-13
use jdk-utils to detect installed runtimes

Eskibear:
use [jdk-utils](https://github.com/Eskibear/node-jdk-utils) to detect installed Java runtimes.


testforstephen (approved):
LGTM

PR #2252, opened 2021-12-17
Fix Jenkinsfile to publish the correct generic version

testforstephen:


rgrunber (approved):
Seems like this should fix things, assuming some previous build artifacts were affecting the deployment.

Issue #2254, opened 2021-12-19
Version 1.2.0 breaks all functionality


brunovieira97:
I've been using the extension for some time now. A couple of days ago, I've updated it to 1.2.0 and since then, when opening a study project it would report errors for everything.

Some of the errors include not finding `Object`'s default no args constructor, failing to import basic core classes from Java (i.e. `java.util`, `java.date`), and so on.

I've already removed my JDK 17 installation, replacing it with JDK 11, and that had no effect. Only fix I've found is to rollback the extension to the previous version (1.1.0, from 3w ago).

Some screenshots are attached to show the errors I'm facing.

##### Environment
- Operating System: macOS 12.1 Monterey arm64 (Apple Silicon)
- JDK version: openjdk version "17.0.1" 2021-10-19, installed via Homebrew
- Visual Studio Code version: 1.63.2 (Universal)
- Java extension version: 1.2.0

##### Steps To Reproduce
1. Clone example project
2. Open project folder in VS Code
3. Open any .java file

Example project: https://github.com/brunovieira97/alura-novidades-java

Logs: [Logs.zip](https://github.com/redhat-developer/vscode-java/files/7741131/Logs.zip)

##### Current Result

![image](https://user-images.githubusercontent.com/19749621/146682624-0faffd37-ae2f-44d2-bf89-df8be87bf574.png)
![image](https://user-images.githubusercontent.com/19749621/146682636-5007bc9b-19b1-4fe1-9573-30d393ff5e3d.png)
![image](https://user-images.githubusercontent.com/19749621/146682643-5139c92b-75f3-4b34-a49d-09f25c95a99d.png)
![image](https://user-images.githubusercontent.com/19749621/146682659-9c3cc5fa-d734-484e-be3f-eafb1baeb205.png)

##### Expected Result (v1.1.0)

![image](https://user-images.githubusercontent.com/19749621/146682679-53a578ad-5f3b-4f35-a4ed-8d6d57b145cd.png)

##### Additional Informations

- Reinstalling the extension had no effect
- Tried removing VS Code completely and reinstalling as well
- As previously stated, I removed `openjdk` via brew and installed `openjdk@11`, properly linking it. No effect at all
- Log for "Language Support for Java (Syntax Server) shows this message, both with 1.1.0 and 1.2.0:
```
[Error - 13:29:00] Connection to server got closed. Server will not be restarted.
```

snjeza:
[...]

PR #2258, opened 2021-12-21
update jdk-utils to detect JDKs installed by Homebrew

Eskibear:
See #2254 

[changelog](https://github.com/Eskibear/node-jdk-utils/blob/main/CHANGELOG.md#042) of jdk-utils


testforstephen (approved):
LGTM

PR #2273, opened 2022-01-05
Add '--add-opens java.base/sun.nio.fs=ALL-UNNAMED' to the JLS jvmargs

### LINKED ISSUE #2264: Projects not recognized, nothing works.

##### Environment
- Operating System: `Windows 10`
- JDK version: `OpenJDK 64-Bit Server VM Temurin-17.0.1+12 (build 17.0.1+12`
- Visual Studio Code version: `1.63.2`
- Java extension version: `Language Support for Java(TM) by Red Hat v1.2.0`

##### Steps To Reproduce

1. Install `Extension Pack for Java v0.20.0` (including the `Language Support for Java(TM) by Red Hat v1.2.0`)
1. Try do anything, open existing project, code Java in VS Code...

##### Current Result

Hardly anything works.

##### Expected Result

Actually anything.

##### Additional Informations

Logs: https://gist.github.com/AgainPsychoX/1c8d75f00fd53282fd6fafd7becf633c

testforstephen:

Fix #2264

### DIFF
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -64,7 +64,11 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
 				'--add-opens',
 				'java.base/java.util=ALL-UNNAMED',
 				'--add-opens',
-				'java.base/java.lang=ALL-UNNAMED');
+				'java.base/java.lang=ALL-UNNAMED',
+				// See https://github.com/redhat-developer/vscode-java/issues/2264
+				// It requires the internal API sun.nio.fs.WindowsFileAttributes.isDirectoryLink() to check if a Windows directory is symlink.
+				'--add-opens',
+				'java.base/sun.nio.fs=ALL-UNNAMED');
 
 	params.push('-Declipse.application=org.eclipse.jdt.ls.core.id1',
 				'-Dosgi.bundles.defaultStartLevel=4',


fbricon:
Please check it runs fine if JDK11 is used (it should be fine)

rgrunber:
I'm able to run with Java 11 and this PR.

However, in the process of testing, I tried to use JAVA_HOME to override the JRE used for JDT-LS. This kept failing until I realized that [it would never work](https://github.com/redhat-developer/vscode-java/blob/master/src/javaServerStarter.ts#L32). We seem to use `requirements.java_home` only for the default JDK configured in JDT-LS in the event there are none. Just making sure we're ok with this since I got the impression we wanted JAVA_HOME to override the embedded JRE as well.

testforstephen:
yes, "--add-opens java.base/sun.nio.fs=ALL-UNNAMED" doesn't break JDK 11 to launch JLS. "--add-opens" argument is introduced from JDK 9, it works from JDK 9.

> Just making sure we're ok with this since I got the impression we wanted JAVA_HOME to override the embedded JRE as well.

Not quite right. We're going to introduce a new setting "java.jdt.ls.java.home", which will override the embedded JRE if it's set. In other cases, the embedded JRE will be picked first as tooling JDK. Since this comment is not related with this PR, let's move the discussion to a new PR about new setting "java.jdt.ls.java.home".

PR #2284, opened 2022-01-14
Add a setting 'java.jdt.ls.java.home' to manually set up a JDK to start JLS

### LINKED ISSUE #2276: Provide a way to disable embedded JRE
The embedded JRE assumes a traditional FHS directory structure on Linux. Nontraditional distributions, like NixOS, do not have directories like _/lib_ and so the embedded Java executable cannot be used. The following opaque error occurs when opening a Java file on NixOS:
```
no such file or directory: /home/user/.vscode/extensions/redhat.java-1.2.0/jre/17.0.1-linux-x86_64/bin/java
```

The file exists, but it requests a nonexistent interpreter, _/lib64/ld-linux-x86-64.so.2_. Deleting the _jre_ directory allows me to use the extension, but that is an unpleasant user experience. A setting to ignore the embedded JRE would be appreciated.

testforstephen:

Close #2276

### DIFF
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -34,11 +34,12 @@ export async function resolveRequirements(context: ExtensionContext): Promise<Re
     let toolingJre: string = await findEmbeddedJRE(context);
     let toolingJreVersion: number = await getMajorVersion(toolingJre);
     return new Promise(async (resolve, reject) => {
-        let source: string;
+        const javaPreferences = await checkJavaPreferences(context);
+        const preferenceName = javaPreferences.preference;
+        let javaHome = javaPreferences.javaHome;
         let javaVersion: number = 0;
-        let javaHome = await checkJavaPreferences(context);
-        if (!toolingJre && javaHome) { // "java.home" setting is only used for the universal version.
-            source = `java.home variable defined in ${env.appName} settings`;
+        if (javaHome) {
+            const source = `${preferenceName} variable defined in ${env.appName} settings`;
             javaHome = expandHomeDir(javaHome);
             if (!await fse.pathExists(javaHome)) {
                 invalidJavaHome(reject, `The ${source} points to a missing or inaccessible folder (${javaHome})`);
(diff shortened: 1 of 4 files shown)


testforstephen:
Add @Eskibear as a reviewer

src/requirements.ts
```diff
@@ -52,51 +53,49 @@ export async function resolveRequirements(context: ExtensionContext): Promise<Re
...
+                javaHome = validJdks[0].homedir;
+                javaVersion = validJdks[0].version.major;
+                toolingJre = javaHome;
+                toolingJreVersion = javaVersion;
+            }
+        } else { // pick a default project JDK/JRE
+            /**
+             * For legacy users, we implicitly following the order below to
+             * set a default project JDK during initialization:
+             * java.jdt.ls.java.home > java.home > env.JDK_HOME > env.JAVA_HOME > env.PATH
```
Eskibear, line 79: IMO `java.jdt.ls.java.home` should be dedicated to specify tooling JRE. I'd  prefer not to consider this setting when we set a default project runtime.

Eskibear:
This PR introduces `java.jdt.ls.java.home` allowing to specify runtime to launch jdtls, and fallback to check the deprecated `java.home` setting. Overall LGTM.

BTW please also add the new setting into `restrictedConfigurations` section for workspace trust.

src/requirements.ts
testforstephen: I understand your concern is to separate the concept of tooling jdk and project jdk. The problem here is that we have to set a default JDK for JDT to compile Java projects that do not have explicitly configured runtimes.

To be honest, the rules about how to set a default JDK is actually an implicit implementation of Java extension, the user doesn't aware of the logic involved. Unless we can make the UX of the default project JDK to be more transparent, I don't think it makes any difference to choose "java.jdt.ls.java.home" or "env.JAVA_HOME" first as default project jdk. 

Besides,  "java.jdt.ls.java.home" is more explicit to me than choosing a JDK from JAVA_HOME, bin path and etc., because it is manually set by user.

Eskibear: Got it. I'm ok with it.
After all, inferring an available JDK as default project runtime is not a big problem, as long as `java.configuration.runtimes` has higher priority regarding to project runtimes.

Actually another problem is how we improve experience of project runtime management if you want to use different JDK distros, e.g. allow workspace level `java.configuration.runtimes` just like eclipse. But that's another story, let's discuss it in a separate issue.

rgrunber (approved):
It's working for me. The property can be used the override the embedded tooling.

I would make sure to update https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements#setting-the-jdk in the wiki eventually.

testforstephen:
> I would make sure to update https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements#setting-the-jdk in the wiki eventually.

wiki updated as well.

PR #2303, opened 2022-01-27
Use ovsx<0.3.0 to ensure we build with Node v12.

rgrunber:


rgrunber:
Due to https://github.com/eclipse/openvsx/pull/403. I believe Theia is still preventing us from upgrading to v14.

PR #2330, opened 2022-02-23
Set java.configuration.workspaceCacheLimit default value to 90.

rgrunber:
- Avoid deleting temporary workspaces as they should have their own
  policy defined by the operating system
- Enable cleaning of cache for workspaces that have been inactive for
  over 90 days

We should set some larger value for the setting (eg `90`), otherwise many users are unlikely to benefit from this. I've also disabled the deletion for temporary workspace folders mainly because the OS should handle policy for temp folders (if they're not deleted on every reboot anyways).

See #2110 for more information.

### DIFF
--- a/package.json
+++ b/package.json
@@ -361,7 +361,7 @@
             "null",
             "integer"
           ],
-          "default": null,
+          "default": 90,
           "minimum": 1,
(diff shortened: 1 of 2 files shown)


Eskibear (approved):
This makes total sense to me.

testforstephen (approved):

Issue #2001, opened 2021-06-25
java.configuration.runtimes in settings.json are ignored


william-mason:
[provide a description of the issue]

##### Environment
- Operating System: Ubuntu linux 20.10
- JDK version: several -- 1.8, 16
- Visual Studio Code version: 1.52.1
- Java extension version: v0.78.0

##### Steps To Reproduce
1. Open VS Code
2. Open settings preferences
3. Enter/search for: java.configuration.runtimes
4. Enter name and path for different JDK-s
5. Nominate a default

There is no project yet -- I can't even establish a session with the JDK of choice.  Also the plugin goes off and finds some Java JDK-s that are not the ones I wish to use.  There does not seem to be any place were I can say what I need to use in the project. 

[Please attach logs](https://github.com/redhat-developer/vscode-java/wiki/Troubleshooting#enable-logging)
-  This gave me the following error:

   ```
   Command 'Java: Open Java Language Server Log File' resulted in an error (command 'java.open.serverLog' not found)
   ```

##### Current Result

- JDK-s configured do not show in the 'Configure Java Runtime' on any tab.

##### Expected Result

- I at least wanted to see the Java versions I entered show up as project JDK-s
- I would also like them to show in the list of "instaled Java-s"
- I would like a place where I can explicitly specify which Java versions I want the Java Plugin to recognise -- ALL those I want to use at a particular time and No Other JDK-s even if there are installed.
  - Not all the relevant JDK-s are available in "expected" locations

rgrunber:
There's a few issues I see here :

- It looks like you're using the "Configure Java Runtime" from https://github.com/microsoft/vscode-java-pack and there may be some limitations on how it can be configured. With that said, I think you need to have some project imported before being able to configure the runtime. Build systems tend to allow configuring the source/compiler targets for Java, so usually such changes need to be made there. @testforstephen, thoughts ? There might be some enhancements that could be made to give users more control here.
- The "Configure Java Runtime" doesn't seem to be reading in any values a user may have set in `java.configuration.runtimes`
- Java commands contributed via package.json will fail if the extension itself hasn't been triggered (ie. due to no project imported). I wonder if we can fix this.
- There may be some locations we miss for locating java installations, though we do check JAVA_HOME, JDK_HOME, PATH, and /usr/lib/jvm/ on Linux. Are there additional locations ?

testforstephen:
[...]

PR #2337, opened 2022-03-02
Make java.configuration.runtimes machine-overridable

thakkarparth007:
Potential fix for https://github.com/redhat-developer/vscode-java/issues/2001

I have no idea if changing this configuration's scope can cause unintentional side effects. If so, do educate me and I'd be happy to make some code changes if it is not a lot.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
       "restrictedConfigurations": [
         "java.jdt.ls.java.home",
         "java.home",
-        "java.jdt.ls.vmargs"
+        "java.jdt.ls.vmargs",
+        "java.configuration.runtimes"
       ]
     },
     "virtualWorkspaces": false
--- a/package.json
+++ b/package.json
@@ -721,7 +721,7 @@
             "additionalProperties": false
           },
           "default": [],
-          "scope": "machine"
+          "scope": "machine-overridable"
         },
         "java.server.launchMode": {
           "type": "string",


rgrunber:
After reconsidering, I think this approach might work. Before my main concern was that allowing a project to set `java.configuration.runtimes` is a security risk in the same way allowing a project to dictate `java.jdt.ls.java.home` or `java.jdt.ls.vmargs` would be. But now we have workspace trust, and if we add the property to [`restrictedConfigurations`](https://github.com/redhat-developer/vscode-java/blob/master/package.json#L16-L19) it should handle that concern.

@testforstephen @fbricon , let me know if there's some case we're missing here.

testforstephen:
i think it's OK, other than we need to add `java.configuration.runtimes` as untrusted via `restrictedConfigurations`. 

https://github.com/redhat-developer/vscode-java/blob/083f580be70dfe7a6f76b0e92b18554a6f1b8c72/package.json#L13-L20

Eskibear:
Regarding this, I had a feature request. In `java.configuration.runtimes`, we store mapping of language level and path to JDK.
Currently I'm unable to use different JDK distribution for different project because the setting scope.

E.g. I have two Maven projects with language level set to `Java SE-1.8`. But for Project A I want to use RedHat OpenJDK, and Micrsoft's OpenJDK for Project B. Currently I'm not able to do that with workspace level setting.

I'm wondering if we'll have this feature with this PR? Or do we need change something in jdtls, as I remember the backend setting is also per machine.

rgrunber:
> Regarding this, I had a feature request. In `java.configuration.runtimes`, we store mapping of language level and path to JDK. Currently I'm unable to use different JDK distribution for different project because the setting scope.
> 
> E.g. I have two Maven projects with language level set to `Java SE-1.8`. But for Project A I want to use RedHat OpenJDK, and Micrsoft's OpenJDK for Project B. Currently I'm not able to do that with workspace level setting.
> 
> I'm wondering if we'll have this feature with this PR? Or do we need change something in jdtls, as I remember the backend setting is also per machine.

I just tried out the PR and it allows me to override `java.configuration.runtimes` from `.vscode/settings.json` in a workspace folder. It seems in line with what https://code.visualstudio.com/api/references/contribution-points implies about  machine-overridable.

I couldn't set a folder-specific setting on the client and get it to be sent to the language server, but maybe I'm not understanding how the setting should work. The problem I see is I don't think we even support a multi-project root. In `JVMConfigurator`, we call `JavaRuntime.setDefaultVMInstall(..)` to set the JVM, which says :

```
Sets a VM as the system-wide default VM, and notifies registered VM install
```

So this would only work with a single "project" per workspace.

Eskibear:
> So this would only work with a single "project" per workspace.

I think that's fine for me. I can always open different projects in different workspace for the moment.

But in Eclipse, when I import multiple projects, I can configure VMInstall per project. So I think it's at least feasible regardless of effort. In long-term plan, I believe we do want to fully support multi-root workspace, maybe we should think about how to refactor the JVMConfigurator then.

rgrunber:
> > So this would only work with a single "project" per workspace.
> 
> I think that's fine for me. I can always open different projects in different workspace for the moment.
> 
> But in Eclipse, when I import multiple projects, I can configure VMInstall per project. So I think it's at least feasible regardless of effort. In long-term plan, I believe we do want to fully support multi-root workspace, maybe we should think about how to refactor the JVMConfigurator then.

Yup! The JDT API definitely has some way to set the JVM per project. We've likely only used the global setting for now. As long as we're fine with some functionality being unsupported for the time being, I think it should be fine to proceed with this.

thakkarparth007:
@rgrunber done.

rgrunber:
If there's no opposition to this, or special case we've missed, I'll merge soon. Last remaining thing is to update https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements to briefly mention workspace settings can now take advantage of the setting.

rgrunber:
If there's no opposition to this, or special case we've missed, I'll merge soon. Last remaining thing is to update https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements to briefly mention workspace settings can now take advantage of the setting.

PR #2343, opened 2022-03-07
Remove 'java.completion.overwrite'

jdneo:
requires: https://github.com/eclipse/eclipse.jdt.ls/pull/2042

We use 'editor.suggest.insertMode' now to control the completion behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,6 @@ The following settings are supported:
 * `java.autobuild.enabled` : Enable/disable the 'auto build'.
 * `java.maxConcurrentBuilds`: Set max simultaneous project builds.
 * `java.completion.enabled` : Enable/disable code completion support.
-* `java.completion.overwrite` : When set to true, code completion overwrites the current text. When set to false, code is simply added instead.
 * `java.completion.guessMethodArguments` : When set to true, method arguments are guessed when a method is selected from as list of code assist proposals.
 * `java.completion.filteredTypes`: Defines the type filters. All types whose fully qualified name matches the selected filter strings will be ignored in content assist or quick fix proposals and when organizing imports. For example 'java.awt.*' will hide all types from the awt packages.
 * `java.completion.favoriteStaticMembers` : Defines a list of static members or types with static members.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -135,7 +135,7 @@ The following settings are supported:
 * `java.autobuild.enabled` : Enable/disable the 'auto build'.
 * `java.maxConcurrentBuilds`: Set max simultaneous project builds.
 * `java.completion.enabled` : Enable/disable code completion support.
-* `java.completion.overwrite` : When set to true, code completion overwrites the current text. When set to false, code is simply added instead.
+* `java.completion.overwrite` : **Deprecated, please use `editor.suggest.insertMode` instead. To make it Java language-specific, run the global command `Preferences: Configure Language Specific Settings` from the `Command Palette`.** When set to true, code completion overwrites the current text. When set to false, code is simply added instead.
```
jdneo, line 138: @rgrunber @testforstephen Do you think we need to add an estimate date(e.g. after 1.x.0) when to remove this setting in the deprecation message?

rgrunber: I don't think so. Maybe we need to come up with some general policy on how many releases, or how long we will commit to supporting a deprecated feature.

The only burden for this feature is keeping the setting description in package.json and `isOverwriteOnCompletion` to check for the old setting.

jdneo: ok

rgrunber:
Note : Looks like we handle the overwrite at https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/contentassist/CompletionProposalReplacementProvider.java#L163 .

How likely would it be for VS Code to provide some kind of "modifier" key that changes the behaviour of the insertMode ? Ie. If we leave completion as overwrite, one can press `Ctrl` just before triggering completion to revert to insert for that completion. This is similar to what Eclipse IDE does. See https://twitter.com/sandipchitale/status/1501772247115071488 .

jdneo:
I guess it's possible.

The LSP has a client capability called `insertReplaceSupport`. What we need to do is resolve both insert text edit and replace text edit for each completion item: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertReplaceEdit

rgrunber:
The results suggest we keep overwrite : https://twitter.com/VSCodeJava/status/1501613413625118721 . So I'll review this patch as-is and maybe we can file a separate issue to consider using `insertReplaceSupport`.

jdneo:
@rgrunber

There is a potential risk here for this patch. My current implementation will pick `editor.suggest.insertMode` if `java.completion.overwrite` is default value (unchanged).

But the default value of `editor.suggest.insertMode` is `insert`, so that means user will observe the behavior changes to `insert` mode.

README.md
```diff
@@ -199,9 +198,13 @@ The following settings are supported:
...
  * `java.import.generatesMetadataFilesAtProjectRoot` : Specify whether the project metadata files(.project, .classpath, .factorypath, .settings/) will be generated at the project root. Defaults to `false`.
+ * `java.inlayHints.parameterNames.enabled`: Enable/disable inlay hints for parameter names. Supported values are: `none`(isable parameter name hints), `literals`(Enable parameter name hints only for literal arguments) and `all`(Enable parameter name hints for literal and non-literal arguments). Defaults to `literals`.
 
-New in 1.3.0:
-* `java.jdt.ls.java.home`: Absolute path to JDK home folder used to launch the Java Language Server. This setting will replace the Java extension's embedded JRE to start the Java Language Server. Requires VS Code restart.
+New in 1.5.0:
+* `java.inlayHints.parameterNames.enabled`: Enable/disable inlay hints for parameter names. Supported values are: `none`(isable parameter name hints), `literals`(Enable parameter name hints only for literal arguments) and `all`(Enable parameter name hints for literal and non-literal arguments). Defaults to `literals`.
```
rgrunber, line 204:
```suggestion
* `java.inlayHints.parameterNames.enabled`: Enable/disable inlay hints for parameter names. Supported values are: `none`(Disable parameter name hints), `literals`(Enable parameter name hints only for literal arguments) and `all`(Enable parameter name hints for literal and non-literal arguments). Defaults to `literals`.
```

rgrunber (approved):
Overall, change looks good, and I think we can merge if intended for upcoming release. (after fixing typo)

Issue #2099, opened 2021-09-06
support inlay hints


Eskibear:
It's available in IntelliJ IDEA,
https://www.jetbrains.com/help/idea/inlay-hints.html

Supported in TS/JS since vscode v1.60 
![image](https://user-images.githubusercontent.com/2351748/132154259-4d47251d-3945-4fec-9ca7-3c8715d150de.png)
See: https://code.visualstudio.com/updates/v1_60#_inlay-hints-for-javascript-and-typescript

Proposed API in VS Code:
https://github.com/microsoft/vscode/blob/997228d528d687fd17cbf3ba117a0d4f668c1393/src/vs/vscode.proposed.d.ts#L1773-L1793

LSP proposal: 
https://github.com/microsoft/language-server-protocol/issues/956

rgrunber:
It's also supported in Eclipse since 4.12 (2019-06) https://www.eclipse.org/eclipse/news/4.12/jdt.php#parameter-name-codemining.

https://git.eclipse.org/c/gerrit/jdt/eclipse.jdt.ui.git/tree/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java

I don't think we can use the class since it's in UI and ICodeMining is tied to some UI rendering elements. However the AST logic is pretty straightforward.

jayy2kok:
I have adopted VS Code for full time java IDE, I love it, just this one feature I am missing.

rgrunber:
I think we should watch out for things like https://bugs.eclipse.org/bugs/show_bug.cgi?id=568844 , and maybe look at possible solution. The jumpiness is caused when a method temporarily can't be resolved to a proper signature (due to syntax error?) and then switches to resolved. The effect is annotations flickering which affect the user's ability to type.

jdneo:
Update: inlay hint is available as client side API in VS Code 1.65.0.

Maybe we can first make a client side implementation. I can volunteer for that. :)

PR #2354, opened 2022-03-11
Support inlay hints for parameter names

jdneo:
fix https://github.com/redhat-developer/vscode-java/issues/2099

requires https://github.com/eclipse/eclipse.jdt.ls/pull/2019

### DIFF
--- a/src/inlayHintsProvider.ts
+++ b/src/inlayHintsProvider.ts
@@ -1,0 +1,95 @@
+import { CancellationToken, EventEmitter, InlayHint, InlayHintKind, InlayHintsProvider, Range, TextDocument } from "vscode";
+import * as ls from 'vscode-languageserver-protocol';
+import { LanguageClient, RequestType } from "vscode-languageclient/node";
+
+export class JavaInlayHintsProvider implements InlayHintsProvider {
+
+    private onDidChange = new EventEmitter<void>();
+    public onDidChangeInlayHints = this.onDidChange.event;
+
+    constructor(private client: LanguageClient) {
+        this.client.onRequest(InlayHintRefreshRequest.type, async () => {
+            this.onDidChange.fire();
+        });
+    }
+
+    public async provideInlayHints(document: TextDocument, range: Range, token: CancellationToken): Promise<InlayHint[]> {
+        const requestParams: InlayHintParams = {
+            textDocument: this.client.code2ProtocolConverter.asTextDocumentIdentifier(document),
+            range: this.client.code2ProtocolConverter.asRange(range)
+        };
+        try {
+            const values = await this.client.sendRequest(InlayHintRequest.type, requestParams, token);
+            if (token.isCancellationRequested) {
+                return [];
+            }
+            return asInlayHints(values, this.client);
+        } catch (error) {
+            return this.client.handleFailedRequest(InlayHintRequest.type, token, error, []);
+        }
+    }
+}
+
+/**
+ * A parameter literal used in inlay hints requests.
+ *
+ * @since 3.17.0 - proposed state
+ */
+ export type InlayHintParams = /*WorkDoneProgressParams &*/ {
+    /**
+     * The text document.
+     */
+    textDocument: ls.TextDocumentIdentifier;
+
+    /**
+     * The document range for which inlay hints should be computed.
+     */
+    range: ls.Range;
+};
+
+/**
+ * Inlay hint information.
+ *
+ * @since 3.17.0 - proposed state
+ */
+export type LSInlayHint = {
+
+    /**
+     * The position of this hint.
+     */
+    position: ls.Position;
+
+    /**
+     * The label of this hint. A human readable string or an array of
+     * InlayHintLabelPart label parts.
+     *
+     * *Note* that neither the string nor the label part can be empty.
+     */
+    label: string; // label: string | InlayHintLabelPart[];
+};
+
+namespace InlayHintRequest {
+    export const type: RequestType<InlayHintParams, LSInlayHint[], any> = new RequestType('textDocument/inlayHint');
+}
+
+/**
+ * @since 3.17.0 - proposed state
+ */
+namespace InlayHintRefreshRequest {
+    export const type: RequestType<void, void, void> = new RequestType('workspace/inlayHint/refresh');
+}
+
+async function asInlayHints(values: LSInlayHint[] | undefined | null, client: LanguageClient, ): Promise<InlayHint[] | undefined> {
+    if (!Array.isArray(values)) {
+        return undefined;
+    }
+    return values.map(lsHint => asInlayHint(lsHint, client));
+}
+
+function asInlayHint(value: LSInlayHint, client: LanguageClient): InlayHint {
+    const label = value.label;
+    const result = new InlayHint(client.protocol2CodeConverter.asPosition(value.position), label);
+    result.paddingRight = true;
+    result.kind = InlayHintKind.Parameter;
+    return result;
+}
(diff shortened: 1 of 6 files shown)


src/inlayHintsProvider.ts
```diff
@@ -0,0 +1,37 @@
...
+                this.onDidChange.fire();
+            }
+        }));
+    }
+
+    public async provideInlayHints(document: TextDocument, range: Range, token: CancellationToken): Promise<InlayHint[]> {
+        const start = performance.now();
+        const hints = await commands.executeCommand<InlayHint[]>(Commands.EXECUTE_WORKSPACE_COMMAND,
+            Commands.JAVA_EDIT_INLAY_HINTS,
+            document.uri.toString(),
```
angelozerr, line 22: IMHO, I think the command should use an InlayHint structure: 

```ts
export interface InlayHint {
		/**
		 * Optional {@link TextEdit text edits} that are performed when accepting this inlay hint. The default
		 * gesture for accepting an inlay hint is the double click.
		 *
		 * *Note* that edits are expected to change the document so that the inlay hint (or its nearest variant) is
		 * now part of the document and the inlay hint itself is now obsolete.
		 */
		textEdits?: TextEdit[];
	}
```

like LSP spec provides

https://github.com/microsoft/vscode-languageserver-node/blob/37cf23b995720b53a19553c60c15a6fb04536a59/client/typings/vscode.proposed.d.ts#L6

It will help more when you will move to standard LSP.

jdneo: Sorry I didn't quite get the idea. Do you mean adding one more field `textEdits`?

angelozerr: Sorry with my mistake, I wanted to say use an `inlayHintParams` https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHintParams

src/standardLanguageClient.ts
```diff
@@ -498,6 +499,14 @@ export class StandardLanguageClient {
 				scheme: "file",
 				pattern: "**/pom.xml"
 			}, new PomCodeActionProvider(context), pomCodeActionMetadata);
+
+			if (languages.registerInlayHintsProvider) {
```
rgrunber, line 503: Since Theia doesn't support this yet : https://eclipse-theia.github.io/vscode-theia-comparator/status.html, I think we may have to us something like `(languages as any).registerInlayHintsProvider` here. See https://github.com/redhat-developer/vscode-java/pull/1975/files .

jdneo: My understand is that: the `languages` namespace should be available in Theia. If `registerInlayHintsProvider` is not supported in Theia, `languages.registerInlayHintsProvider` should be `undefined` so the if -block is skipped. Looks like the current check is enough, right?

jdneo: Ok, I guess I got your idea. Does that mean we should keep the `@types/vscode` version and the `vscode engine` version unchanged. And use `(xxx as any).xxx()` to add those features?

src/inlayHintsProvider.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Commands } from "./commands";
+
+export class JavaInlayHintsProvider implements InlayHintsProvider {
+
+    private onDidChange = new EventEmitter<void>();
+    public onDidChangeInlayHints = this.onDidChange.event;
+
+    constructor(context: ExtensionContext) {
+        context.subscriptions.push(workspace.onDidChangeConfiguration(e => {
+            if (e.affectsConfiguration("java.inlayHints.parameterNames.enabled")) {
```
angelozerr, line 12: IMHO, I think the correct LSP behavior is that LSP track the changed of this inlay hint settings and send  the [workspace/inlayHint/refresh)(https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_inlayHint_refresh).

It will give the capability to refresh inlay hint for any LSP client which implement inlay hint (not only vscode-java).

I implemented this LSP request 

 * on client side at https://github.com/redhat-developer/vscode-quarkus/pull/474/files#diff-aeb59384a9e0b18adc89b06d548c119ed02ead2a48e1c11a26eb4972409dfe27R61
 * on server side at https://github.com/redhat-developer/quarkus-ls/pull/593/files#diff-854c916bc4ae147f987affa94adc4054d62afb789c45d2778a4d2b18164af4dfR163 with https://github.com/redhat-developer/quarkus-ls/pull/593/files#diff-0d2b9d01f2e24f7428e923402711d6b5699b159308d6a8acb14703ac5a5c6bcdR32

I repeat me again, but IMHO I think it's better to use custom LSP request in server side instead of using delegate command handler because:

 * it will be very easy to move to LSP specification on vscode client (just remove the TypeScript code which implement InlayHintsProvider. vscode-languageclient already implemented that https://github.com/microsoft/vscode-languageserver-node/blob/e245f790021a7439f9c54c22ba8d404ceaf14aa5/client/src/common/inlayHint.ts#L120 We must just wait for a new release of vscode-languageclient which will come soon https://github.com/microsoft/vscode-languageserver-node/issues/906#issuecomment-1065117201
 * it will be easy to move to LSP specification on JDT LS server side, when LSP4J will support InlayHint https://github.com/eclipse/lsp4j/issues/570. It's just again a matter of removing LSP java structure.

rgrunber:
I need to :

- [ ] Test this against Theia
- [ ] Play around with the change and see how it behaves
- [ ] Look up some of our pain points on the Eclipse IDE side and see if there's more issues to be aware of (eg. hints flickering)

angelozerr:
> Look up some of our pain points on the Eclipse IDE side and see if there's more issues to be aware of (eg. hints flickering)

You will not able to do that, since LSP4E doesn't support InlayHint because LSP4J doesn't provide InlayHint https://github.com/eclipse/lsp4j/issues/570

src/inlayHintsProvider.ts
jdneo: Question: how do you deal with the client side capability for inlay hint?

angelozerr: My PR consider that server (Qute server) and client (vscode) can support inlay hint, but I agree with you my work must be by using with https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHintClientCapabilities

jdneo: Are you planning to support it at the current stage? Looks like it will become complicated because `TextDocumentClientCapabilities` and `ServerCapabilities` already exists in LSP4J but inlay hint part is missed.

angelozerr: > Are you planning to support it at the current stage?

 * I don"t see the advantage to support TextDocumentClientCapabilities .
 * supporting ServerCapabilities become important when vscode-languageclient will be released on April which will provide Inlay Hint support. My idea is to create an ExtendedServerCapabilities  which extends ServerCapabilities  to update the inlayHintProvider of teh server capability https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint

src/standardLanguageClient.ts
rgrunber: Looks like this code works just fine in Theia. No need for complicating the check. The extension installs just fine even with `@types/vscode` set the way it currently is.

jdneo:
> This is probably more to do with the OS / Desktop Environment but just wondering if others have seen this behaviour.

I can also observe this behavior in VS Code running on Windows. Might be related with VS Code's viewport calculation strategy? Or maybe a bug in VS Code.

Issue #2319, opened 2022-02-14
Command "open Java extension log" doesn't open latest log file


Eskibear:
It always open file `clien.log.xxxx.9` when I have more than 10 files in that folder. Seems like wrong comparison in alphabetical order.

![image](https://user-images.githubusercontent.com/2351748/153824407-8f1264b7-2ab7-44e4-9929-f1ffa013534a.png)

PR #2356, opened 2022-03-14
change max size of client.log

Eskibear:
improve the case mentioned in https://github.com/redhat-developer/vscode-java/issues/2319

With `100k` as max size, sometimes it creates several log files for a single session, which is not convenient for diagnose. E.g. I have to try opening several log files to find the the one containing the command to start Java language server. 

This PR simply increases it to `1m`, in order to mitigate the inconvenience.

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -5,7 +5,7 @@ export function initializeLogFile(filename: string) {
 	logger.add(new DailyRotateFile({
 		filename: filename,
 		datePattern: 'YYYY-MM-DD',
-		maxSize: '100k', // 100k max size per file
+		maxSize: '1m', // 1MB max size per file
 		maxFiles: '2d' // retain logs of the last two days
 	}));
 }


Eskibear:
BTW, I also considered to create client.log per session, then it would be much clearer to inspect without comparing timestamps. 
But seems there's no simple option to set in `winston-daily-rotate-file`. So I choose this cheap way to mitigate.

src/log.ts
```diff
@@ -5,7 +5,7 @@ export function initializeLogFile(filename: string) {
 	logger.add(new DailyRotateFile({
 		filename: filename,
 		datePattern: 'YYYY-MM-DD',
-		maxSize: '100k', // 100k max size per file
+		maxSize: '1m', // max size per file
```
rgrunber, line 8: For convenience, I would say `1MB max size per file` as the comment, so it's clear.

rgrunber (commented):
Looks good to me.

I've run into this as well, when I first started working on the project, but I found it was easier to just look up the logs in the console, since that doesn't get truncated. With that said, this is mainly a problem for the verbose logging, but definitely a great improvement for diagnosing problems.

Only disadvantage I can think of is users would now be uploading a log file that in the worst case is 10x larger, although the current system does risk losing relevant information.

src/log.ts
Eskibear: updated.

Eskibear:
> ...uploading a log file that in the worst case is 10x larger

Exactly. So ideally, how about we rotate client.log per session. 
Advantage:
- easy to look up logs
- usually small file size

Disadvantage:
- creates lots of files if you frequently reload window, but that's acceptable if we combine the strategy to only keep logs for 2 days.
- extra effort on implementation of rotation, and update the way to open corresponding client.log

rgrunber (approved):
For the time being, I'm fine with the current implementation. As you mention, we already rotate the logs every 2 days.

PR #2371, opened 2022-03-22
Support publishing pre-release versions

### LINKED ISSUE #2285: Support publishing pre-release versions
- The client UX for pre-release versions - https://code.visualstudio.com/updates/v1_63#_pre-release-extensions 
- How to publish pre-release versions - https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions

CsCherrYY:
fix #2285 

Notes:
- For the trigger part, I'm not sure the current trigger of Jenkins pipeline since it's on the Jenkins server, I suppose we trigger this pipeline after there is a commit to the vscode-java or eclipse.jdt.ls repository, so in this PR I add a trigger in the scripted pipeline to nightly build when there is SCM changes, if we can merge the  configuration with that in the current Jenkins server that would be fine. @rgrunber any comments?
- Due to the requirement of pre-release channel (see https://github.com/redhat-developer/vscode-java/issues/2285#issuecomment-1072075843), a tool node 14.x should be added to the Jenkins server (in this PR it's nodejs-14.19.1) and we will use the newest version of `vsce`.
- Validation: for the pre-release versions, the packaged vsix file has an 'extension.manifest' file in its root path, and in that file, we can find 
> <Property` Id="Microsoft.VisualStudio.Code.PreRelease" Value="true" />


CsCherrYY:
test this please

CsCherrYY:
Sounds great. In this way, we will not make any SCM trigger change in this PR but it will happen on the server side, is that right? if so, I can remove this part then.

CsCherrYY:
@rgrunber I have removed the SCM trigger part in the Jenkins file.

For the node version and vsce version part, is there any gap in using node 14.x for publishing now?

scripts/prepare-nightly-build.js
```diff
@@ -0,0 +1,25 @@
...
+const month = date.getMonth() + 1;
+const day = date.getDate();
+const hours = date.getHours();
+patch = `${date.getFullYear()}${prependZero(month)}${prependZero(day)}${prependZero(hours)}`;
+
+const insiderPackageJson = Object.assign(json, {
+    version: `${major}.${minor}.${patch}`,
+});
+
+fs.writeFileSync("./package.insiders.json", JSON.stringify(insiderPackageJson));
```
testforstephen, line 25: Use `JSON.stringify(insiderPackageJson, null, 2)` to preserve the indentation.

Jenkinsfile
```diff
@@ -44,10 +60,14 @@ node('rhel8'){
 	sh "mkdir ./server"
 	sh "tar -xvzf ${files[0].path} -C ./server"
 
+	if (!publishToMarketPlace.equals('true')) {
```
testforstephen, line 63: In current Jenkinsfile, every commit will be published to stable or prerelease, which is not what we want.

I suggest adding a new build parameter like `publishPreRelease` to control whether we should package the vsix with `--pre-release` tag.

CsCherrYY: This approach (add a new parameter) makes sense, but I'm not sure if we can specify a parameter in the automatic trigger in Jenkins server. In other words, we should make the build of automatic trigger and manual trigger different parameters (`publishPreRelease`), to distinguish

- SCM trigger, with `publishPreRelease` == `true` and `publishToMarketPlace` == `true`
- manual CI, with `publishPreRelease`  == `false` and `publishToMarketPlace` == `false`
- manual release, with `publishPreRelease`  == `false` and `publishToMarketPlace` == `true`

CsCherrYY: BTW, if we can change the trigger strategy to daily build, like https://github.com/redhat-developer/vscode-java/pull/2371#issuecomment-1076509899, we will not publish a version on each commit, but on each day if new commit comes.

testforstephen: With this commit, current JBoss publish won't work any more. Are you planning to retire JBoss immediately?

testforstephen: > This approach (add a new parameter) makes sense, but I'm not sure if we can specify a parameter in the automatic trigger in Jenkins server. In other words, we should make the build of automatic trigger and manual trigger different parameters (`publishPreRelease`), to distinguish
> 
> * SCM trigger, with `publishPreRelease` == `true` and `publishToMarketPlace` == `true`
> * manual CI, with `publishPreRelease`  == `false` and `publishToMarketPlace` == `false`
> * manual release, with `publishPreRelease`  == `false` and `publishToMarketPlace` == `true`

This is not a technical issue. In Jenkins, there are many ways to pass variables, such as environment variables and build parameters.

CsCherrYY: get it, I'll keep the current build work and add a new build logic to publish a pre-release with a new environment variable.

CsCherrYY:
The newest commit I just separate the pre-release pipeline and keep the current builds work. All we should do in the Jenkins server are:

- Create a new pipeline, and target this repository
- Set the SCM trigger to daily build if new commit comes
- Copy the environment variables from the existing pipeline, without `publishToMarketPlace`. Instead, create a new variable `publishPreRelease` and set it to true.

Jenkinsfile
```diff
@@ -1,17 +1,73 @@
...
+def publishPreReleaseExtensions() {
+	stage "replace extension version"
+	sh "node ./scripts/prepare-nightly-build.js"
+	sh "mv ./package.insiders.json ./package.json"
+	packageGenericExtension()
+	def vsix = findFiles(glob: '**.vsix')
+
+	stage "publish generic version"
+	withCredentials([[$class: 'StringBinding', credentialsId: 'vscode_java_marketplace', variable: 'TOKEN']]) {
+		sh 'vsce publish -p ${TOKEN} --target win32-ia32 win32-arm64 linux-armhf alpine-x64 alpine-arm64'
```
testforstephen, line 57: When `vsce publish` command doesn't specify the vsix path, actually it will build a temp vsix from current src directory and then publish it to marketplace. This means this step will publish to the stable channel instead of pre-release.

Jenkinsfile
```diff
@@ -1,17 +1,73 @@
...
+	}
+
+	stage "publish specific version"
+	packageSpecificExtensions()
+	withCredentials([[$class: 'StringBinding', credentialsId: 'vscode_java_marketplace', variable: 'TOKEN']]) {
+		sh 'npx gulp clean_jre'
```
testforstephen, line 63: Originally this cleanup is used for publishing generic, because the generic publish step will repackage the vsix. It's not needed for platform vsix when using --packagePath in publish command.

Jenkinsfile
```diff
@@ -1,17 +1,73 @@
...
+		def platformVsixes = findFiles(glob: '**.vsix', excludes: vsix[0].path)
```
testforstephen, line 64: This will find all vsixes including the generic version so as to publish generic version twice.

Jenkinsfile
CsCherrYY: Let's use `vsce publish --pre-release` here and skip the steps to package generic vsix when prerelease.

Jenkinsfile
CsCherrYY: Sure, Let's remove it.

Jenkinsfile
```diff
@@ -1,17 +1,55 @@
...
+	}
+}
+
+def publishPreReleaseExtensions() {
+	stage "replace extension version"
+	sh "node ./scripts/prepare-nightly-build.js"
+	sh "mv ./package.insiders.json ./package.json"
+
+	def packageJson = readJSON file: 'package.json'
+	env.EXTENSION_VERSION = "${packageJson.version}"
```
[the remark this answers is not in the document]
CsCherrYY: used in `packageSpecificExtensions()`

Jenkinsfile
```diff
@@ -1,17 +1,55 @@
...
+	stage "publish generic version"
+	withCredentials([[$class: 'StringBinding', credentialsId: 'vscode_java_marketplace', variable: 'TOKEN']]) {
+		sh 'vsce publish --pre-release -p ${TOKEN} --target win32-ia32 win32-arm64 linux-armhf alpine-x64 alpine-arm64'
+	}
+	def vsix = findFiles(glob: '**.vsix')
+
+	stage "publish specific version"
+	packageSpecificExtensions()
+	withCredentials([[$class: 'StringBinding', credentialsId: 'vscode_java_marketplace', variable: 'TOKEN']]) {
+		def platformVsixes = findFiles(glob: '**.vsix', excludes: vsix[0].path)
```
testforstephen, line 46: with the new logic, vsix[0].path doesn't exist any more, won't it report errors?

CsCherrYY: Here I assign the variable in line 41 for the generic vsix file.
```
def vsix = findFiles(glob: '**.vsix')
```

testforstephen (approved):
LGTM

testforstephen:
@rgrunber there are two follow-ups for you:

- Install nodejs-14.19.1 in CI machine.
- Create a new pipeline for pre-release, see more details from @CsCherrYY comment above.

  > The newest commit I just separate the pre-release pipeline and keep the current builds work. All we should do in the Jenkins server are:
  > 
  > * Create a new pipeline, and target this repository
  > * Set the SCM trigger to daily build if new commit comes
  > * Copy the environment variables and credentials from the existing pipeline, without `publishToMarketPlace`. Instead, create a new variable `publishPreRelease` and set it to true.

CsCherrYY:
@rgrunber Could you please verify this change in the Jenkins server? If there is anything I can help, feel free to reach out🙂

scripts/prepare-nightly-build.js
```diff
@@ -0,0 +1,25 @@
+const fs = require("fs");
```
rgrunber, line 1: I guess the script is fine for now. Initially I thought maybe you could use https://www.jenkins.io/doc/pipeline/steps/pipeline-utility-steps/#readjson-read-json-from-files-in-the-workspace , but with all the other manipulations it makes sense. Also, it's nice to run it locally.

I was thinking maybe it would be nice to have it in gulpfile.js so we have a common area for scripts (as opposed to gulpfile.js & scripts/ ) but this isn't the end of the world.

rgrunber (changes requested):
- The build seems to work as expected with `nodejs-14.19.1` :rocket: . I can confirm presence of `<Property Id="Microsoft.VisualStudio.Code.PreRelease" Value="true" />`
- I would also modify the `ovsx@<0.3.0` reference to just `ovsx` . Same situation as vsce ( https://github.com/eclipse/openvsx/commit/f5cc1a9e760039c18bf423b03b78ac620af56ae0 )
- I just need to create a PR on our end to add `publishPreRelease` (default `true`) as a parameter & to see how we can modify the job so it runs daily.

Jenkinsfile
```diff
@@ -44,39 +81,37 @@ node('rhel8'){
...
+		sh "vsce package -o java-${env.EXTENSION_VERSION}-${env.BUILD_NUMBER}.vsix"
 
-	def vsix = findFiles(glob: '**.vsix')
-	stash name:'vsix', includes:vsix[0].path
+		stage 'Test vscode-java for staging'
+		wrap([$class: 'Xvnc']) {
+			sh "npm run compile" //compile the test code too
+			env.SKIP_COMMANDS_TEST="true"
+			sh "npm test --silent"
+		}
```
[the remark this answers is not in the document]
CsCherrYY: Makes sense. Let's move this stage right after we extract server jars into server folder.

Jenkinsfile
```diff
@@ -1,17 +1,54 @@
...
+		sh "npx gulp download_jre --target ${platform} --javaVersion ${embeddedJRE}"
+		if (publishPreRelease.equals('true')) {
+			sh "vsce package --pre-release --target ${platform} -o java-${platform}-${env.EXTENSION_VERSION}-${env.BUILD_NUMBER}.vsix"
+		} else {
+			sh "vsce package --target ${platform} -o java-${platform}-${env.EXTENSION_VERSION}-${env.BUILD_NUMBER}.vsix"
+		}
+	}
+}
+
+def publishPreReleaseExtensions() {
```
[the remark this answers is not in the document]
CsCherrYY: Yeah, here I just used a confused name. Let's divide the package step and the publish step into separate functions, in which the related variables like `publishPreRelease` or `publishToMarketPlace` are used.

scripts/prepare-nightly-build.js
CsCherrYY: 👍 Let's put it in `gulpfile.js`.

rgrunber (approved):
Just one minor thing to fix, but overall, works.

- I think you re-introduced the `"ovsx@<0.3.0"` line. Just make it `"ovsx"`. With node 14 we don't need to restrict it.

Just as a note, the Jenkins job has been configured to use an `scmpoll_spec` of `'0 8 * * *'` so it will build once per day only if there are new changes. We can play with this later if necessary.

CsCherrYY:
@rgrunber I have removed the version striction of ovsx. We can start to play with it!

PR #2375, opened 2022-03-24
Add commands to view Java Language Server debug and error logs

### LINKED ISSUE #2367: m2e logs
I have a Maven build process getting into an infinite loop state, where it's constantly building (I can see it repeatedly spawning processes that should be started once during the build). I would like to see the m2e log, like I did in Eclipse, to try to get a handle on what's going on, but I don't seem to be able to see any output from m2e. Am I missing something or is this something that we could add?

karlvr:
Resolves #2367, which in spite of being closed wasn't resolved as such (we found out how to view the logs but didn't make it easy)

### DIFF
--- a/package.json
+++ b/package.json
@@ -924,13 +924,19 @@
       },
       {
         "command": "java.open.serverLog",
-        "title": "%java.open.serverLog%",
-        "category": "Java"
+        "when": "false"
+      },
+      {
+        "command": "java.open.serverStdoutLog",
+        "when": "false"
+      },
+      {
+        "command": "java.open.serverStderrLog",
+        "when": "false"
       },
       {
         "command": "java.open.clientLog",
-        "title": "%java.open.clientLog%",
-        "category": "Java"
+        "when": "false"
       },
       {
         "command": "java.open.logs",
--- a/package.nls.json
+++ b/package.nls.json
@@ -2,8 +2,6 @@
 	"java.server.mode.switch": "Switch to Standard Mode",
 	"java.projectConfiguration.update": "Update Project",
 	"java.project.import": "Import Java Projects into Workspace",
-	"java.open.serverLog": "Open Java Language Server Log File",
-	"java.open.clientLog": "Open Java Extension Log File",
 	"java.open.logs": "Open All Log Files",
 	"java.workspace.compile": "Force Java Compilation",
 	"java.open.formatter.settings": "Open Java Formatter Settings",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -369,6 +369,8 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 
 			// Register commands here to make it available even when the language client fails
 			context.subscriptions.push(commands.registerCommand(Commands.OPEN_SERVER_LOG, (column: ViewColumn) => openServerLogFile(workspacePath, column)));
+			context.subscriptions.push(commands.registerCommand(Commands.OPEN_SERVER_STDOUT_LOG, (column: ViewColumn) => openRollingServerLogFile(workspacePath, '.out-jdt.ls', column)));
+			context.subscriptions.push(commands.registerCommand(Commands.OPEN_SERVER_STDERR_LOG, (column: ViewColumn) => openRollingServerLogFile(workspacePath, '.error-jdt.ls', column)));
 
 			context.subscriptions.push(commands.registerCommand(Commands.OPEN_CLIENT_LOG, (column: ViewColumn) => openClientLogFile(clientLogFile, column)));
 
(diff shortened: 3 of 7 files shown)


src/extension.ts
```diff
@@ -720,6 +722,25 @@ function openServerLogFile(workspacePath, column: ViewColumn = ViewColumn.Active
...
 }
 
+function openRollingServerLogFile(workspacePath, filename, column: ViewColumn = ViewColumn.Active): Thenable<boolean> {
+	return new Promise((resolve) => {
+		const dirname = path.join(workspacePath, '.metadata');
+
+		// find out the newest one
+		glob(filename + '-*', { cwd: dirname }, (err, files) => {
+			if (!err && files.length > 0) {
+				files.sort();
```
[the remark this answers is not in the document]
karlvr: @rgrunber the file names are like `.out-jdt.ls-YYYYMMDDHHmmss.log` so I `sort`ed and ran. I guess the filenames are designed to make that the easy approach so I hope it's okay to rely upon!

package.nls.json
```diff
@@ -3,6 +3,8 @@
 	"java.projectConfiguration.update": "Update Project",
 	"java.project.import": "Import Java Projects into Workspace",
 	"java.open.serverLog": "Open Java Language Server Log File",
+	"java.open.serverDebugLog": "Open Java Language Server Debug Log File",
+	"java.open.serverErrorLog": "Open Java Language Server Error Log File",
```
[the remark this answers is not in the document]
karlvr: @rgrunber Thank you very much for this feedback. I've pushed a correction for this now.

What do you think about alerting with instructions to set `jdt.ls.debug` to `true` if those files can't be found? I'll experiment a little with that.

testforstephen:
I'm thinking who is the audience for these logs. To me, extension developers might be interested in these logs, but normal users might not be. If this is the case, I'd rather not expose too many log commands in the Command Palette, and instead reuse the existing command `Java: Open All Log Files`. 

We can refactor this to `Java: Open All Log Files...`. And when you click on it, it will list all possible log locations for user to pick, including:
- Java Extension Client Log File
- Java Language Server Log File
- Standard Output Log File
- Standard Error Log File

@karlvr @rgrunber  WDYT?

karlvr:
@testforstephen I started down this path because I wanted to see the logs from Maven; effectively the Maven Console I missed from Eclipse.

@rgrunber directed me to the files, including how to activate them, in #2367.

The contents of the stdout (I think) is actually what I _expected_ to see in the Maven output window in VS Code… whereas at least for me that output only ever consists of one line identifying my local repository. Given how much Maven does, seeing its output is actually something I'm regularly interested in, especially when I'm pondering how long it's taking and how much processing power it's using!

I do agree that the approach I've used in this PR is a proliferation of options.

Perhaps it would be possible to channel items from at least one of these logs into the Maven Output window, so it's directly accessible?

rgrunber:
- I did also feel a bit like adding 2 more commands for logs might cause more confusion/clutter. So how about we remove the code that exposes the 2 additional commands, and simply have `openLogs()` call `openRollingServerLogFile(..)` twice (once for `.out-jdt.ls` & `.err-jdt.ls`). That should still improve things for now.
- What do you mean by "Maven Output window" ? I don't think we provide such a view. Is it coming from vscode-maven ? It would be different than the "Maven" output/error we produce.
- To your question of whether we could redirect the m2e stdout/stderr to somehwere else, I was thinking about this as well! At the end of the day, the only reason those 2 files exist, is because we defined them at https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/logback.xml. But that was just the easiest thing to do at the time. There's plenty of other "Appenders" we could use. See https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.logback.configuration/defaultLogbackConfiguration/logback.xml as an example. The interesting one is [`EclipseLogAppender`](https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.logback.appender/src/org/eclipse/m2e/logback/appender/EclipseLogAppender.java). It looks like it's logging to the Eclipse Platform log, so this might actually be the way to redirect everything into the JDT-LS server logs! The slight downside.. we need to add `org.eclipse.m2e.logback.appender` (~12kb) to our JDT-LS runtime.

karlvr:
@rgrunber sorry, too many plugins! The Maven Output is from the Maven plugin. It has an entry in the Output tab, but it's underused. It's always where I imagined the Maven log would appear. I discovered when I started my investigation there that the Maven plugin didn't control m2e but forgot!

If we can't extract just the m2e bits then getting the logs into the Java Language Server Output would be great.

rgrunber:
Going through trying to get the EclipseLogAppender working made me discover we have an issue with custom logback configs : https://github.com/eclipse/eclipse.jdt.ls/pull/2077 , so the discussion is definitely appreciated.

I was able to get some of the events to be logged to the language server `.log` file, but not all. So until we get it working, I'm fine with accepting this. Just need to make the following change :

- Do not register the stdout/stderr commands in the command palette. They should exist so that `Open All Log Files` calls them but for now let's not expose them beyond this.

rgrunber:
I'll be pushing this in since @karlvr did some good work to get this ready. I've used `preview: false` from [`TextDocumentOptions`](https://code.visualstudio.com/api/references/vscode-api#TextDocumentShowOptions) in order to allow the log file tabs to open under the same view column. This makes them a lot easier to read than opening each one in its own column.

The Maven related logging will show up under the server logs once the logback tracing PR is merged. It's still good to be able to view the stdout/stderr logs. Further improvements might involve adding a quick-pick menu to select which (of the 4) logs.

rgrunber (approved):

package.json
```diff
@@ -924,13 +924,19 @@
       },
       {
         "command": "java.open.serverLog",
-        "title": "%java.open.serverLog%",
-        "category": "Java"
+        "when": "false"
```
testforstephen, line 927: Oh, you hide this command. "Open Java Language Server Log File" is the most frequently command I use for troubleshooting.

You need to update wiki as well.
https://github.com/redhat-developer/vscode-java/wiki/Troubleshooting

rgrunber: I"m open to restoring it, if you feel the convenience of not having the other 3 log files show up is worth it.

testforstephen: Yes, I'd like to keep server log command.

jdneo: The current change breaks the VS Code's `package.json` schema:
- title is a mandatory field here.
- `when` is not a valid field here, should be used under `menus` section.

A warning can be seen after startup:

image

testforstephen: Agreed. For those commands that are taken as APIs, we don't need to declare it in package.json.

PR #2376, opened 2022-03-24
Move Type Hierarchy implementation to LSP

CsCherrYY:
Currently vscode-java supports type hierarchy via serval delegate commands. Now the support of type hierarchy has been available in serval places:

- LSP: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareTypeHierarchy
- language client: https://github.com/microsoft/vscode-languageserver-node/pull/779
- vscode: https://github.com/microsoft/vscode/pull/128667
- vscode-reference-view: https://github.com/microsoft/vscode-references-view/pull/71

Based on these changes, we can move the implementation to LSP based.

Issue #1928, opened 2021-05-04
Syntax server no longer exits once JDT server is initialized


rgrunber:
Using vscode-java 0.79.0 or newer, and running `watch -n1 -d 'jps -l | grep -v jps'` in some terminal, I noticed that the syntax server no longer exits once the language server has completed its startup.

##### Environment
Linux (Fedora 33)
VS Code 1.55.2

This seems to have started happening with https://github.com/redhat-developer/vscode-java/pull/1894 ( 9f32875a67352487f5c414bb7fef04c9b00af89d ) , but I don't think we do anything with the exit logic. Is there a chance this is a bug in that library ?

- shutdown() is sent by client, and syntax server responds by saving the workspace
- client confirms it received a response from shutdown
- client then sends an exit() request to server
- the SyntaxLanguageServer.exit() on the server side never gets called by the client

Instead, all I see is the following show up in the errors :

```
rejected promise not handled within 1 second: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
stack trace: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
    at writeAfterEnd (_stream_writable.js:266:14)
    at Socket.write (_stream_writable.js:315:5)
    at /home/rgrunber/git/vscode-java/dist/extension.js:2:882887
    at new Promise (<anonymous>)
    at c.write (/home/rgrunber/git/vscode-java/dist/extension.js:2:882805)
    at _.doWrite (/home/rgrunber/git/vscode-java/dist/extension.js:2:872391)
    at /home/rgrunber/git/vscode-java/dist/extension.js:2:872286
```

Debugging this I see it happens for the following data :
`'Content-Length: 33\r\n\r\n'`
`{"jsonrpc":"2.0","method":"exit"}`

rgrunber:
Looks to be an upstream issue https://github.com/microsoft/vscode-languageserver-node/issues/723 .

rgrunber:
As soon as the upstream issue is resolved (scheduled for 3.16.1?) we can remove this option. This should work around the issue until then.

We're basically calling exit() as part of shutdown() if the client sets this property. I've tested this in lightweight and hybrid modes, and it seems to behave correctly.

PR #2377, opened 2022-03-24
Update language-client to 8.1.0

CsCherrYY:
This PR is a part of #2376, to reduce the complexity of that PR. Type Hierarchy support has been introduced since language-client 8.0.0-next.4 so we should update this dependency, the language-client version requires vscode 1.61.0 for compiling, so I update this dependency as well.

- for notification part: see https://github.com/microsoft/vscode-languageserver-node#3170-next-protocol-800-next-json-rpc-800-next-client-and-800-next-server, now all `sendNotification` methods will return a promise.
 
This update is incompatible with current Theia. see: https://github.com/eclipse-theia/theia/issues/10993

### DIFF
--- a/package.json
+++ b/package.json
@@ -1397,7 +1397,7 @@
     "@types/glob": "5.0.30",
     "@types/lodash.findindex": "^4.6.6",
     "@types/mocha": "^5.2.5",
-    "@types/node": "^8.10.51",
+    "@types/node": "^16.11.7",
     "@types/semver": "^7.3.8",
     "@types/sinon": "^10.0.12",
     "@types/vscode": "^1.74.0",
@@ -1418,7 +1418,7 @@
     "request": "^2.88.2",
     "sinon": "^14.0.0",
     "ts-loader": "^9.2.6",
-    "typescript": "^4.2.4",
+    "typescript": "^4.6.4",
     "webpack": "^5.28.0",
     "webpack-cli": "^4.6.0"
   },
--- a/package.json
+++ b/package.json
@@ -1431,7 +1431,7 @@
     "htmlparser2": "6.0.1",
     "jdk-utils": "^0.4.4",
     "semver": "^7.3.5",
-    "vscode-languageclient": "8.0.2",
+    "vscode-languageclient": "8.1.0",
     "winreg-utf8": "^0.1.1",
     "winston": "^3.2.1",
     "winston-daily-rotate-file": "^3.10.0"
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "virtualWorkspaces": false
   },
   "engines": {
-    "vscode": "^1.65.0"
+    "vscode": "^1.67.0"
   },
   "repository": {
     "type": "git",
@@ -1400,7 +1400,7 @@
     "@types/node": "^8.10.51",
     "@types/semver": "^7.3.8",
     "@types/sinon": "^10.0.12",
-    "@types/vscode": "^1.65.0",
+    "@types/vscode": "^1.67.0",
     "@types/winreg": "^1.2.30",
     "@types/winston": "^2.4.4",
     "@vscode/test-electron": "^2.1.5",
@@ -1431,7 +1431,7 @@
     "htmlparser2": "6.0.1",
     "jdk-utils": "^0.4.4",
     "semver": "^7.3.5",
-    "vscode-languageclient": "7.1.0-next.5",
+    "vscode-languageclient": "8.0.2",
     "winreg-utf8": "^0.1.1",
     "winston": "^3.2.1",
     "winston-daily-rotate-file": "^3.10.0"
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -6,7 +6,7 @@ import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
 import { CodeActionContext, CodeActionTriggerKind, commands, ConfigurationTarget, Diagnostic, env, EventEmitter, ExtensionContext, extensions, IndentAction, InputBoxOptions, languages, RelativePattern, TextDocument, UIKind, Uri, ViewColumn, window, workspace, WorkspaceConfiguration } from 'vscode';
-import { CancellationToken, CodeActionParams, CodeActionRequest, Command, DidChangeConfigurationNotification, ExecuteCommandParams, ExecuteCommandRequest, LanguageClientOptions, RevealOutputChannelOn } from 'vscode-languageclient';
+import { CancellationToken, CodeActionParams, CodeActionRequest, Command, DidChangeConfigurationNotification, ExecuteCommandParams, ExecuteCommandRequest, LanguageClientOptions, RevealOutputChannelOn, State } from 'vscode-languageclient';
 import { LanguageClient } from 'vscode-languageclient/node';
 import { apiManager } from './apiManager';
 import { ClientErrorHandler } from './clientErrorHandler';
@@ -175,8 +175,8 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 				},
 				middleware: {
 					workspace: {
-						didChangeConfiguration: () => {
-							standardClient.getClient().sendNotification(DidChangeConfigurationNotification.type, {
+						didChangeConfiguration: async () => {
+							await standardClient.getClient().sendNotification(DidChangeConfigurationNotification.type, {
 								settings: {
 									java: getJavaConfig(requirements.java_home),
 								}
@@ -185,12 +185,12 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 					},
 					// https://github.com/redhat-developer/vscode-java/issues/2130
 					// include all diagnostics for the current line in the CodeActionContext params for the performance reason
-					provideCodeActions: (document, range, context, token, next) => {
+					provideCodeActions: async (document, range, context, token, next) => {
 						const client: LanguageClient = standardClient.getClient();
 						const params: CodeActionParams = {
 							textDocument: client.code2ProtocolConverter.asTextDocumentIdentifier(document),
 							range: client.code2ProtocolConverter.asRange(range),
-							context: client.code2ProtocolConverter.asCodeActionContext(context)
+							context: await client.code2ProtocolConverter.asCodeActionContext(context)
 						};
 						const showAt  = getJavaConfiguration().get<string>("quickfix.showAt");
 						if (showAt === 'line' && range.start.line === range.end.line && range.start.character === range.end.character) {
@@ -209,12 +209,12 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 								const codeActionContext: CodeActionContext = {
 									diagnostics: allDiagnostics,
 									only: context.only,
-									triggerKind: CodeActionTriggerKind.Invoke,
+									triggerKind: context.triggerKind,
 								};
-								params.context = client.code2ProtocolConverter.asCodeActionContext(codeActionContext);
+								params.context = await client.code2ProtocolConverter.asCodeActionContext(codeActionContext);
 							}
 						}
-						return client.sendRequest(CodeActionRequest.type, params, token).then((values) => {
+						return client.sendRequest(CodeActionRequest.type, params, token).then(async (values) => {
 							if (values === null) {
 								return undefined;
 							}
@@ -224,7 +224,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 									result.push(client.protocol2CodeConverter.asCommand(item));
 								}
 								else {
-									result.push(client.protocol2CodeConverter.asCodeAction(item));
+									result.push(await client.protocol2CodeConverter.asCodeAction(item));
 								}
 							}
 							return result;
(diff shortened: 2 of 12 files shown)


src/extension.ts
```diff
@@ -297,8 +304,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 								}
 							}
 							return result;
-						}, (error) => {
-							client.logFailedRequest(CodeActionRequest.type, error);
```
CsCherrYY, line 301: I can't find this method in both the current version of language-client and 8.0.0 language-client. So there should be another fix?

src/hoverAction.ts
```diff
@@ -75,10 +75,20 @@ class JavaHoverProvider implements HoverProvider {
 
         const contributed = new MarkdownString(contributedCommands.map((command) => this.convertCommandToMarkdown(command)).join(' | '));
         contributed.isTrusted = true;
-        let contents: MarkedString[] = [ contributed ];
+        let contents: MarkdownString[] = [ contributed ];
         let range;
         if (serverHover && serverHover.contents) {
-            contents = contents.concat(serverHover.contents);
+            const serverHoverContentsMarkdown = [];
```
CsCherrYY, line 81: the type `MarkedString` and `MarkdownString` are incompatible now.

src/extension.ts
```diff
@@ -278,6 +284,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 								}
 								const codeActionContext: CodeActionContext = {
 									diagnostics: allDiagnostics,
+									triggerKind: context.triggerKind,
```
CsCherrYY, line 287: `triggerKind` is a required field in VS Code now. Theia doesn't support this field now, see: https://eclipse-theia.github.io/vscode-theia-comparator/status.html but IMO this change will not introduce NPE since we just assign an undefined to a field.

src/extension.ts
Eskibear: I don't see method `logFailedRequest` in LanguageClient. 

It's introduced by https://github.com/redhat-developer/vscode-java/commit/78c82941606c1392f56b76491dc8aba87cd47faa , where client is of type 'any'. I'll invite @snjeza  to explain the details.

src/extension.ts
Eskibear: This LGTM, should be no problem.

src/hoverAction.ts
Eskibear: FYI, `MarkedString` is deprecated in favor of `MarkdownString`. And `vscode.d.ts` changed. 

v1.53
```ts
export type MarkedString = MarkdownString | string | { language: string; value: string };
export class Hover {
		contents: MarkedString[];
```

v1.61
```ts
export type MarkedString = string | { language: string; value: string };
export class Hover {
		contents: Array<MarkdownString | MarkedString>;
```

Eskibear: LGTM as long as we process it properly, making sure no regression in hover

src/standardLanguageClient.ts
```diff
@@ -311,15 +311,15 @@ export class StandardLanguageClient {
 				commands.executeCommand(Commands.SHOW_REFERENCES, Uri.parse(uri), this.languageClient.protocol2CodeConverter.asPosition(position), locations.map(this.languageClient.protocol2CodeConverter.asLocation));
 			}));
 
-			context.subscriptions.push(commands.registerCommand(Commands.CONFIGURATION_UPDATE, uri => projectConfigurationUpdate(this.languageClient, uri)));
+			context.subscriptions.push(commands.registerCommand(Commands.CONFIGURATION_UPDATE, async (uri) => projectConfigurationUpdate(this.languageClient, uri)));
```
Eskibear, line 314: what's the difference if we only prepend 'async' here without 'await'?

CsCherrYY: I think I miss the `await` here. The same to line 322.

src/hoverAction.ts
CsCherrYY: just checked again, and find hover in super implementation works.
image

Eskibear (approved):
LGTM.

rgrunber:
I ran into a problem updating to the language client to 8.0.1 in vscode-xml, and I suspect it'll be the same here. The extension will break on Theia.

Note: Node 14 needs to be used as with 12, I ran into `SyntaxError: Unexpected Token '?'`. (nullish coalescing issue).
Afterwards I ran into :

```
2022-05-26T13:34:36.203Z root INFO [hosted-plugin: 9987] PLUGIN_HOST(9987): PluginManagerExtImpl/loadPlugin(/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension)
2022-05-26T13:34:36.257Z root ERROR [hosted-plugin: 9987] Activating extension 'XML' failed: TypeError: Class extends value undefined is not a constructor or null
    at Object.1365 (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:447499)
    at n (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:655557)
    at Object.71 (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:340748)
    at n (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:655557)
    at Object.4384 (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:307797)
    at n (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:655557)
    at Object.2850 (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:481672)
    at n (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:655557)
    at Object.4060 (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:251645)
    at n (/home/rgrunber/.theia/extensions/vscode-xml-0.21.0.vsix/extension/dist/extension.js:2:655557)
```

`extension.js:2:447499` points to `class ProtocolTypeHierarchyItem extends code.TypeHierarchyItem`
`extension.js:2:251645` points to `const vscode_languageclient_1 = __webpack_require__(48);`

If I'm understanding how the `__webpack_require__` works, `const vscode_languageclient_1` is in `47`, which gets called at the very beginning by `const xmlExtensionApiImplementation_1 = __webpack_require__(47);` . This is from `xmlExtensionApiImplementation.ts` which originates in the vscode-xml sources.

So the act of referencing the language client causes TypeHierarchyItem to be referenced, which [Theia does not support](https://eclipse-theia.github.io/vscode-theia-comparator/status.html)

CC'ing @fbricon , @testforstephen in case I'm missing an obvious way to work around this.

CsCherrYY:
@rgrunber Yeah I had opened an issue here: https://github.com/eclipse-theia/theia/issues/10993, it seems Theia should support the related types (e.g., TypeHierarchyItem).

rgrunber:
We should re-investigate this soon. Once Theia releases, I'm hoping this will work. It seems they recently started supporting the Type Hierarchy API.

https://eclipse-theia.github.io/vscode-theia-comparator/status.html

CsCherrYY:
it seems that Theia has released a new version 1.31.0 which supports TypeHierarchyItem: https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md#breaking_changes_1.31.0

I will update the PR these days.

rgrunber:
I may have been a bit too optimistic. We will still need to wait for the end of the year to be able to make this change, but at least it will happen eventually :(

CsCherrYY:
@rgrunber is there anything I miss about this change? 

One concern is that we seem not to have a way to define the required Theia version in package.json, since only the newest Theia supports this API.

CsCherrYY:
https://github.com/redhat-developer/vscode-java/pull/2377/commits/3fba2fee58db1310dba9b8be3093c698bb38a20f updates the language client to newest 8.0.2. Comparing it with the former 8.0.0-next 4, the changes can be found also in https://github.com/microsoft/vscode-languageserver-node#3170-protocol-800-json-rpc-800-client-and-800-server:

- removeal of `onReady()`
- change of `ErrorHandler`

datho7561 (changes requested):

datho7561:
I tried this out, and I get an error notification when the syntax server is shut down. Is there a way to avoid this?

CsCherrYY:
> I tried this out, and I get an error notification when the syntax server is shut down. Is there a way to avoid this?

In the current version, when switching from syntax server to the standard server, we can also see these error messages in the syntax server output channel:
image

And the behavior is introduced in https://github.com/microsoft/vscode-languageserver-node/issues/824, which forces the error message shown in the pop up. So I think the current way to solve this is to know why there is an error when syntax server stop.

CsCherrYY:
maybe https://github.com/eclipse/eclipse.jdt.ls/pull/1790 is related. @rgrunber could you provide any comments?

package.json
```diff
@@ -1414,7 +1414,7 @@
     "htmlparser2": "6.0.1",
     "jdk-utils": "^0.4.4",
     "semver": "^7.3.5",
-    "vscode-languageclient": "7.1.0-next.5",
+    "vscode-languageclient": "^8.0.2",
```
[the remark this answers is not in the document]
CsCherrYY: Sounds great. added in https://github.com/redhat-developer/vscode-java/pull/2377/commits/c5afa6043ccdd3e51915a3c5a2ac4935ed30847b :)

CsCherrYY:
@datho7561 It seems that if we disable the client capability `shouldLanguageServerExitOnShutdown`, the error notification will not appear and the syntax server can shutdown successfully. Could you help verify it?

image

@rgrunber If it's true, I guess language client 8+ has fixed what we patch in https://github.com/eclipse/eclipse.jdt.ls/pull/1790.

datho7561:
> ... if we disable the client capability `shouldLanguageServerExitOnShutdown`, the error notification will not appear and the syntax server can shutdown successfully

Yes, i think this is the right fix. This is what we did in https://github.com/redhat-developer/vscode-xml/commit/825823c6e7199052d338166d91dca13d434bf9a6 for vscode-xml

src/providerDispatcher.ts
```diff
@@ -99,10 +99,10 @@ function createDocumentSymbolProvider(): DocumentSymbolProvider {
 			}
 
 			if ((<any>symbolResponse[0]).containerName) {
-				return languageClient.protocol2CodeConverter.asSymbolInformations(<clientSymbolInformation[]>symbolResponse);
+				return await languageClient.protocol2CodeConverter.asSymbolInformations(<clientSymbolInformation[]>symbolResponse);
```
datho7561, line 102: If the result of the `protocol2CodeConverter` is a `Promise`, I think you should be able to return the `Promise` without using `await` and it will do the same thing. (this isn't a big deal though).

datho7561 (changes requested):
After `shouldLanguageServerExitOnShutdown` is removed/set to false, I think this PR will be good

jdneo:
BTW, dose the new version of language client has native support of inlay hint? If yes, maybe we can remove https://github.com/redhat-developer/vscode-java/blob/master/src/inlayHintsProvider.ts and its related code.

CsCherrYY:
@jdneo I guess the anwser is yes. See: https://github.com/microsoft/vscode-languageserver-node#3170-protocol-800-json-rpc-800-client-and-800-server inlay hint support is in the list here. I also find the related types in the language-client library.

CsCherrYY:
it looks like `vscode-languageclient` version 8.1.0 is available, I'll take a look if we can directly update to this version.

src/extension.ts
```diff
@@ -167,16 +167,17 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 						resolveAdditionalTextEditsSupport: true,
 						advancedIntroduceParameterRefactoringSupport: true,
 						actionableRuntimeNotificationSupport: true,
-						shouldLanguageServerExitOnShutdown: true,
+						// https://github.com/redhat-developer/vscode-java/pull/2377#issuecomment-1427268344
+						shouldLanguageServerExitOnShutdown: false,
```
rgrunber, line 171: Yes, this was fixed in the language client release that supported 3.17. Original issue is at https://github.com/redhat-developer/vscode-java/issues/1928#issuecomment-832679999 . There's no need to set it to false as that's the default value anyways on the server side. In fact, there's a good argument for getting rid of it entirely since it was really to handle some non-protocol behaviour.

rgrunber (commented):
Overall, looks fine. I tried out rename, codelens, inlay hints, type hierarchy as those seem to be more at risk for breakage. It all seems to work well. The syntax server shuts down on its own as well.

I did noticed : https://github.com/redhat-developer/vscode-java/actions/runs/4200047119/jobs/7285652596#step:9:13 . The build still passes. Is this safe ?

Eskibear:
> The build still passes. Is this safe ?

@rgrunber 
Probably we should upgrade version of typescript. See https://github.com/microsoft/TypeScript/issues/1778#issuecomment-913202933

rgrunber:
Ok, then let's go ahead and update typescript here if that will fix it.

package.json
```diff
@@ -1397,7 +1397,7 @@
     "@types/glob": "5.0.30",
     "@types/lodash.findindex": "^4.6.6",
     "@types/mocha": "^5.2.5",
-    "@types/node": "^8.10.51",
```
CsCherrYY, line 1400: Since VS Code has switched to node 16 since 1.66 (https://code.visualstudio.com/updates/v1_66#_nodemoduleversion-and-nodejs-api-update), let's update this deps as well. The update will resolve the following error:
> node_modules/vscode-jsonrpc/lib/node/main.d.ts(8,37): error TS2307: Cannot find module 'worker_threads' or its corresponding type declarations.

CsCherrYY:
@rgrunber it looks like that issue got resolved after updating typescript and node :)

PR #2403, opened 2022-04-12
fix client.log: wrong default project jdk

### LINKED ISSUE #2401: wrong message about default project jdk in client.log
With below settings, drag a .java file into an empty window.
```
"java.jdt.ls.java.home": "C:\\Users\\admin\\AppData\\Local\\Programs\\Eclipse Adoptium\\jdk-17.0.1.12-hotspot"
```

In client.log, you'll see below, which is wrong.

```
{
  message: "Use the JDK from 'java.home' setting as the initial default project JDK.",
  level: 'info',
  timestamp: '2022-04-12 14:30:10.983'
}
```

'java.home' should be 'java.jdt.ls.java.home'.

Eskibear:
fix #2401

### DIFF
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -81,7 +81,7 @@ export async function resolveRequirements(context: ExtensionContext): Promise<Re
              * We'll keep it for compatibility.
              */
             if (javaHome) {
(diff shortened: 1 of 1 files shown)


Eskibear:
@testforstephen please take a look.

testforstephen (approved):
LGTM

PR #2409, opened 2022-04-13
Bugfix: New java file snippet is broken in 1.5.0

### LINKED ISSUE #2408: New Java file snippet is broken in 1.5.0
In 1.5.0 RC bits, new a Java file from File Explorer, it just creates an empty file without any class body generated. This is a regression introduced by https://github.com/redhat-developer/vscode-java/pull/2349

testforstephen:

Fix #2408


rgrunber (approved):
You're right. If we're calling with the first argument being `Commands.EXECUTE_WORKSPACE_COMMAND`, then this is meant for the language server. If that's the case then we don't want the `CMD` version. That one is meant to be the first argument and execute locally.

PR #2415, opened 2022-04-15
Migrate from tslint to eslint

FieteO:
This PR aims to replace tslint with eslint, since the former [is deprecated since 2019](https://blog.palantir.com/tslint-in-2019-1a144c2317a9).


datho7561 (changes requested):
Looks like a promising start, thanks for the contribution! I just have a few suggestions that might make the eslint setup a bit more helpful. As well, I found a few linting errors in the codebase after I rebased this PR.

datho7561:
I'm going to push a commit onto this PR in order to address the conflicts and the feedback I had...

FieteO:
Sorry for not giving feedback @datho7561. I didn't really have time to check what my reasoning was :D As far as I can remember including the custom arrow plugin and not including the recommended rules was to recreate the tslint status-quo as close as possible.

datho7561:
> Sorry for not giving feedback @datho7561. I didn't really have time to check what my reasoning was :D
No worries!
> As far as I can remember including the custom arrow plugin and not including the recommended rules was to recreate the tslint status-quo as close as possible.
Okay. That makes sense. Using the recommended rules meant that I needed to refactor a large portion of the codebase to fix lint errors, so it might make sense to stick with what you had. @rgrunber, what do you think about this?

datho7561:
I removed the part where I inherited the default eslint configuration and the refactoring I did to adapt to this, but left in switching from the extension for arrow functions to the default eslint arrow function functionality.

datho7561:
Taking a look at it, I don't think `eslint` is being run on TypeScript files as-is. If I run `npx eslint --ext .js,.ts .`, then I get a bunch of naming errors in the typescript files.

datho7561:
@rgrunber I just pushed a commit that makes sure that the `webpack watch` task works correctly (VS Code waits until webpack finishes bundling the extension before launching)

rgrunber (approved):
I've tested this out and it works for me. With the installed recommended problem matchers we even wait for the compilation to complete, which is nice.

The linting is quite a bit more strict, but that may be for the best.

PR #2426, opened 2022-04-21
Trigger signature help on completion item selected

jdneo:
requires https://github.com/eclipse/eclipse.jdt.ls/pull/2065

Now when a completion item is selected, it will automatically trigger signature help:

a

The value of the setting `java.signatureHelp.enabled` is still disabled by default. The reason is that, currently we cannot get which item is selected and provide that signature correctly. Considering method like `System.out.println()`, which has multiple overload versions, the current implementation won't work.

a

To solve this problem, we can leverage the command field of the completion item. Send back the selected one's information and cache it. But that needs some refactoring works to be done in IntelliCode as well. So currently, just keep it to `false`.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -246,7 +246,8 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 						resolveAdditionalTextEditsSupport: true,
 						advancedIntroduceParameterRefactoringSupport: true,
 						actionableRuntimeNotificationSupport: true,
-						shouldLanguageServerExitOnShutdown: true
+						shouldLanguageServerExitOnShutdown: true,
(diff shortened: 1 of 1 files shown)


rgrunber:
Right. I ran into this and wondered if https://github.com/eclipse/eclipse.jdt.ls/issues/2025 might improve the situation.

rgrunber (approved):
Change works for me. Feel free to merge when ready.

jdneo:
> Right. I ran into this and wondered if https://github.com/eclipse/eclipse.jdt.ls/issues/2025 might improve the situation.

Unfortunately, that PR won't help in this case, since currently the user's selection does not report back to the server. But since we have the completion item command. Technically, it should be achievable.

PR #2427, opened 2022-04-22
Remove the setting: 'java.completion.overwrite'

jdneo:
- Instead, we use 'editor.suggest.insertMode' to control the default
  completion behavior.

requires: https://github.com/eclipse/eclipse.jdt.ls/pull/2057

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ New in 1.6.0
   - `setDefault`: Set the default workspace encoding settings.
 
 Removed in 1.6.0
-* `java.completion.overwrite`: Please use `editor.suggest.insertMode` instead. Meanwhile, you can quickly switch to another mode by pressing `Shift` + `Tab` or `Shift` + `Enter` (the shortcut may be different depending on the local keymap settings).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -209,6 +208,9 @@ New in 1.6.0
   - `warning`: Show warning if a project has no explicit encoding set.
   - `setDefault`: Set the default workspace encoding settings.
 
+Removed in 1.6.0
+* `java.completion.overwrite`: Please use `editor.suggest.insertMode` instead. Meanwhile, you can quickly switch to another mode by pressing `Shift` + `Tab` or `Shift` + `Enter` (the shortcut may be different depending on the local keymap settings).
```
rgrunber, line 212: I would mention the keybinding ID that manages this behaviour. It's `insertPrevSuggestion` right ?

```suggestion
* `java.completion.overwrite`: Please use `editor.suggest.insertMode` instead. You can quickly toggle between `insert` & `replace` modes with `Shift` + `Tab` or `Shift` + `Enter` (default keyboard shortcuts). See `insertPrevSuggestion` from the keyboard shortcuts menu for further details.
```

jdneo: Did some tests, I think the command ID is: `acceptAlternativeSelectedSuggestion`

PR #2442, opened 2022-05-05
Support Gradle invalid type code error check

### LINKED ISSUE #1594: Gradle using Java 8 - Any gradle build script error returns as a StreamCorruptedException with no useful content instead of the actual gradle exception
When using Gradle with java.configuration.runtimes set to a JavaSE-1.8 JVM, any problems in a build.gradle script will come up with a java.io.StreamCorruptedException instead of the actual exception.

This can be very painful to debug, because it might happen that when executing gradle manually via command line, you might find that the build itself will go through with one of the default targets, but if one of the sub projects is executed it will throw an error.

##### Environment
- Operating System: Centos 7
- JDK version: 1.8.0b252
- Visual Studio Code version: 1.48.1
- Java extension version: 0.65.0

##### Steps To Reproduce
1. Write a gradle script with any sort of gradle parsing error
2. See problems menu that only describes the error as
```Could not run build action using Gradle distribution '(gradle distro)'.
java.io.StreamCorruptedException: invalid type code: 00
invalid type code: 00
```

Example script:
```gradle
this is just random text in a build.gradle
```

##### Current Result
Returns blank exception

##### Expected Result
* What went wrong:
Could not compile build file 'build.gradle'.
> startup failed:
  build file 'build.gradle': 1: unexpected token: in @ line 1, column 26.
     this is just random text in a build.gradle

CsCherrYY:
fix #1594, requires https://github.com/eclipse/eclipse.jdt.ls/pull/2082

The upstream issues are https://github.com/gradle/gradle/issues/13957, https://github.com/gradle/gradle/issues/9339, and https://github.com/eclipse/buildship/issues/1084.

There is an upstream PR: https://github.com/gradle/gradle/pull/17397, which fixes this issue and has been available since Gradle 7.2. So we offer a quickfix and notification to help users upgrade wrapper to see the exact error.

quickfix

### DIFF
--- a/src/gradle/gradleCodeActionProvider.ts
+++ b/src/gradle/gradleCodeActionProvider.ts
@@ -1,0 +1,55 @@
+'use strict';
+
+import * as fse from "fs-extra";
+import * as path from "path";
+import { CancellationToken, CodeAction, CodeActionContext, CodeActionKind, CodeActionProvider, CodeActionProviderMetadata, Command, commands, Diagnostic, DiagnosticRelatedInformation, ExtensionContext, ProviderResult, Range, Selection, TextDocument, Uri } from "vscode";
+import { Commands } from "../commands";
+import { upgradeGradle } from "../standardLanguageClient";
+
+export class GradleCodeActionProvider implements CodeActionProvider<CodeAction> {
+
+	private static UPGRADE_GRADLE_WRAPPER_TITLE = "Upgrade Gradle Wrapper";
+	private static WRAPPER_PROPERTIES_DESCRIPTOR = "gradle/wrapper/gradle-wrapper.properties";
+	private static GRADLE_INVALID_TYPE_CODE_MESSAGE = "Exact exceptions are not shown due to an outdated Gradle version, please consider to update your Gradle version.";
+
+	constructor(context: ExtensionContext) {
+		context.subscriptions.push(commands.registerCommand(Commands.UPGRADE_GRADLE_WRAPPER, (projectUri: string) => {
+			upgradeGradle(projectUri);
+		}));
+	}
+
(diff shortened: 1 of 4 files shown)


src/gradle/gradleCodeActionProvider.ts
```diff
@@ -0,0 +1,55 @@
...
+			return this.provideGradleCodeActions(document, context.diagnostics);
+		}
+		return undefined;
+	}
+
+	private async provideGradleCodeActions(document: TextDocument, diagnostics: readonly Diagnostic[]): Promise<CodeAction[]> {
+		const codeActions = [];
+		for (const diagnostic of diagnostics) {
+			const documentUri = document.uri.toString();
+			if (documentUri.endsWith(GradleCodeActionProvider.WRAPPER_PROPERTIES_DESCRIPTOR) && diagnostic.message === GradleCodeActionProvider.GRADLE_INVALID_TYPE_CODE_MESSAGE) {
```
testforstephen, line 32: I suggest to assign a diagnostic `code` for this kind of diagnostic in language server side, and then the client side can compare with the code directly.

CsCherrYY: Good idea. In this way, which `code` should we use? (to avoid conflicting with existing ones) In the meantime, the `code` will also show in the PROBLEMS panel.

testforstephen: See the current JDT problem code list.
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/2f33cdff07b101d0b9965f6602699acff3fd31b3/org.eclipse.jdt.core/compiler/org/eclipse/jdt/core/compiler/IProblem.java#L365

I would add a new category for gradle related problem.
```java
int GradleRelated = 0x00080000;
int InvalidTypeCode = GradleRelated + 1;
```

testforstephen (approved):

PR #2449, opened 2022-05-12
Reveal text changes from Source Actions

CsCherrYY:
Currently if you trigger source actions from light bulb in type declaration, the generated code will be at the end of the type declaration, which makes users hard to find. This PR adds a reveal step to the source actions.

https://user-images.githubusercontent.com/45906942/167979784-29b15a1b-6a85-426c-92ae-9fcb5109e152.mp4

Signed-off-by: Shi Chen <chenshi@microsoft.com>

PR #2450, opened 2022-05-12
Support separate "Generate Getters" and "Generate Setters"

### LINKED ISSUE #2362: Add individual "Generate Getters" & "Generate Setters" to Source Action menu
**Feature Request:**
We already have an option for generating Getter and Setter together. But we don't have an option to generate Getter and Setter separately while IntelliJ Idea by Jet Brains already has this feature.  **:(** 
image

Can you please disable this option for adding this line when Overriding a method?
image
The line starting with TODO is a very annoying feature. I always have to remove it. There does not seem to be a big problem here, but still.

**Bug:**
Whenever I generate a constructor in the active cursor line. The constructor is always generated after the last character (excluding the space). Here's a video clip explaining my issue if you didn't get it.

https://user-images.githubusercontent.com/90521537/160248863-b5c9d328-cab6-449d-afff-c13b26f983d3.mp4

CsCherrYY:
fix #2362, related to https://github.com/eclipse/eclipse.jdt.ls/pull/2086

https://user-images.githubusercontent.com/45906942/167986447-07cf3847-4910-4d8d-a5f0-79fc9b19c2c4.mp4

Signed-off-by: Shi Chen [chenshi@microsoft.com](mailto:chenshi@microsoft.com)


src/sourceAction.ts
```diff
@@ -63,6 +65,7 @@ function registerOverrideMethodsCommand(languageClient: LanguageClient, context:
             overridableMethods: selectedItems.map((item) => item.originalMethod),
         });
         await applyWorkspaceEdit(workspaceEdit, languageClient);
+        await revealWorkspaceEdit(workspaceEdit, languageClient);
```
CsCherrYY, line 68: see https://github.com/redhat-developer/vscode-java/pull/2449

rgrunber (approved):
LGTM. Only question to address is on the LS-side, whether it makes sense to unify the 3 registered commands into one by also passing the accessor kind in the callback with the `CodeActionParams`.

src/sourceAction.ts
```diff
@@ -106,13 +107,15 @@ function registerHashCodeEqualsCommand(languageClient: LanguageClient, context:
...
         await applyWorkspaceEdit(workspaceEdit, languageClient);
+        await revealWorkspaceEdit(workspaceEdit, languageClient);
     }));
 }
 
 function registerOrganizeImportsCommand(languageClient: LanguageClient, context: ExtensionContext): void {
     context.subscriptions.push(commands.registerCommand(Commands.ORGANIZE_IMPORTS, async (params: CodeActionParams) => {
         const workspaceEdit = await languageClient.sendRequest(OrganizeImportsRequest.type, params);
         await applyWorkspaceEdit(workspaceEdit, languageClient);
+        await revealWorkspaceEdit(workspaceEdit, languageClient);
```
testforstephen, line 118: For code generation features, it's good to move the focus to the generated code. But I doubt if it's a good idea to enable it for "Organize Imports" feature. Because "Organize Imports" might do nothing, or just remove and reorder import statements.

src/sourceAction.ts
```diff
@@ -356,5 +380,18 @@ function registerGenerateDelegateMethodsCommand(languageClient: LanguageClient,
             delegateEntries,
         });
         await applyWorkspaceEdit(workspaceEdit, languageClient);
+        await revealWorkspaceEdit(workspaceEdit, languageClient);
     }));
 }
+
+async function revealWorkspaceEdit(workspaceEdit: WorkspaceEdit, languageClient: LanguageClient): Promise<void> {
+    const codeWorkspaceEdit = languageClient.protocol2CodeConverter.asWorkspaceEdit(workspaceEdit);
+    for (const entry of codeWorkspaceEdit.entries()) {
```
testforstephen, line 389: NPE when "Organize Imports" do nothing.

image

CsCherrYY: Yeah, we need a null check here.

src/sourceAction.ts
CsCherrYY: Agree. Let's keep the current behavior for "Organize Imports"

testforstephen (approved):

Issue #2429, opened 2022-04-24
Should explicitly report the fatal settings error like "URI is not hierarchical"


testforstephen:
See https://github.com/redhat-developer/vscode-java/issues/2260#issuecomment-1008825791 and https://github.com/microsoft/vscode-java-debug/issues/1148#issuecomment-1107679538.

We frequently received issues about `"java.lang.IllegalArgumentException: URI is not hierarchical"` during "Initialize workspace". The root cause is that the user's settings like `java.settings.url` or `java.format.settings.url` point to an illegal url format, and JLS just fails silently. 

I think we can improve two things:
- Since this error is fatal, we should report a more explicit notification to tell user what's happening and guide user how to fix it.
- We should catch this exception in JLS, and avoid breaking the startup of language server.

rgrunber:
Is there any way to catch these on validating the setting ? Or is `pattern` not enough to handle every case ?

testforstephen:
It definitly helps if the client side could provide some validation logic to guard these settings. Also it's better to provide some samples of these settings in the description section.

Currently these url settings accept a file path or a url, that's too flexible and error-prone to configure manually. To be honest, I didn't really know how to properly configure these settings at first.

PR #2451, opened 2022-05-13
Should explicitly report the fatal settings error

snjeza:
Fixes https://github.com/redhat-developer/vscode-java/issues/2429
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2089

Test vsix: https://github.com/snjeza/vscode-test/raw/master/java-1.6.1.vsix


rgrunber (changes requested):
Seems pretty reasonable, but things like :
`~/Scripts/org.eclipse.jdt.core.prefs`
`${user.home}/Scripts/org.eclipse.jdt.core.prefs`
are valid.

We did this using https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/ResourceUtils.java#L249 on the server side.

Is there a way to allow such cases for now ?

snjeza:
> Is there a way to allow such cases for now ?

I will try.

gorkem:
@snjeza should this be closed?

Issue #2093, opened 2021-09-02
Java Model Exception: Java Model Status [<parent POM project name here> does not exist]


tigerinus:
[provide a description of the issue]

##### Environment
- JDK version: GraalVM-8 for project, GraalVM-11 for JDT extension.
- Java Language Server extension version: 0.81.0
- Extension Pack for Java version: 0.18.4 
- Visual Studio Code version: 

```
Version: 1.60.0 (system setup)
Commit: e7d7e9a9348e6a8cc8c03f877d39cb72e5dfb1ff
Date: 2021-09-01T10:41:52.311Z
Electron: 13.1.8
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Windows_NT x64 10.0.19043
```

- This is running on a SSH remote environment - not sure if it happens locally or not since we don't have such setup.
- Remote OS: Oracle Linux 8.

- Java related settings in VSCode (remote):

```json
    "java.completion.guessMethodArguments": true,
    "java.configuration.runtimes": [
        {
            "name": "JavaSE-1.8",
            "path": "/home/opc/.local/jdk",
            "default": true
        }
    ],
    "java.debug.settings.console": "internalConsole",
    "java.dependency.packagePresentation": "hierarchical",
    "java.dependency.showMembers": true,
    "java.home": "/home/opc/.local/graalvm-ce-java11",
    "java.implementationsCodeLens.enabled": true,
    "java.project.referencedLibraries": [
        "**/target/**/*.class",
        "/home/opc/.ivy2/cache/**/*.jar",
        "/home/opc/.m2/repository/**/*.jar"
    ],
    "java.referencesCodeLens.enabled": true,
```

##### Steps To Reproduce
1. Have a parant POM project whose `pom.xml` looks like:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>com.example</groupId>
	<artifactId>parent-example</artifactId>
	<version>0.1.6</version>
	<packaging>pom</packaging>
	<modules>
		<module>common</module>
		<module>service1</module>
		<module>service2</module>
[...]

PR #2459, opened 2022-05-18
Show field type when generating accessors

CsCherrYY:
requires [#2093](https://github.com/eclipse/eclipse.jdt.ls/pull/2093)

addtype


rgrunber (approved):
Looks fine. Just need to address a small thing on the JDT-LS end.

PR #2461, opened 2022-05-20
Add a new API to wait for standard server ready

jdneo:
- Add waitForServerReady() which returns a promise. The promise will
  only be resolved when the standard server is ready.

Currently we have the API `serverMode` to denote which mode the current language server is working on. But so far there is no API to let other extensions know when the standard server is ready. (Ready means it's ok to send request to server and an immediate response should be returned)

The newly added API can be used by all the side extensions (test runner, intellicode, ...), to let them only register their components after the standard server is ready. This can avoid lots of LSP requests pending when the server is building the project during startup.

A sample usage for the API could be:

```js
if (extensionApi.serverMode == LanguageServerMode.LightWeight) {
    if (extensionApi.onDidServerModeChange) {
        const onDidServerModeChange: Event<string> = extensionApi.onDidServerModeChange;
        context.subscriptions.push(onDidServerModeChange(async (mode: string) => {
            if (mode === LanguageServerMode.Standard) {
                registerComponents();
            }
        }));
    }
} else {
    await extensionApi.waitForServerReady();
    registerComponents();
}
```

### DIFF
--- a/src/apiManager.ts
+++ b/src/apiManager.ts
@@ -16,6 +16,7 @@ class ApiManager {
     private onDidClasspathUpdateEmitter: Emitter<Uri> = new Emitter<Uri>();
     private onDidServerModeChangeEmitter: Emitter<ServerMode> = new Emitter<ServerMode>();
     private onDidProjectsImportEmitter: Emitter<Uri[]> = new Emitter<Uri[]>();
+    private serverReadyPromiseResolve: (result: boolean) => void;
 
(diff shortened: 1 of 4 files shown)


testforstephen (approved):
LGTM

jdneo:
Note: The API is renamed to `serverReady()` in https://github.com/redhat-developer/vscode-java/pull/2469

PR #2472, opened 2022-05-26
VS Code doesn't call shutdown when exiting the standard client

### LINKED ISSUE #2471: VS Code doesn't call shutdown when exiting the standard client
##### Environment
- Operating System: any
- JDK version: any
- Visual Studio Code version: 1.67.2, code-insiders
- Java extension version: 1.6.0

##### Steps To Reproduce
- open a java project with VS Code
- stop it
- check the standard server log

##### Current Result
shutdown is not called

##### Expected Result
shutdown should be called

snjeza:
Fixes #2471 


rgrunber (approved):
This solves the issue for me. I tested multiple times as a packaged vsix as well as in a child vscode instance and it's working again.

PR #2484, opened 2022-05-30
added completion support for multiline comment blocks

### LINKED ISSUE #2464: Add completion support for multiline comment blocks
When typing following in a java file

```
/*
```
and pressing enter, I would expect it to be completed as follows.

```
/*
*
*/
```
which happens in Eclipse.

##### Environment
- Operating System: Any
- JDK version: Any
- Visual Studio Code version: 1.67.2
- Java extension version: 1.6.0

JessicaJHee:
Peek 2022-05-30 10-06

Fixes #2464 

Signed-off-by: Jessica He <jhe@redhat.com>


src/extension.ts
```diff
@@ -684,6 +684,12 @@ function enableJavadocSymbols() {
 				afterText: /^\s*\*\/$/,
 				action: { indentAction: IndentAction.IndentOutdent, appendText: ' * ' }
 			},
+			{
+				// e.g. /* | */
```
rgrunber, line 688: If this case is basically taken from the `/** | */` case, with just the 2nd `*` removed, have a look at how you can use the [`question mark (?)`](https://www.regular-expressions.info/optional.html) to support both cases in one expressions

rgrunber:
@snjeza you mentioned this would break `/**` in https://github.com/redhat-developer/vscode-java/issues/2464#issuecomment-1136331009 ? Can you explain ? If not, I'd like to merge this eventually as I see no issues.

fbricon:
* are not aligned, compared to javadoc behavior or in Eclipse.
You get 
```
/*
* 
*/
```
instead of 
```
/*
 * 
 */
```

snjeza:
> you mentioned this would break /** in https://github.com/redhat-developer/vscode-java/issues/2464#issuecomment-1136331009 ? Can you explain ? To me it doesn't really break,

It works fine.
com

src/extension.ts
```diff
@@ -680,7 +680,7 @@ function enableJavadocSymbols() {
 		onEnterRules: [
 			{
 				// e.g. /** | */
```
rgrunber, line 682: One small thing : `// e.g. /** | */ or /* | */` for the sake of documenting it.

snjeza:
> * are not aligned, compared to javadoc behavior or in Eclipse.
You get...

@fbricon I can't reproduce the issue. It works fine for me.

JessicaJHee:
> > * are not aligned, compared to javadoc behavior or in Eclipse.
> >   You get...
> 
> @fbricon I can't reproduce the issue. It works fine for me.

Sorry forgot to mention that it was fixed in the most recent commit.

rgrunber (approved):

PR #2495, opened 2022-06-06
Move to Java 17

snjeza:
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2118

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Set Up Java
         uses: actions/setup-java@v2
         with:
-          java-version: '11'
+          java-version: '17'
           distribution: 'adopt'
       - run: npm install -g typescript "vsce@<2"
       - run: wget http://download.eclipse.org/jdtls/snapshots/jdt-language-server-latest.tar.gz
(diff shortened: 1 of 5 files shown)


rgrunber:
Thanks for doing this. There's also a reference inside `pr-verify.yml` that should be updated. I'm not sure about `.sdkmanrc` but I suspect that as well. May need to figure out what the proper version is through there.

snjeza:
> There's also a reference inside pr-verify.yml

Fixed.

snjeza:
test this please

rgrunber:
Can we also just change the `.sdkmanrc` to contain `17.0.3-tem` . I think that should be the version from Adoptium. Otherwise, I think this is ready to be merged.

rgrunber:
We should also update https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements on the wiki page once this is merged.

snjeza:
> Can we also just change the .sdkmanrc to contain 17.0.3-tem . I think that should be the version from Adoptium. Otherwise, I think this is ready to be merged.

rgrunber (approved):

snjeza:
> We should also update https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements on the wiki page once this is merged.

@rgrunber I have updated it.

afriza:
Which last extension version supports < JavaSE-17 ? v1.7.0?
Because of extension auto-update, I get the install JDK 17 prompt.

rgrunber:
> Which last extension version supports < JavaSE-17 ? v1.7.0? Because of extension auto-update, I get the install JDK 17 prompt.

Yes, 1.7.0 would be the last version that can be started with a Java runtime less than version 17.

PR #2506, opened 2022-06-13
Fix moving refactoring not working in standard mode

### LINKED ISSUE #2503: TypeError: Cannot read properties of undefined (reading 'stop')

##### Environment
Version: 1.68.0 (user setup)
Commit: 4af164ea3a06f701fe3e89a2bcbb421d2026b68f
Date: 2022-06-08T11:44:16.822Z
Electron: 17.4.7
Chromium: 98.0.4758.141
Node.js: 16.13.0
V8: 9.8.177.13-electron.0
OS: Windows_NT x64 10.0.22000
- Java extension version: 1.7.0

Sometimes, after Java extension activated, I'll see `TypeError: Cannot read properties of undefined (reading 'stop')` in the debug console. Looks like the error happens at: https://github.com/redhat-developer/vscode-java/blob/4d954f70dd46aafd219245bf8e38c361492dfbfc/src/extension.ts#L430

image

When this error happens, we will lose some feature like auto refactoring when moving files in the explorer.

So far from my observation, this error happens with a high possibility on my Windows machine.

jdneo:
fix #2503 

Did some refactoring for each language client's stop() method. They still do the real stop stuff as the method name suggests, and the returned promise can be combined in the deactivated() function via `Promise.all`

### DIFF
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -527,6 +527,7 @@ export class StandardLanguageClient {
 				this.languageClient = null;
 			}
 		}
+		return Promise.resolve();
 	}
 
 	public getClient(): LanguageClient {
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -518,9 +518,14 @@ export class StandardLanguageClient {
 		}
 	}
 
-	public stop() {
+	public stop(): Promise<void> {
+		this.status = ClientStatus.Stopping;
 		if (this.languageClient) {
-			this.status = ClientStatus.Stopping;
+			try {
+				return this.languageClient.stop();
+			} finally {
+				this.languageClient = null;
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/standardLanguageClient.ts
```diff
@@ -518,9 +518,14 @@ export class StandardLanguageClient {
...
+	public stop(): Promise<void> {
+		this.status = ClientStatus.Stopping;
 		if (this.languageClient) {
-			this.status = ClientStatus.Stopping;
+			try {
+				return this.languageClient.stop();
+			} finally {
+				this.languageClient = null;
+			}
 		}
```
testforstephen, line 529: I suggest always returning a resolved promise via `Promise.resolve()` in the fallback path, so that the deactivate function doesn't need to check null value.

Same for syntaxLanguageClient.

jdneo: Makes sense

testforstephen (approved):

PR #2512, opened 2022-06-15
Verify that java.jdt.ls.java.home meets minimum JRE requirement.

rgrunber:
1. Set `java.jdt.ls.java.home` to some value (eg. `/usr/lib/jvm/java-11-openjdk/`).
2. Import some project
3. **Result below** :
java-jdt-ls-java-home-fails

- java.jdt.ls.java.home should only be used if it meets the minimum JRE
  requirement
- Default to the embedded JRE (if available) otherwise

@fbricon , @testforstephen , let me know if I've missed something here.

**Update** : Maybe there's an argument to be made for a notification given that we're ignoring the expected behaviour (of using the `java.jdt.ls.java.home` value) ?


testforstephen (approved):
> Maybe there's an argument to be made for a notification given that we're ignoring the expected behaviour (of using the java.jdt.ls.java.home value) ?

Agree. It makes sense to display such a notification. Also, for a workspace, we should only display it once.

Issue #2473, opened 2022-05-27
Expose 'Update Project' to more places


jdneo:
Quoting user's feedback:

Add a command to rebuild the project the current active file is part of.  Currently, I have to open the pom file and run "Java: Update Project".  **It would be nice to be able to run that command without having to go open the pom file first**.

a command that brings up a list of projects within the workspace, and **allows me to select which project to update would be nice**.  Sometimes I need to update multiple projects (for example, projects that generate code frequently get out of sync) but not all of them.  Currently, I either have to choose to update all projects and take a break while that happens, or open the POM file for each project needing to be updated and run "Java: Update Project".  Both options kinda suck, so improving this experience would be awesome!

jdneo:
Expose update project action to more places (for example, can be invoked in Java Project explorer, Maven explorer, Gradle explorer)
Allow users to select projects to update, and not limit the trigger files to pom and gradle files. (as long as the language server can find the belonging project with the uri string).

rgrunber:
As discussed, for the  labeling of the action, maybe `Update Project Configuration` makes more sense. `Update Configuration` (if you feel "Project" is implied by the context ?

The goal is to make it clear that this isn't an action you need to perform unless a configuration change to build files has been made.

jdneo:
> The goal is to make it clear that this isn't an action you need to perform unless a configuration change to build files has been made.

Yes, the current `Update Project` seems not clearly infers the meaning of the command. `Update Project Configuration` can be a candidate. (Only one concern: I'm not quite sure if users can connect the `project configuration` with the `build files`, will they think it's some other kind of configurations?)

----

By the way, I have some new findings:

### Auto Update

M2E does have the auto update feature enabled by default. If I added a new dependency and save the `pom.xml`, new types are available in the completion list.

But for Buildship it still needs an update to include new dependency.

Based on that, I think the change that pop up project lists when users are editing java files still makes sense. Considering when users are working on a Gradle project, he introduced a new dependency in `build.gradle` and forget to update. He might realize that when he cannot find the new types in the completion list, then he can trigger an update from command palette directly.

### Our recommend choice for configuration update

I found that currently our recommend choice for configuration update is `Never`, which surprises me. (I guess we don't have any recommend at first, but after some version of VS Code, it will always highlight the first item?)

![WeChat Screenshot_20220616085436](https://user-images.githubusercontent.com/6193897/173970165-0f7736c9-a07b-4bb8-a32e-38c85edd5cca.png)

[...]

PR #2513, opened 2022-06-17
Support project update selection

jdneo:
- If no uri is passed into the projectConfigurationUpdate() function,
  show a quick pick list and let user select which projects to update.
- If there is only one project in the workspace, project update request
  for that project will directly send to server without asking.

part of #2473, requires https://github.com/eclipse/eclipse.jdt.ls/pull/2131

### DIFF
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -642,31 +642,29 @@ async function askForProjectToUpdate(): Promise<Uri[]> {
 
 	if (projectPicks.length === 0) {
 		return [];
-	}
-
-	// pre-select an active project based on the uri candidate.
-	if (uriCandidate) {
-		const candidatePath = uriCandidate.fsPath;
-		let belongingIndex = -1;
-		for (let i = 0; i < projectPicks.length; i++) {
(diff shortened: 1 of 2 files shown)


snjeza (approved):

src/standardLanguageClient.ts
```diff
@@ -585,21 +585,97 @@ function setIncompleteClasspathSeverity(severity: string) {
...
+
+		return {
+			label: path.basename(projectPath),
+			detail: projectPath,
+		};
+	}).filter(Boolean);
+
+	if (projectPicks.length === 0) {
+		return [];
+	}
```
rgrunber, line 645: I would make this :

```
if (projectPicks.length === 0) {
  ...
} else if (projectPicks.length === 1) {
  ...
} else {
 // pre-select an active project based on the uri candidate.
 ...
 ...
 const choices:  ...
}
```

rgrunber (commented):
LGTM. Just one style-based suggestion for ease of reading the cases and at best avoiding performing some unnecessary logic in certain cases.

PR #2514, opened 2022-06-17
Allow to clean workspace without pop-up notification

testforstephen:

This won't affect the UX when triggering it from command palette. It's used for side extensions, where we need to invoke "Clean Workspace" command in some failing case. Giving an option "force" is used to avoid poping up notification multiple times.


rgrunber (approved):
LGTM. Just an API addition that is no-op on this end.

PR #2517, opened 2022-06-21
adopt change of ref-viewlet's publisher

Eskibear:
publisher of ref-viewlet extension has been changed in upstream. This PR keeps compatibility.

See:
https://github.com/microsoft/vscode/pull/152213


jdneo (approved):

testforstephen (approved):
LGTM. Adding a fallback check is a safe strategy.

PR #2519, opened 2022-06-22
Lombok support v1

XiaoYuhao:
Add the lombok support:

1. It can automatically detect whether the project has imported the lombok by checking all classpath.
2. It can show the current version of the lombok in the language status bar.
3. It can dynamically detect the version change of the lombok.


testforstephen:
Pls fix tslint errors first. You can run the command `npm run tslint` locally to check tslint errors.

package.json
```diff
@@ -898,6 +898,12 @@
           "default": "ignore",
           "markdownDescription": "Project encoding settings",
           "scope": "window"
+        },
+        "java.jdt.ls.lombokSupport.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable lombok support",
```
testforstephen, line 905: "Whether to load lombok processors from project classpath"

src/lombokSupport.ts
```diff
@@ -0,0 +1,304 @@
...
+						}
+						if (name==="parent") {
+							hasParent = true;
+						}
+						if (hasParent&&name==="relativePath") {
+							getRelativePath = true;
+						}
+					},
+					ontext(text) {
+						if (tagList.length>0) {
```
testforstephen, line 161: `if (tagList.length)` is equivalent as `if (tagList.length > 0)`

src/settings.ts
```diff
@@ -107,7 +107,8 @@ function hasJavaConfigChanged(oldConfig: WorkspaceConfiguration, newConfig: Work
 		|| hasConfigKeyChanged('home', oldConfig, newConfig)
 		|| hasConfigKeyChanged('jdt.ls.vmargs', oldConfig, newConfig)
 		|| hasConfigKeyChanged('progressReports.enabled', oldConfig, newConfig)
-		|| hasConfigKeyChanged('server.launchMode', oldConfig, newConfig);
+		|| hasConfigKeyChanged('server.launchMode', oldConfig, newConfig)
+		|| hasConfigKeyChanged('jdt.ls.lombokSupport.enabled', oldConfig, newConfig);
```
testforstephen, line 111: We could ignore lombokSupport settings change if the user project is not using lombok. Only requires a restart when both the setting is changed and lombok agent path is cached in state.

src/lombokSupport.ts
```diff
@@ -0,0 +1,304 @@
...
+async function getParentBuildFilePath(buildFilePath: string, relativePath: string) {
+	const path = require('path');
+	const parentBuildFilePath = path.join(path.dirname(buildFilePath), relativePath, path.basename(buildFilePath));
+	if (await fse.pathExists(parentBuildFilePath)) {
+		return parentBuildFilePath;
+	}
+	return undefined;
+}
+
+export function addLombokChangeVersionCommand(context: ExtensionContext) {
```
testforstephen, line 133: => registerLombokConfigureCommand

src/lombokSupport.ts
```diff
@@ -0,0 +1,304 @@
...
+	if (hasLombokChangeVersionCommand) {
```
testforstephen, line 134: => isLombokCommandInitialized

src/lombokSupport.ts
```diff
@@ -0,0 +1,304 @@
...
+			const action = 'Reload';
+			const restartId = Commands.RELOAD_WINDOW;
+			window.showInformationMessage(msg, action).then((selection) => {
+				if (action === selection) {
+					commands.executeCommand(restartId);
+				}
+			});
+		}
+		else {
+			const msg = `Do you want to reload the vscode for Lombok support?`;
```
testforstephen, line 109: After a rethink, we can always store the lombok path in the workspace state first. The message is just used to ask user whether to reload and make it take effect immediately.

**The message could be:**
"Lombok is detected in your project, please restart ${env.appName} to enable lombok support."

**The button could be:**
Restart Now

src/lombokSupport.ts
```diff
@@ -0,0 +1,304 @@
...
+		return {
+			title: `Change Verison`,
+			command: Commands.LOMBOK_CONFIGURE,
+			arguments: [buildFilePath],
+			tooltip: `Open ${relativePath}`
+		};
+	}
+}
+
+function gotoLombokConfigure(position: number, buildFilePath: string): void {
```
testforstephen, line 272: => gotoLombokLocation

src/lombokSupport.ts
```diff
@@ -0,0 +1,304 @@
...
+	if (context.workspaceState.get(JAVA_LOMBOK_IMPORTED)===true) {
+		// Exclude user setting lombok agent parameter
+		const reg = /[\\|/]lombok.*\.jar/;
+	    const deleteIndex = [];
+		for (let i = 0; i<params.length; i++) {
+			if (reg.test(params[i])) {
+				deleteIndex.push(i);
+			}
+		}
+		for (const index of deleteIndex) {
```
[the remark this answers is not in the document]
XiaoYuhao: Thanks for your careful review. I have fixed this issue by traversing the array `deleteIndex`.

```
for (let i = deleteIndex.length - 1; i>=0; i--) {
	params.splice(deleteIndex[i], 1);
}
```

src/lombokSupport.ts
```diff
@@ -0,0 +1,304 @@
...
+	{ scheme: 'untitled', language: 'java' },
+	{ pattern: '**/pom.xml' },
+	{ pattern: '**/{build,settings}.gradle'},
+	{ pattern: '**/{build,settings}.gradle.kts'}
+];
+
+let hasLombokChangeVersionCommand: boolean = false;
+const lombokSupportEnable: boolean = vscode.workspace.getConfiguration().get("java.jdt.ls.lombokSupport.enabled");
+
+export function enableLombokSupport(): boolean {
```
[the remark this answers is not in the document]
XiaoYuhao: I think that the `java.jdt.ls.lombokSupport.enabled` setting should take effect after restarting the vscode. So I cache this setting value when vscode starts.

There will be something wrong if we get the isLombokSupportEnabled from current configuration. For example, if the user changes the `java.jdt.ls.lombokSupport.enabled` setting from disable to enable but does not restart the vscode, the language status bar will show the Lombok version although there is no Lombok support yet.

src/lombokSupport.ts
```diff
@@ -0,0 +1,320 @@
...
+	const reg = /lombok-.*\.jar/;
+	const lombokVersion = reg.exec(context.workspaceState.get(JAVA_ACTIVE_LOMBOK_PATH))[0].split('.jar')[0];
+	return lombokVersion;
+}
+
+export function addLombokParam(context: ExtensionContext, params: string[]) {
+	context.workspaceState.update(JAVA_ACTIVE_LOMBOK_PATH, "");
+	if (isLombokImported(context)) {
+		// Exclude user setting lombok agent parameter
+		const reg = /-javaagent:[\s\S]*?[\\|/]lombok.*\.jar/;
```
testforstephen, line 54: A simpler reg `/-javaagent:.*[\\|/]lombok.*\.jar/` ?

src/lombokSupport.ts
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}
+
+	function getLombokChangeCommand(buildFilePath: string): vscode.Command {
+		const relativePath = vscode.workspace.asRelativePath(buildFilePath);
+		return {
+			title: `Change Verison`,
+			command: Commands.LOMBOK_CONFIGURE,
+			arguments: [buildFilePath],
+			tooltip: `Open ${relativePath}`
```
testforstephen, line 283: Configure Lombok Version

src/lombokSupport.ts
```diff
@@ -0,0 +1,320 @@
...
+	commands.executeCommand(
+		Commands.GOTO_LOCATION,
+		window.activeTextEditor.document.uri,
+		window.activeTextEditor.selection.active,
+		[newLocation],
+		'goto'
+	);
+}
+
+function isMavenProject(buildFilePath: string): boolean {
```
testforstephen, line 302: https://github.com/redhat-developer/vscode-java/blob/master/src/extension.ts#L524

You need to consider the case where the project contains both maven and gradle build files. You could reference the link code to get the active build tool that the language server picked.

testforstephen (changes requested):
Play it on a multi-module maven projects, overall it works well. Here is a minor issue.

Toggle on/off the setting `java.jdt.ls.lombokSupport.enabled` a few times, the language status shows some weird lombok status.
image

testforstephen (approved):
LGTM. Tried it on maven and gradle lombok project, it works well. I think it's ready to merge.

In the current version, it prompts the user to reload the window for lombok support. We have tried restarting the language server without reloading, which is technically possible, but we found that this has side effects on third-party extensions, some of which report errors about delegate command exists. So in v1, we still use reloading to restart the language server. We will further investigate how to make third-party extensions work properly with language server restart.

Once the PR is merged, we'll update the wiki description as well. https://github.com/redhat-developer/vscode-java/wiki/Lombok-support

rgrunber:
I've rebased this change and just reverted the part about using lockFileVersion 2 since that would require npm v7, and we currently build with Node v14 (npm v6). The additiona of htmlparser2 seems to work just fine keeping the lockfile version the same.

rgrunber (approved):

PR #2521, opened 2022-06-22
Add DCO documentation

fbricon:
Add DCO documentation


CONTRIBUTING.md
```diff
@@ -193,3 +193,10 @@ If you encounter a problem and know it is caused by eclipse.jdt.ls, then please
 In doubt, you can report issues in the [vscode-java issue tracker](https://github.com/redhat-developer/vscode-java/issues).
 
 Try to collect as much informations as you can to describe the issue and help us reproduce the problem. Head over to the [troubleshooting page](https://github.com/redhat-developer/vscode-java/wiki/Troubleshooting#enable-logging) to see how to collect useful logging informations.
+
+### Certificate of Origin
```
rgrunber, line 197: If you think it should be placed at the bottom of the file so it's out of the way of more useful contributing instructions I would make it an h1 header (#), so it doesn't seem like it is a child of `Reporting Issues`.

Otherwise, I think it makes more sense to place `Certificate of Origin` under the `How to contribute` heading with the "h2" header (##).

rgrunber (commented):
Looks good, just a comment on placement.

CONTRIBUTING.md
rgrunber: NVM, I'm wrong.

PR #2522, opened 2022-06-22
Fix issues from the merging of the "Open All Log Files" command.

rgrunber:
See https://github.com/redhat-developer/vscode-java/pull/2375#pullrequestreview-1014460921 for reasoning.

- Make "java.open.serverLog" visible
- Properly hide the stdout, stderr & client logs

Although the titles won't be seen, I guess it makes sense to be consistent and add them back. Let me know if the new translations for the stdout/stderr make sense in `package.nls.zh.json`.


package.nls.zh.json
```diff
@@ -2,6 +2,10 @@
 	"java.server.mode.switch": "切换到标准模式",
 	"java.projectConfiguration.update": "更新项目",
 	"java.project.import": "将 Java 项目导入工作区",
+	"java.open.serverLog": "打开 Java 语言服务器日志文件",
+	"java.open.serverStdoutLog": "打开 Java 语言服务器输出日志文件",
+	"java.open.serverStderrLog": "打开 Java 语言服务器错误日志文件",
+	"java.open.clientLog": "打开 Java 插件日志文件",
```
jdneo, line 8: 👍

package.json
```diff
@@ -1086,6 +1087,18 @@
           "command": "java.project.import.command",
           "when": "javaLSReady"
         },
+        {
+          "command": "java.open.clientLog",
```
jdneo, line 1091: Is this hide on purpose? Currently it can be seen in command palette (`Open Java Extension Log File`)

rgrunber: Yes, I guess the goal was to reduce the amount of commands but if we keep the server logs, we may as well do client, so I have added it back.

jdneo: Got it.

Issue #2510, opened 2022-06-15
Expose the build command to more places


jdneo:
Below are some feedbacks from user surveys:

- there no build and run feature in a click like netbeans.
- Add a command to rebuild the project the current active file is part of. 
- "I really like *editing* in VS Code but find it not as powerful as tools like Eclipse for build and debug. " 
- The burden I got is there was not shortcut for compile java code, it would be awesome if we had it." 
- For a maven project I find I have to routinely run "mvn clean compile" after making code changes before it will build.

---------------

Actually, we have build action which is called `Force Java Compilation`, it looks like we need to do some enhancement to expose it to more places. Some ideas:

1. Expose build action to more places (for example, can be invoked in Java Project explorer)
2. Allow to build some of the projects instead of the entire workspace
3. Considering renaming it?? - Because in other IDEs they all call it `build`, here we call it `compilation`
4. Ask VS Code to have a dedicate tool bar menu called `build` or sth...

jdneo:
Some investigation about the naming.

#### Eclipse
In Eclipse, the action that generate the jar artifact is called `export`.
![export](https://user-images.githubusercontent.com/6193897/174934046-4c25bc1f-8411-404f-8647-1576cc713d7e.PNG)

And the build action will generate all the compiled classes in the output path:

https://user-images.githubusercontent.com/6193897/174934131-dc11b2d3-308c-4e7d-a4be-3238f239f0c2.mp4

#### IDEA
The term `Compile` and `Build` can be found both in IDEA. While, `Compile` is used to compile a single source file, `Build` is used for an entire module/project
![image](https://user-images.githubusercontent.com/6193897/174934365-8e647db4-97c5-4d9d-ac27-3f3d9195be7d.png)


https://user-images.githubusercontent.com/6193897/174934378-8bc0f3b5-e9f6-425f-bb9c-c12f50af92fb.mp4

Looks like the term `build` is commonly used in popular IDEs and does very similar things with our `Force Java compilation` command. I think it makes sense to make this command's name include the keyword `build`, so that users can find it more easily.

// cc @rgrunber @fbricon

PR #2526, opened 2022-06-23
Allow rebuild selected projects

jdneo:
part of https://github.com/redhat-developer/vscode-java/issues/2510, requires https://github.com/eclipse/eclipse.jdt.ls/pull/2138

- Add a new extended LSP request: BuildProjectRequest.
- the request param has two fields: `identifiers` - uri and `isFullBuild` - boolean.

Currently, `isFullBuild` is always set to true. Since we enable auto-build by default, triggering an incremental build does not make much sense. But a clean full build is helpful when the build status is out of sync. In the future, if users ask for incremental build support, we can expose this param.

The name of the command is called `Rebuild Projects` now, indicates it will clean and do a full build.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -80,6 +80,11 @@ export namespace Commands {
     export const COMPILE_WORKSPACE = 'java.workspace.compile';
 
     /**
+     * Execute build for projects
+     */
+    export const REBUILD_PROJECT = 'java.project.build';
+
+    /**
     * Open Java Language Server Log file
     */
     export const OPEN_SERVER_LOG = 'java.open.serverLog';
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -82,7 +82,7 @@ export namespace Commands {
     /**
      * Execute build for projects
      */
-    export const REBUILD_PROJECT = 'java.project.build';
+    export const BUILD_PROJECT = 'java.project.build';
 
     /**
     * Open Java Language Server Log file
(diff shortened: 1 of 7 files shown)


snjeza (approved):

src/commands.ts
```diff
@@ -79,6 +79,11 @@ export namespace Commands {
 	 */
     export const COMPILE_WORKSPACE = 'java.workspace.compile';
 
+    /**
+     * Execute build for projects
+     */
+    export const REBUILD_PROJECT = 'java.project.build';
```
CsCherrYY, line 85: The variable name "Rebuild" seems not consistent with the value "java.project.build". Maybe "java.project.rebuild" would be better?

src/standardLanguageClient.ts
```diff
@@ -640,40 +717,24 @@ async function askForProjectToUpdate(): Promise<Uri[]> {
...
-			const candidatePath = uriCandidate.fsPath;
-			let belongingIndex = -1;
-			for (let i = 0; i < projectPicks.length; i++) {
-				if (candidatePath.startsWith(projectPicks[i].detail)) {
-					if (belongingIndex < 0
-							|| projectPicks[i].detail.length > projectPicks[belongingIndex].detail.length) {
-						belongingIndex = i;
-					}
+	// pre-select an active project based on the uri candidate.
+	if (activeFileUri && activeFileUri.scheme === "file") {
```
CsCherrYY, line 721: activeFileUri?.scheme === "file"

src/commands.ts
jdneo: It's because I still reserve the possibility to let users configure whether it's a full build or incremental build. (the command accept an argument to specify that).

How about change it to `export const BUILD_PROJECT = 'java.project.build';` to remove the ambiguity?

CsCherrYY: sounds good. since this command also receives parameter `isFullBuild`.

jdneo:
@CsCherrYY PR updated. Meanwhile, I also make the first argument of the command either `Uri[]` or `Uri` to make it easier to call via `executeCommand()`

CsCherrYY (approved):
LGTM

PR #2554, opened 2022-07-12
Scan two levels of directories for activation indicators

### LINKED ISSUE #2280: Should support polyglot projects better
Java projects are not automatically detected when a polyglot project is opened in VS Code. For example, microservice app (e.g. https://github.com/elgris/microservice-app-example) usually places each service into a separate folder, see the structure below. Since the root folder doesn't have any build file, Java extension is not be activated automatically. Every time users have to open a Java file before using any Java feature, which is not convenient. 

```
microservice
├── frontend
│   ...
│   └── package.json
├── backend
│   ...
│   └── pom.xml
└── README
```

To support polyglot project better, we can relax the activation event a little bit. For example, allow detection of build files from direct subfolders. VS Code activation events support glob pattern, which we can implement using `workspaceContains:*/pom.xml`. The glob pattern `*/pom.xml` only scans two levels of directories and doesn't add much scanning cost.

```
    "workspaceContains:*/pom.xml",
    "workspaceContains:*/build.gradle",
    "workspaceContains:*/settings.gradle",
    "workspaceContains:*/build.gradle.kts",
    "workspaceContains:*/settings.gradle.kts",
```

jdneo:
fix #2280 

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,12 +40,12 @@
   ],
   "activationEvents": [
     "onLanguage:java",
-    "workspaceContains:pom.xml",
-    "workspaceContains:build.gradle",
-    "workspaceContains:settings.gradle",
-    "workspaceContains:build.gradle.kts",
-    "workspaceContains:settings.gradle.kts",
-    "workspaceContains:.classpath",
+    "workspaceContains:{*/,}pom.xml",
+    "workspaceContains:{*/,}build.gradle",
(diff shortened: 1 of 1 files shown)


testforstephen (approved):
LGTM

jdneo:
macos CI failed, looks like due to: https://github.com/microsoft/vscode-test/issues/153. We need to update `vscode-test` to `@vscode/test-electron`

Issue #2474, opened 2022-05-27
Support vscode-languageclient@8.0.1


testforstephen:
Starting from LanguageClient@8.0.1, the `BaseLanguageClient.start()` has been changed to return a promise instead of Dispose, and `onReady()` is removed, they're breaking changes to vscode-java. 

https://github.com/microsoft/vscode-languageserver-node/blob/release/client/8.0.1/client/src/common/client.ts#L933

We need to make changes to the client code in vscode-java to support the new behavior.

https://github.com/redhat-developer/vscode-java/blob/97b29fa68b7b5db8d05e805cbf6cc11b9d33efe5/src/standardLanguageClient.ts#L105

rgrunber:
Just for reference, we're likely going to need to wait on Theia to support things like TypeHierarchyItem. Currently if we run the extension there, it will fail in the vscode-langaugeclient initialization code. See https://github.com/redhat-developer/vscode-java/pull/2377 .

We can just update that PR when ready.

PR #2586, opened 2022-07-21
Added restart java language server command

AlexXuChen:
Added Restart Java Language Server command, `java.language.server.restart`.
Screen Recording 2022-08-05 at 8 14 19 PM

Signed-off-by: Alexander Chen <alchen@redhat.com>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,7 +5,7 @@ import * as fs from 'fs';
 import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import { CodeActionContext, commands, ConfigurationTarget, Diagnostic, env, EventEmitter, ExtensionContext, extensions, IndentAction, InputBoxOptions, languages, RelativePattern, TextDocument, UIKind, Uri, ViewColumn, window, workspace, WorkspaceConfiguration } from 'vscode';
+import { CodeActionContext, commands, ConfigurationTarget, Diagnostic, env, EventEmitter, ExtensionContext, extensions, IndentAction, InputBoxOptions, languages, RelativePattern, TextDocument, UIKind, Uri, ViewColumn, window, workspace, WorkspaceConfiguration, ProgressLocation } from 'vscode';
 import { CancellationToken, CodeActionParams, CodeActionRequest, Command, DidChangeConfigurationNotification, ExecuteCommandParams, ExecuteCommandRequest, LanguageClientOptions, RevealOutputChannelOn } from 'vscode-languageclient';
 import { LanguageClient } from 'vscode-languageclient/node';
 import { apiManager } from './apiManager';
@@ -328,6 +328,8 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			}));
 
 			context.subscriptions.push(onConfigurationChange(workspacePath, context));
+
+			registerRestartJavaLanguageServerCommand(context);
 
 			/**
 			 * Command to switch the server mode. Currently it only supports switch from lightweight to standard.
@@ -964,4 +966,25 @@ function registerOutOfMemoryDetection(storagePath: string) {
 		}
 		showOOMMessage();
 	});
+}
+
+function registerRestartJavaLanguageServerCommand(context: ExtensionContext) {
+	context.subscriptions.push(commands.registerCommand(Commands.RESTART_LANGUAGE_SERVER, async () => {
+		switch (getJavaServerMode()) {
+			case (ServerMode.standard):
+				// Standard server restart
+				await standardClient.getClient().restart();
+				break;
+			case (ServerMode.lightWeight):
+				// Syntax server restart
+				await syntaxClient.getClient().restart();
+				break;
+			case (ServerMode.hybrid):
+				if (syntaxClient.isAlive()) {
+					await syntaxClient.getClient().restart();
+				}
+				await standardClient.getClient().restart();
+				break;
+		}
+	}));
 }
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -14,7 +14,7 @@ import { Commands } from "./commands";
 import { ClientStatus } from "./extension.api";
 import * as fileEventHandler from './fileEventHandler';
 import { gradleCodeActionMetadata, GradleCodeActionProvider } from "./gradle/gradleCodeActionProvider";
-import { awaitServerConnection, prepareExecutable } from "./javaServerStarter";
+import { awaitServerConnection, prepareExecutable, DEBUG } from "./javaServerStarter";
 import { logger } from "./log";
 import { checkLombokDependency } from "./lombokSupport";
 import { markdownPreviewProvider } from "./markdownPreviewProvider";
@@ -106,7 +106,7 @@ export class StandardLanguageClient {
 		}
 
 		// Create the language client and start the client.
-		this.languageClient = new TracingLanguageClient('java', extensionName, serverOptions, clientOptions);
+		this.languageClient = new TracingLanguageClient('java', extensionName, serverOptions, clientOptions, DEBUG);
 
 		this.registerCommandsForStandardServer(context, jdtEventEmitter);
 		fileEventHandler.registerFileEventHandlers(this.languageClient, context);
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ The following commands are available:
 - `Java: List All Java Source Paths`: lists all the Java source paths recognized by the Java Language Server workspace.
 - `Java: Show Build Job Status`: shows the Java Language Server job status in Visual Studio Code terminal.
 - `Java: Go to Super Implementation`: goes to the super implementation for the current selected symbol in editor.
+- `Java: Restart Java Language Server`: restarts the Java language server.
 
 Supported VS Code settings
 ==========================
--- a/package.json
+++ b/package.json
@@ -1250,6 +1250,11 @@
         "command": "java.clean.sharedIndexes",
         "title": "%java.clean.sharedIndexes%",
         "category": "Java"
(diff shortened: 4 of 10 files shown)


AlexXuChen:
It seems the standard server won't start back up when the server mode is `HYBRID`

rgrunber:
> It seems the standard server won't start back up when the server mode is `HYBRID`

What happens if in the "Hybrid" case, you use `syntaxClient.isAlive()` and only stop/start it if true. Otherwise, just stop/start the standard client for now.

rgrunber:
`StandardLanguageClient.start()` does nothing because the underlying `languageClient` is null. It's null because `StandardLanguageClient.stop()` sets it that way.

Originally `stop()` was only really called on extension `deactivate()` so setting everything null was probably fine. Now, we want to stop the client but still re-use the connection so I would add a parameter to the `stop()`, something like `boolean cleanup`. It would be called with `false` for the restart command and `true` in `deactivate()`

AlexXuChen:
> Originally `stop()` was only really called on extension `deactivate()` so setting everything null was probably fine. Now, we want to stop the client but still re-use the connection so I would add a parameter to the `stop()`, something like `boolean cleanup`. It would be called with `false` for the restart command and `true` in `deactivate()`

Ah I see, because StandardLanguageClient.stop() is a wrapper for the LanguageClient.stop()

src/extension.ts
```diff
@@ -1065,3 +1066,25 @@ function registerOutOfMemoryDetection(storagePath: string) {
...
 	});
 }
+
+function registerRestartJavaLanguageServerCommand(context: ExtensionContext) {
+	context.subscriptions.push(commands.registerCommand(Commands.RESTART_LANGUAGE_SERVER, async () => {
+		// Can be replaced with `await languageClient.restart()` with vscode-languageclient ^8.0.1,
+		switch (getJavaServerMode()) {
+			case (ServerMode.STANDARD):
+				// Standard server restart
+				await standardClient.restart();
```
AlexXuChen, line 1076: Instead of using the wrapper start(), stop(), I instead created a restart for both the standardClient and syntaxClient. This way, the languageClient won't be set to null and the ClientStatus won't change.

src/standardLanguageClient.ts
```diff
@@ -591,6 +591,16 @@ export class StandardLanguageClient {
 		return Promise.resolve();
 	}
 
+	public async restart(): Promise<void> {
+		await this.languageClient.stop();
+		if (this.languageClient.isInDebugMode) {
```
rgrunber, line 596: Is `isInDebugMode` `true` in debug mode ? I would debug this could to see if that happens. Note that there is an exported `DEBUG` variable you can also use, much like there was for the same change in vscode-xml.

rgrunber:
@testforstephen , you mentioned there's already an issue, or PR for this that's incoming ? I wasn't able to find any issue on the vscode-java-debug end.

testforstephen:
> @testforstephen , you mentioned there's already an issue, or PR for this that's incoming ? I wasn't able to find any issue on the vscode-java-debug end.

@rgrunber I tried this PR and Java debugger works fine after a language server restart. The reason is that the current restart implementation does not trigger a server mode change event and has no effect on the debugger.

Previously we saw some side effects when restarting because that implementation would destroy the languageClient and recreate a new one, which would cause the Java extension to resend the server mode change event.

However, I found another issue that the status icon is always spinning after language server restart.
image

datho7561:
I just tried this PR with vscode-microprofile, which provides a eclipse.jdt.ls extension and a language server that communicates with the eclipse.jdt.ls extension using language server commands. Restarting works very well; as soon as the server restarts, all the microprofile features start working again. The only issue I ran into while doing this is that the microprofile language server sends sends error notifications while eclipse.jdt.ls is down. This makes sense, since the microprofile language server is trying to make requests to eclipse.jdt.ls while it's not running. However, it might confuse the end user, since they might think something is wrong. It would be nice to prevent these errors from appearing, perhaps with an API to detect if the server is restarting, but I think that could wait to a future PR.

rgrunber:
I'm still seeing the build status provider spinning even after the server has been restarted and is working normally. I think i had one or two instances where it didn't spin, but more often than not, it doesn't behave correctly. Looks like this has to do with whether the language server is able to send back the `Shutdown...` or  progress report. Since the shutdown message is represented as incomplete, it results in the server status being set as busy, and since the project isn't re-imported, the last state will always be busy. One workaround we could do is to make sure the restart command sets the server status as ready after the restart is complete :

```diff
diff --git a/src/extension.ts b/src/extension.ts
index d76364c..c999912 100644
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1074,6 +1074,7 @@ function registerRestartJavaLanguageServerCommand(context: ExtensionContext) {
                        case (ServerMode.STANDARD):
                                // Standard server restart
                                await standardClient.restart();
+                               serverStatusBarProvider.setReady();
                                break;
                        case (ServerMode.LIGHTWEIGHT):
                                // Syntax server restart
@@ -1087,6 +1088,7 @@ function registerRestartJavaLanguageServerCommand(context: ExtensionContext) {
                                        await new Promise((resolve) => setTimeout(resolve, 3000));
                                }
                                await standardClient.restart();
+                               serverStatusBarProvider.setReady();
                                break;
                }
        }));
```

AlexXuChen:
Server status should no longer be busy.

src/extension.ts
```diff
@@ -167,6 +167,7 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 	context.subscriptions.push(commands.registerCommand(Commands.NOT_COVERED_EXECUTION, async () => {
 		markdownPreviewProvider.show(context.asAbsolutePath(path.join('document', `_java.notCoveredExecution.md`)), 'Not Covered Maven Plugin Execution', "", context);
 	}));
+	registerRestartJavaLanguageServerCommand(context);
```
rgrunber, line 170: This should probably be moved to https://github.com/redhat-developer/vscode-java/blob/821ba94145c9d983d8f53a90e70627029fea030d/src/extension.ts#L384 with the other commands. I would want to avoid it being used prior to there being any kind of client object.

rgrunber (commented):
@testforstephen , let me know if this works for you now. I think the last thing to test is specifically how it interacts when one is using vscode-java-debug or anything else.

To a large extent, I don't think we can easily avoid dependent extensions throwing errors while the restart is occuring.

testforstephen:
The message id doesn't exist in package.nls.json.
image

> let me know if this works for you now. I think the last thing to test is specifically how it interacts when one is using vscode-java-debug or anything else.

Tried again with the latest commit and if there is a debug session running, the restart operation will terminate the debugger. Once the server is ready, the debugging feature works again. It doesn't break the Java debugger, which is good for me.

Just one concern is about the transparency of the operation. When I execute "Restart Java Language Server", there is currently no indicator to tell me what is happening in the background and when the server will be ready again. I'm wondering whether it's better to show a progress notification on restart opreation.

rgrunber:
The only other issue I see is that in syntax client mode (`"java.server.launchMode": "LightWeight"`), I get an error indicating the connection was disposed every time I trigger the language server restart command, and the server fails to restart. If I call the same command while the language server is down, it does start back up.

> Just one concern is about the transparency of the operation. When I execute "Restart Java Language Server", there is currently no indicator to tell me what is happening in the background and when the server will be ready again. I'm wondering whether it's better to show a progress notification on restart opreation.

This could probably be done with ` window.withProgress(..)`, (which we use in a few places). It would at least indicate how long the restart operation takes but to have the progress disappear exactly when the server is back up (rather when the restart operation finishes) might be tricky. Maybe listening for **any** notification after the restart would be indication it has returned and then disposing of that `onNotification` listener might work.

rgrunber:
In the process of trying to make the progress notification disappear once the server is responding again (as opposed to once the restart command has finished being issued), I've discovered that once the language server is restarted, it fails to receive notifications on **any** of the `onNotification(...)` callbacks. That's... not good.

Although this is a deal breaker, it's actually https://github.com/microsoft/vscode-languageserver-node/issues/770, which was fixed in 8.0. So we now depend on https://github.com/redhat-developer/vscode-java/issues/2474 in order to merge this change. Hopefully we'll be able to do that pretty soon.

JessicaJHee:
As a side note, I run into the following problem while in debug mode sometimes:
```
ERROR: transport error 202: bind failed: Address already in use
ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [../src/jdk.jdwp.agent/share/native/libjdwp/debugInit.c:734]
```

The problem is solved with the timeout set to 2000 instead of 1000 here:
https://github.com/microsoft/vscode-languageserver-node/blob/ffaf05ddd1fa9ca5f258cd0b08cc6cd1794c5b57/client/src/node/main.ts#L197

rgrunber:
In the previous iteration of this PR, we had it set at 2000 (I suspect I saw issues with 1000). However now it's handled by the library so we don't really have control. It's mainly an annoying problem for developers because in production those debug ports aren't open though.

Worst case, we could just do what we did before. Copy the `restart()` logic and do it ourselves (replacing 1000 -> 2000). I'll try and see how bad it is.

rgrunber:
This works really well for me with the eclipse.jdt.ls modification to the Shutdown job. I've tried restarting with various launch modes. I noticed that if no server is running (eg. invisible project with no files opened, or empty workspace), one can still restart the language server (though it never started once to begin with). Have a look at the `"when": "javaLSReady"` context and let me know if that makes sense for such cases.

rgrunber (approved):
I would just add your info as a `Co-authored-by:` and I think we can merge.

PR #2594, opened 2022-07-25
Make the wording in notifications consistent

### LINKED ISSUE #2585: Keep consistent style in notification when prompted to restart/reload VSCode
##### Environment
- Operating System: Mac intel
- JDK version: 17.0.2
- Visual Studio Code version: 1.69.2
- Java extension version: 1.9.0

##### Steps To Reproduce
1. [step 1] Change `java.jdt.ls.java.home` in settings > Notification dialog shows as screenshot 1
2. [step 2] Install a new extension, like debugger for java > Notification dialog shows as screenshot 2

##### Current Result
Screenshot 1:
image
Screenshot 2:
image

##### Expected Result
As restart and reload work in a same way, better to only keep one, both reload or both restart.

jdneo:
fix #2585 

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -720,7 +720,7 @@ function makeRandomHexString(length) {
 
 async function cleanWorkspace(workspacePath, force?: boolean) {
 	if (!force) {
-		const doIt = 'Restart and delete';
+		const doIt = 'Reload and delete';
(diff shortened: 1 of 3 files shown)


jdneo:
Thanks for pointing out that. Let's make them align.

Except for the last one, since it's about the language server not the vs code window, so maybe using restart is fine here?

rgrunber (approved):
I guess we can leave the wording in the crash logic since restart makes more sense to me. Feel free to squash and merge when ready.

PR #2612, opened 2022-08-03
Use plural form for reload projects command title

jdneo:
Use plural form here since now we support reload multiple projects when multiple projects exist in workspace.


rgrunber (approved):
Looks good. Does this also apply to https://github.com/redhat-developer/vscode-java/blob/184111d90e8822af1e37cc0ba4292514da384790/src/gradle/gradleCodeActionProvider.ts#L33-L35 ?

Either way, feel free to merge if it doesn't, or once you modify the PR to address that also.

jdneo:
Thanks for notifying that. Let's keep singular for that case since it's updating the targeting project under that context.

PR #2614, opened 2022-08-05
Support "Sort Members" Code Action

CsCherrYY:

requires https://github.com/eclipse/eclipse.jdt.ls/pull/2169

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ The following settings are supported:
 
 New in 1.10.0
 * `java.import.maven.offline.enabled` : Enable/disable the Maven offline mode. Defaults to `false`.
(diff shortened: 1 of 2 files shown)


rgrunber:
Should we use a description that explains why this is needed as opposed to just what it will do ?

`Reordering of fields, enum constants, and initializers can result in semantic and runtime changes due to different initialization and persistence order. Enabling this setting prevents certain re-ordering from taking place.`

CsCherrYY:
Sure. Let me update the configuration name and discription.

rgrunber:
Only thing left is to add this in the README at the bottom of the settings as `New in 1.10.0`. Looks good otherwise.

rgrunber (approved):

PR #2624, opened 2022-08-11
Make the setting 'java.import.gradle.java.home' to be machine-overridable

testforstephen:

We have an enterprise partner that is trying to adopt VS Code as their Java development tool. To help their team to on board to VS Code more smoothly, they developed an internal command line tool to generate Java settings in .vscode/settings.json by one-click. They hope all JDK related settings to be workspace changable. This is a reasonable idea since workspace-level settings are more easily customized by third-party tools.

In addition, we already do that for most JDK-related settings such as 'java.jdt.ls.java.home', 'java.configuration.runtimes', it makes sense for me to change `java.import.gradle.java.home` to machine-overridable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
         "java.jdt.ls.java.home",
         "java.home",
         "java.jdt.ls.vmargs",
-        "java.configuration.runtimes"
+        "java.configuration.runtimes",
+        "java.import.gradle.java.home"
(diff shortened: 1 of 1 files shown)


jdneo (approved):

Eskibear:
it makes sense to me.

PR #2629, opened 2022-08-16
Automatically add Protobuf output source directories to clsspath

jdneo:
More details, please check https://github.com/eclipse/eclipse.jdt.ls/pull/2189

### DIFF
--- a/package.json
+++ b/package.json
@@ -917,7 +917,7 @@
           "description": "Whether to load lombok processors from project classpath",
           "scope": "window"
         },
-        "java.protobufSupport.enabled": {
+        "java.jdt.ls.protobufSupport.enabled": {
           "type": "boolean",
           "default": true,
           "markdownDescription": "Specify whether to automatically add Protobuf output source directories to the classpath.\n\n**Note:** Only works for Gradle `com.google.protobuf` plugin `0.8.4` or higher.",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -917,6 +917,12 @@
           "description": "Whether to load lombok processors from project classpath",
           "scope": "window"
         },
+        "java.protobufSupport.enabled": {
```
jdneo, line 920: Another candidate for the setting name is `java.jdt.ls.protobufSupport.enabled`, which will be aligned with `java.jdt.ls.lombokSupport.enabled`.

snjeza (approved):

package.json
rgrunber: Agreed. I even had a pending comment on the JDT-LS change that read :

>One other thing is, should we called this java.jdt.ls.protobufSupport.enabled so it matches the pattern of things like java.jdt.ls.lombokSupport.enabled.

>I guess if we knew we'd be doing more of this kind of support, we might have created it under java.jdt.ls.support.${feature}.enabled but the name I mention above is fine for now.

:joy:

jdneo: Let's call it `java.jdt.ls.protobufSupport.enabled` then. 😁

PR #2633, opened 2022-08-17
Skip default project when detecting lombok

jdneo:

### DIFF
--- a/src/lombokSupport.ts
+++ b/src/lombokSupport.ts
@@ -10,6 +10,7 @@ import { apiManager } from "./apiManager";
 import { supportsLanguageStatus } from "./languageStatusItemFactory";
 import { runtimeStatusBarProvider } from './runtimeStatusBarProvider';
 import { logger } from './log';
+import { getAllJavaProjects } from "./utils";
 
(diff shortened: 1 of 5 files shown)


rgrunber:
Change looks good to me so can definitely be merged.

However I noticed a few instances in the where we verify if one of the project URIs (often from `java.project.getAll`) is the default java project :

https://github.com/redhat-developer/vscode-java/blob/4a1b23dabdfd03cee0324f2d8487df9b5a9d3911/src/standardLanguageClient.ts#L612-L613

https://github.com/redhat-developer/vscode-java/blob/4a1b23dabdfd03cee0324f2d8487df9b5a9d3911/src/runtimeStatusBarProvider.ts#L243

https://github.com/redhat-developer/vscode-java/blob/4a1b23dabdfd03cee0324f2d8487df9b5a9d3911/src/standardLanguageClientUtils.ts#L94-L95

It might be nice to provide some utility method that checks this.

jdneo:
> However I noticed a few instances where we verify if one of the project URIs (often from java.project.getAll) is the default java project

Let me append one more commit in this PR to extract the utility.

testforstephen (approved):
LGTM

PR #2644, opened 2022-08-24
Support Android project

CsCherrYY:

related to https://github.com/eclipse/eclipse.jdt.ls/pull/2197

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,7 @@ New in 1.10.0
 * `java.import.maven.offline.enabled`: Enable/disable the Maven offline mode. Defaults to `false`.
 * `java.codeAction.sortMembers.avoidVolatileChanges`: Reordering of fields, enum constants, and initializers can result in semantic and runtime changes due to different initialization and persistence order. This setting prevents this from occurring. Defaults to `true`.
 * `java.jdt.ls.protobufSupport.enabled`: Specify whether to automatically add Protobuf output source directories to the classpath. **Note:** Only works for Gradle `com.google.protobuf` plugin `0.8.4` or higher. Defaults to `true`.
+* `java.jdt.ls.androidSupport.enabled`: Specify whether to automatically import Android project. **Note:** Only works for Android Gradle Plugin `3.2.0` or higher. Defaults to `true`.
 
 Semantic Highlighting
 ===============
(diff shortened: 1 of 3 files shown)


fbricon:
At this stage we should probably label android support as "experimental"

CsCherrYY:
> At this stage we should probably label android support as "experimental"

Agree. I'll change the description later.

package.json
```diff
@@ -924,6 +924,12 @@
           "markdownDescription": "Specify whether to automatically add Protobuf output source directories to the classpath.\n\n**Note:** Only works for Gradle `com.google.protobuf` plugin `0.8.4` or higher.",
           "scope": "window"
         },
+        "java.jdt.ls.androidSupport.enabled": {
```
jdneo, line 927: Let's have 3 options here:

auto: Only enabled in VS Code Insider, and this is the default value.
on: ...
off: ...

If User selects auto, based on what client is used, pass true/false to the server side.

CsCherrYY: added in https://github.com/redhat-developer/vscode-java/pull/2644/commits/1cb4745de7dcbfb1cfd5e7055ef29cbbf5783837

package.json
```diff
@@ -924,6 +924,17 @@
...
+          "type": "string",
+          "enum": [
+            "auto",
+            "on",
+            "off"
+          ],
+          "default": "auto",
+          "markdownDescription": "[Experimental] Specify whether to enable Android project importing. When set to `auto`, the Android support will be enabled in VS Code insider and disabled in VS Code stable.\n\n**Note:** Only works for Android Gradle Plugin `3.2.0` or higher.",
```
jdneo, line 935: Suggest not mentioning VS Code stable. Since the vsix might be installed in other clients.

And note the official name of insider. It's `Visual Studio Code - Insiders`

README.md
```diff
@@ -210,6 +208,7 @@ New in 1.10.0
 * `java.import.maven.offline.enabled`: Enable/disable the Maven offline mode. Defaults to `false`.
 * `java.codeAction.sortMembers.avoidVolatileChanges`: Reordering of fields, enum constants, and initializers can result in semantic and runtime changes due to different initialization and persistence order. This setting prevents this from occurring. Defaults to `true`.
 * `java.jdt.ls.protobufSupport.enabled`: Specify whether to automatically add Protobuf output source directories to the classpath. **Note:** Only works for Gradle `com.google.protobuf` plugin `0.8.4` or higher. Defaults to `true`.
+* `java.jdt.ls.androidSupport.enabled`: [Experimental] Specify whether to enable Android project importing. When set to `auto`, the Android support will be enabled in VS Code insider and disabled in VS Code stable. **Note:** Only works for Android Gradle Plugin `3.2.0` or higher. Defaults to `auto`.
```
jdneo, line 211: The description here also needs to be updated

CsCherrYY: updated in https://github.com/redhat-developer/vscode-java/pull/2644/commits/0a277212842c5b70d4eab3646527de1806ecfd95

jdneo (approved):

akash07k:
@CsCherrYY  Does it support Kotlin based Android projects too or java only? and, are there any steps which we need to follow for enabling the full support?
I'm getting classpath error while launching the app.

CsCherrYY:
@akash07k it's for Java only, since the vscode-java extension can only provide language features for java files. 

> and, are there any steps which we need to follow for enabling the full support? 

only set `java.jdt.ls.androidSupport.enabled` to `on` if you're using VS Code stable version. If you're using VS Code insider version, nothing is needed. If you meet further problems, please open an issue for that.

akash07k:
Ok, thanks.

Issue #2577, opened 2022-07-19
Extension never activates, throwing error "Header must provide a Content-Length property"


atcericandrews:
As per the title, the extension never activates; the java language server status never switches to the thumbs-up. The extension runtime status shows 37 copies of the same uncaught error: `Header must provide a Content-Length property.` The developer tools logs show the same error, slightly more verbosely:

```
workbench.desktop.main.js:2272 Error: Header must provide a Content-Length property.
	at p.onData (c:\Users\Eric.Andrews\.vscode\extensions\redhat.java-1.8.0-win32-x64\dist\extension.js:2:1138205)
	at Socket.<anonymous> (c:\Users\Eric.Andrews\.vscode\extensions\redhat.java-1.8.0-win32-x64\dist\extension.js:2:1137926)
	at Socket.emit (node:events:390:28)
	at addChunk (node:internal/streams/readable:315:12)
	at readableAddChunk (node:internal/streams/readable:289:9)
	at Socket.Readable.push (node:internal/streams/readable:228:10)
	at Pipe.onStreamRead (node:internal/stream_base_commons:199:23)
```

##### Environment
- Operating System: Windows 10 Enterprise (10.0.19044)
- JDK version: Eclipse Adoptium 17.0.3.7-hotspot
- Visual Studio Code version: 1.69.2
- Java extension version: 1.8.0

##### Steps To Reproduce
1. Install jdk and VSCode
2. Install the Extension Pack for Java
3. Update settings.json to point to jdk: `"java.jdt.ls.java.home": "C:\\Program Files\\Eclipse Adoptium\\jdk-17.0.3.7-hotspot"`
4. Open any .java file

[language server logs](https://github.com/redhat-developer/vscode-java/files/9142497/language-server-log.txt)
[dev tools console logs](https://github.com/redhat-developer/vscode-java/files/9142505/devtool-logs.txt)

devin-lo:
same issue on:

VS code versions 1.68.1 and 1.67.0
RedHat Java extension 1.7.0, 1.8.0, 1.90
(multiple combinations of the above were tried, all had same issue)

JDK version 11.0.12 (oracle)

elizalde1981:
Seems to be similar to #2292

victorlee0505:
same issue here, out of the blue it just stopped working.

Foxhunter981:
Also happening to me. Its very annoying. I have to switch to Eclipse for now 
[...]

PR #2646, opened 2022-08-25
GH2577: Add '-Xlog:jni+resolve=off' to the default vmargs

testforstephen:

Mitigate the issue https://github.com/redhat-developer/vscode-java/issues/2577#issuecomment-1208922149. The parameter `-Xlog:jni+resolve=off` is used to prevent some unexpected jni logs from being logged to stdout, because Java language server uses stdio as the jsonrpc channel by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,8 +196,8 @@
             "string",
             "null"
           ],
-          "default": "-XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true -Xmx1G -Xms100m",
-          "description": "Specifies extra VM arguments used to launch the Java Language Server. Eg. use `-XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true -Xmx1G -Xms100m ` to optimize memory usage with the parallel garbage collector",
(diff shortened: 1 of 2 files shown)


fbricon:
I think you should also add it dynamically if necessary (like we do for logging or jpms modules), in case user already uses custom VM args.

testforstephen:
> I think you should also add it dynamically if necessary (like we do for logging or jpms modules), in case user already uses custom VM args.

done.

rgrunber (approved):
Looks fine to me. I did some quick research and noticed there's `-Xlog:disable` also. If the language server is communicating via std streams, there's probably a good argument to make for disabling logging completelyin the JVM when the client/server are using the standard streams, but maybe we can investigate that after.

It would be interesting to know if there's older issues with similar failure.

Aankhen:
> Looks fine to me. I did some quick research and noticed there's `-Xlog:disable` also.

Just wanted to note that this fixed the issue for me (Windows 10, VS Code 1.72.2, Java 17) when the other version didn’t.

rgrunber:
> > Looks fine to me. I did some quick research and noticed there's `-Xlog:disable` also.
> 
> Just wanted to note that this fixed the issue for me (Windows 10, VS Code 1.72.2, Java 17) when the other version didn’t.

This should be the default now. After discovering other issues that seemed very related to this, I thought there might be other JVM components that use the logging facility, so best to disable it entirely ( https://github.com/redhat-developer/vscode-java/pull/2690 )

PR #2674, opened 2022-09-13
Add buildship auto sync preference

### LINKED ISSUE #2673: Add buildship auto sync preference
hello everyone, 
**TLDR**: anyone knows how to control the `.settings/org.eclipse.buildship.core.prefs` > `auto.sync` flag in VSC, without touching that file specifically? 
(...to make it 'true', as it was false by default for me and at the origin of my problem explained below..) 
*************************

I'm working on a *gradle / java / spring-boot* project using vsc (with '*extension pack for java*' + '*gradle for java*' extensions installed to name a few..), and I've been losing hours trying to figure out how updating my `build.gradle` file by adding/removing dependencies can automatically update my vsc java coding context so that I can have access to those libs in my code, more precisely, how can the `Java Projects > ... > Project and External Dependencies` UI get updated automatically when I update and save new dependencies in my `build.gradle` file...

After hours of research, I managed to find the setting `java.import.generatesMetadataFilesAtProjectRoot` (from this [link](https://github.com/eclipse/eclipse.jdt.ls/pull/1900)) , I turned it on, and then discovered the new generated file `.settings/org.eclipse.buildship.core.prefs` in my project root.

One of its setting captured my attention, e.g. `auto.sync=false`, I wondered what if I turn it on and try just for fun, and boom, it solved my problem!

Now, with `.settings/org.eclipse.buildship.core.prefs` > `auto.sync=true`, I can finally update my `build.gradle` file, add/remove dependencies, save the file, and finally (actually, even if I don't answer "Yes" to the notification asking if I want to synchronize my java classpath/configuration and let the notif. displayed for a little while..), I will see the dependencies changes being reflected in vs code, e.g. changes being reflected under `Java Projects` tab of the explorer under its `Project and External Dependencies`, which directly means, I can finally use those libs in my java code (before that, I couldn't, the libs were not recognized / could not be imported by vs code, and my only way around was to do a `Java > Clean Java Language Server Workspace`, and get vs code restarted, which was super annoying for every single gradle dependency that I was playing with, lots of time lost restarting vsc all the time, generating frustration as usual etc..).

Now my simple question is, what is the "official" way of getting the `auto.sync` flag to be `true`, without having to manually modify that `.settings/org.eclipse.buildship.core.prefs` file myself? As I understand it, we are not supposed to play with that file directly, as it gets generated by the Java Language Server extension, based on some other external settings/inputs?

Any help much appreciated, thanks!

##### Environment
- Operating System: macOS big sur (11.6.5)
- JDK version: openjdk version "17.0.4.1" 2022-08-12
- Visual Studio Code version: 1.71.0
- Java extension version: 1.10.0

##### Steps To Reproduce
1. **Install** `VS Code` and at least the following extensions: `Extension pack for Java` + `Gradle for Java`.
2. **Create** a new simple gradle java project (the `gradle for java` extension can help you with that, e.g. use the vsc command `Gradle: Create a Gradle Java Project... (Advanced)`.
3. **Add** new gradle dependencies in your `build.gradle` file, and **Save** the file changes.
4. When asked by a vsc notification if you want to synchronize the java classpath/configuration, say **Yes**.
5. **Go** to `Explorer` > `Java Projects` > `[...your app module...]` > `Project and External Dependencies`, and **expend** the list.

Here's a simple `build.gradle` file that I used to reproduce the problem described above (but the exact same file will work with that `auto.sync=true` though..):
```
plugins {
    // Apply the application plugin to add support for building a CLI application in Java.
    id 'application'
    // Apply the java plugin
    id 'java'
}

sourceCompatibility = '17'

repositories {
    // Use Maven Central for resolving dependencies.
    mavenCentral()
}

dependencies {
    implementation 'org.apache.commons:commons-lang3:3.0'
}
    
application {
    // Define the main class for the application.
    mainClass = 'com.demo.App'
}
```

##### Current Result
Notice that the gradle dependencies added at `step 3` are not listed under `step 5`, e.g. Gradle and VSC (vscode-java e.g. the Java Language Server) remain out of sync in terms of dependencies.
Consequence of the above is that those dependencies/libs won't be available in my java code, so I'm unable to import any of them or else I'll get java compiling errors.

##### Expected Result
Gradle and VSC/vscode-java have my dependencies automatically synchronized.
Consequence of the above is that those dependencies/libs BECOME automatically available in my java code, so I'm able to import any of them without any java compiling errors.

##### Additional Informations
One workaround to fix the described situation and ensure `step 5` above would list the added dependencies of `step 3` above, was to execute the vsc command `Java > Clean Java Language Server Workspace`, but that is super annoying to do, every single time you have to modify your gradle dependencies (add/remove), like in `step 3` above, since it requires you to restart your entire editor/vsc, and it then takes a little while for it to be back opened and ready, since the java language server will take a little while to reload, re-index everything etc. But at that point, it would at least be finally in sync with my `build.gradle` file. 

Obviously, the other solution that I found around the `.settings/org.eclipse.buildship.core.prefs` > `auto.sync=true` is much much better and actually fixes everything for me, but I'm looking for the right way to fix this through any other vs code settings, rather than modifying directly that file myself, if at all possible.

snjeza:
Fixes #2673 
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2224

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,12 +203,13 @@ The following settings are supported:
 * `java.configuration.workspaceCacheLimit` : The number of days (if enabled) to keep unused workspace cache data. Beyond this limit, cached workspace data may be removed.
 * `java.import.generatesMetadataFilesAtProjectRoot` : Specify whether the project metadata files(.project, .classpath, .factorypath, .settings/) will be generated at the project root. Defaults to `false`.
 * `java.inlayHints.parameterNames.enabled`: Enable/disable inlay hints for parameter names. Supported values are: `none`(disable parameter name hints), `literals`(Enable parameter name hints only for literal arguments) and `all`(Enable parameter name hints for literal and non-literal arguments). Defaults to `literals`.
-
-New in 1.10.0
 * `java.import.maven.offline.enabled`: Enable/disable the Maven offline mode. Defaults to `false`.
 * `java.codeAction.sortMembers.avoidVolatileChanges`: Reordering of fields, enum constants, and initializers can result in semantic and runtime changes due to different initialization and persistence order. This setting prevents this from occurring. Defaults to `true`.
 * `java.jdt.ls.protobufSupport.enabled`: Specify whether to automatically add Protobuf output source directories to the classpath. **Note:** Only works for Gradle `com.google.protobuf` plugin `0.8.4` or higher. Defaults to `true`.
 * `java.jdt.ls.androidSupport.enabled`: [Experimental] Specify whether to enable Android project importing. When set to `auto`, the Android support will be enabled in Visual Studio Code - Insiders. **Note:** Only works for Android Gradle Plugin `3.2.0` or higher. Defaults to `auto`.
+
+New in 1.11.0
+* `java.import.gradle.auto.sync`: Automatically start project synchronization in the background when the build script changes. Defaults to `false`.
 
 Semantic Highlighting
 ===============
--- a/package.json
+++ b/package.json
@@ -280,6 +280,12 @@
           "description": "Enable/disable the Gradle importer.",
           "scope": "window"
         },
+        "java.import.gradle.auto.sync": {
+          "type": "boolean",
+          "default": false,
+          "description": "Automatically start project synchronization in the background when the build script changes.",
+          "scope": "window"
+        },
         "java.import.gradle.wrapper.enabled": {
           "type": "boolean",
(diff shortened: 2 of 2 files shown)


snjeza:
See https://github.com/eclipse/eclipse.jdt.ls/pull/2224#issuecomment-1247480825

PR #2677, opened 2022-09-15
Enable annotation-based null analysis

CsCherrYY:

related to https://github.com/eclipse/eclipse.jdt.ls/pull/2228


CsCherrYY:
The workflow is:

1. project imported
2. check corresponding jar in the classpath
3. if exists, update compile options at the server side

since changing compile options needs re-compile to make it effective, I call compile workspace here: https://github.com/redhat-developer/vscode-java/pull/2677/files#diff-c894be06f39638b5cd4094b12af0ae3effd7de5bc75b6a4bdf43d3b283168b65R37

This call is on demand since we will only call this command when the null analysis status has changed. For users don't have related annotation jar on their classpath, it will not execute.

rgrunber:
Thanks @CsCherrYY , you read my mind. I was going to comment on whether the classpath check could be done purely on the server-side to simplify this. My main concern was whether you did it here to avoid a performance issues in the `preferencesChanged` callback, which shouldn't be doing heavy computations.

CsCherrYY:
> My main concern was whether you did it here to avoid performance issues in the preferencesChanged callback, which shouldn't be doing heavy computations.

Yeah, we can do things at the client side can make `preferencesChanged` callback not block the server, but the server still has to check the classpath and full build the affected project to make it work. Currently I just move all the logics to the server side as we discussed (https://github.com/eclipse/eclipse.jdt.ls/pull/2228/files). It will check the classpath when 

- project imported
- null analysis configurations changed
- classpath changed

and if the compiler options in specific Java project (when classpath changed) or JavaCore (when null analysis configurations changed) changed, we trigger a full build for the project or the workspace. Is there any place to enhance?

package.json
```diff
@@ -958,6 +958,26 @@
...
           "scope": "window"
+        },
+        "java.compile.nullAnalysis.nonnull": {
+          "type": "array",
+          "default": [
+            "javax.annotation.Nonnull",
+            "org.eclipse.jdt.annotation.NonNull",
+            "org.springframework.lang.NonNull"
+          ],
+          "markdownDescription": "Specify the nonnull annotation types used by the annotation-based null analysis. Leave it empty to disable the null analysis.",
```
testforstephen, line 969: Also mention that the top of this list will be used first. Same for nullable config.

`
Specify the NonNull annotation to be used for null analysis. If more than one annotation is specified, then the topmost annotation will be used first if it exists in your project dependencies.
`

CsCherrYY: Yeah, let me correct it.

testforstephen (approved):

PR #2685, opened 2022-09-20
Document import order for static members.

### LINKED ISSUE #711: Static imports order
The _organize imports_ feature puts the static imports at the beginning of the _imports_ block (just after package).
It would be nice to have an option that gives you the choice to put it either at the beginning or at the end.

##### Environment
- Operating System: Fedora 29
- JDK version: 1.8.0_191
- Visual Studio Code version: 1.28.2
- Java extension version: 0.33.0

##### Steps To Reproduce
1. Add some static imports before the class definition
2. Fire organize imports

rgrunber:
- Fixes #711

I considered also updating the description on the README.md page, but it seems we decided to keep that short.

### DIFF
--- a/package.json
+++ b/package.json
@@ -550,12 +550,14 @@
         },
         "java.completion.importOrder": {
           "type": "array",
-          "description": "Defines the sorting order of import statements. A package or type name prefix (e.g. 'org.eclipse') is a valid entry. An import is always added to the most specific group.",
+          "description": "Defines the sorting order of import statements. A package or type name prefix (e.g. 'org.eclipse') is a valid entry. An import is always added to the most specific group. As a result, the empty string (e.g. '') can be used to group all other imports. Static imports are prefixed with a '#'",
           "default": [
+            "#",
             "java",
             "javax",
             "org",
-            "com"
+            "com",
+            ""
           ],
           "scope": "window"
         },


package.json
```diff
@@ -550,7 +550,7 @@
         },
         "java.completion.importOrder": {
           "type": "array",
-          "description": "Defines the sorting order of import statements. A package or type name prefix (e.g. 'org.eclipse') is a valid entry. An import is always added to the most specific group.",
+          "description": "Defines the sorting order of import statements. A package or type name prefix (e.g. 'org.eclipse') is a valid entry. An import is always added to the most specific group. Static imports are supported, and must be prefixed with a '#'",
           "default": [
```
testforstephen, line 554: Can we tweak the default value a little bit to make it more transparent? And also mention `""` represents unknown category in description?

```
"default": [
            "#",
            "java",
            "javax",
            "org",
            "com",
            ""
          ],
```

rgrunber: Sure. Though we may as well synchronize with https://github.com/eclipse/eclipse.jdt.ls/blob/3d1a42c80abb12357a1152a097ec63b14ea5068c/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/Preferences.java#L596-L600 so it's the default for all clients also.

I remember @jeffmaury mentioned it would be good to add `io` as well. Could that be added after `com` ?

rgrunber: I've decided to leave this out for now as it's tough to find any standard on where to place it, let alone a place where other configurations are using it. Should be easy enough for users to define for themselves.

PR #2703, opened 2022-09-27
Format content pasted into string literals

### LINKED ISSUE #1249: paste a multi-line string should create a java string
try paste this into a string, i.e. between two quotes "":

```
SELECT payment_date, amount, sum(amount) OVER (ORDER BY payment_date)
FROM (
  SELECT CAST(payment_date AS DATE) AS payment_date, SUM(amount) AS amount
  FROM payment
  GROUP BY CAST(payment_date AS DATE)
) p
ORDER BY payment_date;
```

in eclipse it becomes a valid java string in vscode it generates non-compilable code.

Would be great if vscode could help here :)

datho7561:
Testing this PR requires using `vscode-insiders` (until VS Code 1.74.0 is released).

Uses the proposed API to modify pasted strings. It invokes jdt.ls's `java.edit.handlePasteEvent` command when vscode-java receives a paste event. When pasting into a string literal, jdt.ls modifies the content that's inserted so that the string content reflects the original value of the copied text. It uses string concatenation instead of the newer Java multiline string feature to represent multiline text.

Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2349
Closes https://github.com/redhat-developer/vscode-java/issues/1249

### DIFF
--- a/src/pasteEventHandler.ts
+++ b/src/pasteEventHandler.ts
@@ -1,0 +1,102 @@
+import { CancellationToken, commands, DataTransfer, DocumentPasteEdit as VDocumentPasteEdit, DocumentPasteEditProvider, DocumentPasteProviderMetadata, ExtensionContext, languages, Range, TextDocument, window } from "vscode";
+import { FormattingOptions, Location, WorkspaceEdit as PWorkspaceEdit } from "vscode-languageclient";
+import { LanguageClient } from "vscode-languageclient/node";
+import { Commands } from "./commands";
+import { JAVA_SELECTOR } from "./standardLanguageClient";
+
+const TEXT_MIMETYPE: string = "text/plain";
+const MIMETYPES: DocumentPasteProviderMetadata = {
+	pasteMimeTypes: [TEXT_MIMETYPE]
+};
+
+/**
+ * Parameters for `Commands.HANDLE_PASTE_EVENT`
+ */
+interface PasteEventParams {
+	location: Location;
+	text: string;
+	formattingOptions: FormattingOptions;
+	copiedDocumentUri?: string;
+}
+
+/**
+ * Response from jdt.ls for `Commands.HANDLE_PASTE_EVENT` that's similar, but not identical, to VS Code's paste edit.
+ *
+ * @see VDocumentPasteEdit
+ */
+interface PDocumentPasteEdit {
+	insertText: string;
+	additionalEdit: PWorkspaceEdit;
+}
+
+/**
+ * Registers the vscode-java DocumentPasteEditProviders and sets them up to be disposed.
+ *
+ * @param context the extension context
+ */
+export function registerPasteEventHandler(context: ExtensionContext, languageClient: LanguageClient) {
+	if (languages["registerDocumentPasteEditProvider"]) {
+		context.subscriptions.push(languages["registerDocumentPasteEditProvider"](JAVA_SELECTOR, new PasteEditProvider(languageClient), MIMETYPES));
+	}
+}
+
+/**
+ * `DocumentPasteEditProvider` that delegates to jdt.ls to make any changes necessary to the pasted text and add any additional workspace edits.
+ */
+class PasteEditProvider implements DocumentPasteEditProvider {
+
+	private languageClient: LanguageClient;
+	private copiedContent: string | undefined;
+	private copiedDocumentUri: string | undefined;
+
+	constructor(languageClient: LanguageClient) {
+		this.languageClient = languageClient;
+	}
+
+	async prepareDocumentPaste?(document: TextDocument, _ranges: readonly Range[], dataTransfer: DataTransfer, _token: CancellationToken): Promise<void> {
+		const copiedContent: string = await dataTransfer.get(TEXT_MIMETYPE).asString();
+		if (copiedContent) {
+			this.copiedDocumentUri = document.uri.toString();
+			this.copiedContent = copiedContent;
+		}
+	}
+
+	async provideDocumentPasteEdits(document: TextDocument, ranges: readonly Range[], dataTransfer: DataTransfer, token: CancellationToken): Promise<VDocumentPasteEdit> {
+
+		const insertText: string = await dataTransfer.get(TEXT_MIMETYPE).asString();
+
+		// don't try to provide for multi character inserts; the implementation will get messy and the feature won't be that helpful
+		if (!insertText || token.isCancellationRequested || ranges.length !== 1) {
+			return null;
+		}
+
+		const range = ranges[0];
+
+		const location: Location = {
+			range: this.languageClient.code2ProtocolConverter.asRange(range),
+			uri: document.uri.toString(),
+		};
+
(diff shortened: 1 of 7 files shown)


datho7561:
Here is a demo:

paste-string-demo

Roland is unable to get it working on his computer, so something is up.

rgrunber:
Please make sure this doesn't affect text blocks (eg. `""" ... """ )`

src/pasteEventHandler.ts
```diff
@@ -0,0 +1,94 @@
+import { CancellationToken, DataTransfer, DocumentPasteEdit, DocumentPasteEditProvider, DocumentPasteProviderMetadata, DocumentSelector, EndOfLine, ExtensionContext, languages, Range, TextDocument, window } from "vscode";
+
+/**
+ * Registers the vscode-java DocumentPasteEditProviders and sets them up to be disposed.
+ *
+ * @param context the extension context
+ */
+export function registerPasteEventHandlers(context: ExtensionContext) {
+	if (languages["registerDocumentPasteEditProvider"]){
```
rgrunber, line 9: @datho7561 @fbricon  Looking at https://github.com/eclipse-theia/theia/blob/7abca9d57c966d621f2acc26008cf1a9e454949b/packages/plugin-ext/src/plugin/plugin-context.ts#L674 , I don't see `registerDocumentPasteEditProvider`, so doesn't that imply this check would safely handle the case in such clients ?

fbricon: `languages["registerDocumentPasteEditProvider"]` is probably correct, but yeah the concern is what's the behavior of theia when enableProposedApi is on

datho7561: From my testing in Theia, no warnings appear and the language server starts as normal, but none of the features I introduced are present (paste works exactly as it did before my change).

src/pasteEventHandler.ts
```diff
@@ -0,0 +1,87 @@
...
+	private languageClient: LanguageClient;
+	private copiedDocumentUri: string | undefined;
+
+	constructor(languageClient: LanguageClient) {
+		this.languageClient = languageClient;
+	}
+
+	async prepareDocumentPaste?(document: TextDocument, _ranges: readonly Range[], dataTransfer: DataTransfer, _token: CancellationToken): Promise<void> {
+		const copiedContent: string = await dataTransfer.get(TEXT_MIMETYPE).asString();
+		if (copiedContent) {
```
CsCherrYY, line 47: I suggest to cache the copiedConent when copying. When pasting, compare the content with cached one to ensure the content comes from the `copiedDocumentUri`.

Consider a user copied a snippet from a Java source file (1), and then copied a snipped from browser (2)(copying outside VS Code would not trigger this event), then paste to another Java source file (3). now we have a former `copiedDocumentUri`, using imports from (1) to resolve the import conflict in the snippet from (2), might get unexpect result.

src/pasteEventHandler.ts
```diff
@@ -0,0 +1,87 @@
...
+			copiedDocumentUri: this.copiedDocumentUri,
+			formattingOptions: {
+				insertSpaces: <boolean>activeTextEditor.options.insertSpaces,
+				tabSize: <number>activeTextEditor.options.tabSize
+			}
+		};
+
+		const pasteResponse: DocumentPasteEdit = await commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.HANDLE_PASTE_EVENT, pasteEventParams);
+
+		if (pasteResponse) {
```
CsCherrYY, line 82: Here we should use `languageClient.protocol2CodeConverter.asWorkspaceEdit()` to convert the additionalEdit, since it's in LSP format.

vscode.d.ts
```diff
@@ -0,0 +1,16535 @@
+/*---------------------------------------------------------------------------------------------
```
CsCherrYY, line 1: should we keep this file in the repository? I guess it's not required to activate the extension at runtime, but for testing?

src/pasteEventHandler.ts
datho7561: Okay, I made sure to do the protocol conversion. I tested out the workspace edit feature by adding an edit that adds a comment at the top of the file, and it seems to be working.

datho7561:
Thanks for the review @CsCherrYY !

vscode.d.ts
CsCherrYY: That's good, but in this way the engine requirement is still `1.65.0` in the `package.json`. Could you please check if the extension works in VS Code version `1.65.0`?(which meets the requirement defined via the engine, but our extension uses APIs which are not exists in the VS Code runtime)

datho7561: Yep, I installed the .vsix on 1.65.0, and the extension still starts up.

CsCherrYY: Sounds great!

CsCherrYY:
I just checked the behaviours in different VS Code versions:

1.65.2 (the olderest version matches engine):
The extension work well without paste handler, and there is an error about the missing proposed API:
image

1.73.1 (newest stable): 
The extension work well without paste handler, and there is an error about the proposed API is not permitted (since redhat.java is not present at the whitelist in this release):
image

in the next stable release (suppose 1.74.0) our extension will work with the paste handler and there should not be any error. I'm considering if it's possible to update the [engine](https://github.com/redhat-developer/vscode-java/blob/1895510182daa35c7dff620539eb749d6b34e54f/package.json#L27) in the extension, to restrict the new version is only available to new engine users. Is there any consideration from Theia side? @rgrunber

CsCherrYY:
looks like the newest VS Code release (1.74.0) includes the allowance of the proposed API of redhat.java, @datho7561 could you please help verify if this PR works as expected?

datho7561:
I tried updating the engine version, but that didn't help. I also confirmed that the product.json contains the change.

CsCherrYY:
Did you set `editor.experimental.pasteActions.enabled` to true?

datho7561:
> Did you set `editor.experimental.pasteActions.enabled` to true?

Nope, I forgot. Thanks for your help, it's working!

CsCherrYY (approved):
LGTM!

src/pasteEventHandler.ts
```diff
@@ -0,0 +1,102 @@
...
+			location: location,
+			text: insertText,
+			copiedDocumentUri: this.copiedContent === insertText ? this.copiedDocumentUri : undefined,
+			formattingOptions: {
+				insertSpaces: <boolean>activeTextEditor.options.insertSpaces,
+				tabSize: <number>activeTextEditor.options.tabSize
+			}
+		};
+
+		const pasteResponse: PDocumentPasteEdit = await commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.HANDLE_PASTE_EVENT, pasteEventParams);
```
rgrunber, line 92: Calling `JSON.stringify(pasteEventParams)` will save some code on the language server side deserialization. If you look around the codebase, you'll see we use this in many places. I can make the change.

Issue #2721, opened 2022-10-05
org.eclipse.jdt.core.prefs overridden with default? values even if specified differently


stefanrybacki:
In a gradle project we have:

```
eclipse.jdt.file.withProperties { props ->
        props.setProperty "org.eclipse.jdt.core.circularClasspath", "warning"
        props.setProperty "org.eclipse.jdt.core.compiler.processAnnotations", "enabled"
        props.setProperty "org.eclipse.jdt.core.compiler.annotation.nullanalysis", "disabled"
    }
```

and we have

```
eclipse {
      synchronizationTasks eclipseJdtApt, eclipseJdt, cleanEclipseFactorypath, eclipseFactorypath
    }
```

so running `./gradlew eclipseJdt` for instance generated a correct `org.eclipse.jdt.core.prefs` file in `.settings`.
However, when clean and rebuilding from within vscode the `org.eclipse.jdt.core.prefs` file will be overridden not honoring the settings we set above, specifically `org.eclipse.jdt.core.compiler.annotation.nullanalysis=disabled` is overridden with enabled which causes class hierarchy issues in the compiler.

##### Environment
- Operating System: Windows but running SSH remote in Linux
- JDK version: 11
- Visual Studio Code version: 1.71.2
- Java extension version: 1.11.0

snjeza:
I can't reproduce the issue.
@stefanrybacki could you attach your settings.json, build.gradle and your log. See [Enable logging](https://github.com/redhat-developer/vscode-java/wiki/Troubleshooting#enable-logging)

snjeza:
@stefanrybacki you may want to take a look at https://github.com/redhat-developer/vscode-java/issues/2712#issuecomment-1268502179

stefanrybacki:
@snjeza thanks for looking into this. Unfortunately I cannot share the gradle and settings file as is, due to company policies. Let me check if I can create a reproducible project without critical information. I will also check if the linked component fixes my compile issues, however it might only the transitive compile problem I get due vscode overriding the jdt file.

snjeza:
@stefanrybacki  Could you try to add

```
{
    "java.compile.nullAnalysis.nonnull": [],
    "java.compile.nullAnalysis.nullable": [],
}
```
to your settings.json?

stefanrybacki:
@snjeza I can confirm that applying those settings and also changes the value of `org.eclipse.jdt.core.compiler.annotation.nullanalysis` to disabled in the jdt file. 
[...]

PR #2722, opened 2022-10-06
The type javax.annotation.Nonnull cannot be resolved

### LINKED ISSUE #2712: The type javax.annotation.Nonnull cannot be resolved
In one of my projects, I'm getting an error after upgrading to the latest version:

```
The type javax.annotation.Nonnull cannot be resolved. It is indirectly referenced from required .class filesJava(16777540)
Annotation type 'javax.annotation.Nonnull' cannot be found on the build path, which is implicitly needed for null analysisJava(536871894)
```

I saw the changelog and found that the issue relates to _enhancement - Enable annotation-based null analysis. See [#1693](https://github.com/redhat-developer/vscode-java/issues/1693)._

Removing "javax.annotation.Nonnull" from "java.compile.nullAnalysis.nonnull" config fixes the problem for me.

##### Environment
- Operating System: MacOS 12.6
- JDK version: openjdk 17 2021-09-14
- Visual Studio Code version: 1.71.2
- Java extension version: 1.11.0

snjeza:
Fixes #2712 

### DIFF
--- a/package.json
+++ b/package.json
@@ -963,21 +963,13 @@
         },
         "java.compile.nullAnalysis.nonnull": {
           "type": "array",
-          "default": [
-            "javax.annotation.Nonnull",
-            "org.eclipse.jdt.annotation.NonNull",
-            "org.springframework.lang.NonNull"
-          ],
+          "default": [],
           "markdownDescription": "Specify the Nonnull annotation types to be used for null analysis. If more than one annotation is specified, then the topmost annotation will be used first if it exists in your project dependencies.",
           "scope": "window"
         },
         "java.compile.nullAnalysis.nullable": {
           "type": "array",
-          "default": [
-            "javax.annotation.Nullable",
-            "org.eclipse.jdt.annotation.Nullable",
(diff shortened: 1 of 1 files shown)


rgrunber:
Either we introduce an option to disable null analysis and enable it by default, or we try to raise the priority of the spring framework annotations under the assumption that, if they're present, they should be preferred over others.

CsCherrYY:
it seems that https://github.com/eclipse/eclipse.jdt.ls/pull/2264 will solve https://github.com/redhat-developer/vscode-java/issues/2712, but I notice that there are several users want a clear way to disable the null analysis. e.g., https://github.com/redhat-developer/vscode-java/issues/2725, https://github.com/redhat-developer/vscode-java/issues/2721. So we might consider providing a simple way to disable it or disable it by default (like eclipse does). @rgrunber Any thoughts?

rgrunber:
It would have been nice to have a simple button to disable it, but I'm fine with having it enabled by default as long as we don't run into more issues with it.

PR #2729, opened 2022-10-10
Synchronize bundles on demand

jdneo:
- Allow server to ask bundles to synchronize before ready.
- When new bundle is detected, instead of asking user to reload window, we will try to let JDT.LS hot load the bundle first.

requires https://github.com/eclipse/eclipse.jdt.ls/pull/2267

### How to test
1. disable some extensions that have java extension contributions
2. launch redhat.java
3. after server is ready, enable an extension

https://user-images.githubusercontent.com/6193897/194814406-5f1925ee-df0a-40f3-905e-bb643c6c63fd.mp4

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -3,9 +3,10 @@
 import * as vscode from 'vscode';
 import * as path from 'path';
 import { Commands } from './commands';
+import { extensions } from 'vscode';
 
-let existingExtensions: Array<string>;
-export let buildFilePatterns: Array<string>;
+export let existingExtensions: Array<string> = [];
+export let buildFilePatterns: Array<string> = [];
 
 export function collectJavaExtensions(extensions: readonly vscode.Extension<any>[]): string[] {
 	const result = [];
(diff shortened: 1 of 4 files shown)


snjeza (approved):

jdneo:
This PR should also help the case that `no delegate command` exceptions are thrown. See: https://github.com/microsoft/vscode/issues/161563#issuecomment-1273214358

src/plugin.ts
```diff
@@ -44,19 +45,41 @@ export function collectBuildFilePattern(extensions: readonly vscode.Extension<an
...
-			if (action === selection) {
-				vscode.commands.executeCommand(restartId);
-			}
-		});
+export function getBundlesToRefresh(): IBundlesToRefresh {
+	const previousCache: string[] = [...existingExtensions];
+	const currentExtensions = collectJavaExtensions(extensions.all);
+	return {
+		toInstall: currentExtensions.filter(e => !previousCache.includes(e)),
+		toUninstall: previousCache.filter(e => !currentExtensions.includes(e)),
```
testforstephen, line 53: I will be cautious about the uninstall operation. Saying the bundle list of extension java-test-runner and vscode-pde both contain a bundle for JUnit test, jdt.ls will just install one of them based on the singleton capability of osgi bundle. If I disable java-test-runner extension, it may cause the shared bundle is eventually uninsalled.

jdneo: Ah, I didn't take care of this case, thanks for pointing out.

If we just ignore the bundles to be uninstalled, do you think it's acceptable?

Or, each time we simply send the whole newest bundle list to the server and let server reload them again?

Eskibear: > If we just ignore the bundles to be uninstalled

pros: fast.
cons: malformed/useless bundles remain there even if we uninstall corresponding extensions.

> each time we simply send the whole newest bundle list

pros: correctness. 
cons: slow. But if we do is only when list of `javaExtensions` changes, that's acceptale and I prefer this way.

jdneo: I prefer the second approach as well. Besides the correctness, the implementation is also more straightforward then the first one. Let's go for the second approach then.

rgrunber (approved):
Works for me. Feel free to merge when ready.

PR #2746, opened 2022-10-20
Add postfix completion support

jdneo:
requires https://github.com/eclipse/eclipse.jdt.ls/pull/2275


jdneo:
We re-discussed about the default value for the feature flag, and hope to make it enabled by default. Given the following reasons:

- User can use the setting `editor.snippetSuggestions` to change the sort locations.
- The number of supported postfix completions is still relatively small now.
- A way to expose the new capabilities.

PR #2747, opened 2022-10-21
Add an option to enable/disable null analysis

CsCherrYY:

requires https://github.com/eclipse/eclipse.jdt.ls/pull/2279

add new preference java.compile.nullAnalysis.mode.

- When set to interactive, a notification will show once null annotation types are detected when project is imported or null analysis related configuration changed.
- When set to automatic, the null analysis will be enabled.
- When set to disabled (default), the null analysis will be disabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,13 +203,16 @@ The following settings are supported:
 * `java.configuration.workspaceCacheLimit` : The number of days (if enabled) to keep unused workspace cache data. Beyond this limit, cached workspace data may be removed.
 * `java.import.generatesMetadataFilesAtProjectRoot` : Specify whether the project metadata files(.project, .classpath, .factorypath, .settings/) will be generated at the project root. Defaults to `false`.
 * `java.inlayHints.parameterNames.enabled`: Enable/disable inlay hints for parameter names. Supported values are: `none`(disable parameter name hints), `literals`(Enable parameter name hints only for literal arguments) and `all`(Enable parameter name hints for literal and non-literal arguments). Defaults to `literals`.
-* `java.compile.nullAnalysis.nonnull`: Specify the Nonnull annotation types to be used for null analysis. If more than one annotation is specified, then the topmost annotation will be used first if it exists in your project dependencies.
-* `java.compile.nullAnalysis.nullable`: Specify the Nullable annotation types to be used for null analysis. If more than one annotation is specified, then the topmost annotation will be used first if it exists in your project dependencies.
+* `java.compile.nullAnalysis.nonnull`: Specify the Nonnull annotation types to be used for null analysis. If more than one annotation is specified, then the topmost annotation will be used first if it exists in your project dependencies. This setting will be ignored if `java.compile.nullAnalysis.enabled` is set to `disabled`.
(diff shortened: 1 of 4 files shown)


CsCherrYY:
Since switching between `enabled` and `disabled` mode causes the deadlock of buildship easily (see https://github.com/eclipse/eclipse.jdt.ls/pull/2279#issuecomment-1288477050) and there are some regression of null analysis encountered, I suggest we make `disabled` the default value and once the buildship deadlock bug has been addressed, we can make `interactive` default.

package.json
```diff
@@ -984,7 +984,18 @@
...
+        },
+        "java.compile.nullAnalysis.mode": {
+          "type": "string",
+          "enum": [
+            "disabled",
+            "interactive",
+            "automatic"
+          ],
+          "default": "disabled",
+          "markdownDescription": "Specify how to enable the annotation-based null analysis. When set to `interactive`, you will have a chance to turn on the null analysis when the null annotations are detected in your project dependencies.",
```
rgrunber, line 998: I wonder if we need to explain what happens when `interactive` is set. If you look at other options we have (eg. `java.configuration.updateBuildConfiguration`, `java.project.importOnFirstTimeStartup`) that use `interactive`, we never go into details. Maybe it's understood that the setting implies some kind of UI prompt when the capability is about to change/enable.

Also, there's 3 other places in previous changes where we use ".. if it exists in your project dependencies". I would avoid this in favour of just ".. if it exists in project dependencies"

rgrunber (approved):
This part of the change looks pretty good. Just some minor things to address.

README.md
```diff
@@ -203,15 +203,16 @@ The following settings are supported:
...
 * `java.codeAction.sortMembers.avoidVolatileChanges`: Reordering of fields, enum constants, and initializers can result in semantic and runtime changes due to different initialization and persistence order. This setting prevents this from occurring. Defaults to `true`.
 * `java.jdt.ls.protobufSupport.enabled`: Specify whether to automatically add Protobuf output source directories to the classpath. **Note:** Only works for Gradle `com.google.protobuf` plugin `0.8.4` or higher. Defaults to `true`.
 * `java.jdt.ls.androidSupport.enabled`: [Experimental] Specify whether to enable Android project importing. When set to `auto`, the Android support will be enabled in Visual Studio Code - Insiders. **Note:** Only works for Android Gradle Plugin `3.2.0` or higher. Defaults to `auto`.
 
 New in 1.12.0
 * `java.completion.postfix.enabled`: Enable/disable postfix completion support. Defaults to `true`.
+* `java.compile.nullAnalysis.mode`: Specify how to enable the annotation-based null analysis. When set to `interactive`, you will have a chance to choose whether to turn on the null analysis when the null annotation types are detected in your project dependencies. Defaults to `disabled`.
```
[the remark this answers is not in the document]
CsCherrYY: Sure.

package.json
CsCherrYY: Yeah, let me remove the detail description of `interactive` and make the phase more readable.

jdneo:
Is `java.compile.nullAnalysis.enabled` a typo? I didn't find that setting. Do you mean to `java.compile.nullAnalysis.mode`? @CsCherrYY

CsCherrYY:
looks like an outdated configure name, let me correct it.

PR #2753, opened 2022-10-24
Support "Add all missing imports"

CsCherrYY:

requires: https://github.com/eclipse/eclipse.jdt.ls/pull/2292

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -172,10 +172,6 @@ export namespace Commands {
      */
     export const ORGANIZE_IMPORTS_SILENTLY = "java.edit.organizeImports";
     /**
-     * Add all missing imports.
-     */
-     export const ADD_ALL_MISSING_IMPORTS = "java.action.addAllMissingImports";
-    /**
      * Custom paste action (triggers auto-import)
      */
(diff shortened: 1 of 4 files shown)


CsCherrYY:
currently the definition of `addMissingImportsCommandSupport` is: supports command `java.action.addAllMissingImports`.

As for choosing imports, this PR just add a tail and optional argument to indicate if `restoreExistingImports` should be enabled, so IMO it's compatible with previous capability `advancedOrganizeImportsSupport`.

rgrunber:
Overall the change looks good. We have JDT-LS (quick-fix) calling to the client with `java.action.addAllMissingImports`, which calls back to JDT-LS with `java/addAllMissingImports`, which may call back to the client with `java.action.organizeImports.chooseImports`  before returning to the client with the chosen result. Is this the standard way to do it so we don't compute anything until we need to resolve it ?

CsCherrYY:
You're right. It seems that we don't need these capability/request/response, and just return the code action/quick fix as proposal and lazy resolve them. I'll update this PR later.

rgrunber (approved):

PR #2768, opened 2022-10-28
Add new settings for clean ups

### LINKED ISSUE #2144: Generate `@Override` on save when applicable
I want to know when a method is overriding its super class by automatically adding the `@Override` annotation on save.

##### Current Result

```java
public String toString() {
	return super.toString();
}
```

##### Expected Result

```java
@Override
public String toString() {
	return super.toString();
}
```

##### Additional Informations

Example config:
```json
{
	"editor.codeActionsOnSave": {
		"source.generate.finalModifiers": true,
		"source.generate.override": true
	}
}
```

datho7561:
- Add documentation (accessible through a command) for what each clean up does

Depends on eclipse/eclipse.jdt.ls#2298
Closes #2144

### DIFF
--- a/document/_java.learnMoreAboutCleanUps.md
+++ b/document/_java.learnMoreAboutCleanUps.md
@@ -1,0 +1,124 @@
+# Java Clean Ups
+
+Java clean ups are run on the current document whenever it's saved.
+They can fix a wide variety of issues, from code style to programming mistakes,
+and can even modernize the sources based on new Java language features.
+Here is some information on the supported cleanups and the details of what they do.
+
+### `qualifyMembers`
+
+Whenever a member (field or method) of a class is accessed from within the class,
+prefix the access with `this`.
+This is similar to how Python requires the programmer to access members using `self`.
+
+For instance:
+
+```java
+private int value;
+
+public void getValue() {
+	return value;
+}
+```
+
+becomes:
+
+```java
+private int value;
+
+public void getValue() {
+	return this.value;
+}
+```
+
+### `qualifyStaticMembers`
+
+Whenever there is a static variable or function, prefix the access with the name of the class that the static variable or function belongs to.
+
+For instance:
+
+```java
+import static java.lang.System.out;
+
+public class MyClass {
+	public static final double FACTOR = 0.5;
+
+	public double getNumber(double value) {
+		out.println("moo");
+		return value * FACTOR;
+	}
+}
+```
+
+becomes:
+
+```java
+import static java.lang.System.out;
+
+public class MyClass {
+	public static final double FACTOR = 0.5;
+
+	public double getNumber(double value) {
+		System.out.println("moo");
+		return value * MyClass.FACTOR;
+	}
+}
+```
+
+### `addOverride`
+
+When a method of a class that overrides a method from a parent class or provides an implementation for a method from an interface, add the `@Override` annotation.
+
+For example:
+
+```java
+public class MyRunner implements Runnable {
+	public void run() {
+		System.out.println("Hello, World!");
+	}
+}
+```
+
+becomes:
+
+```java
+public class MyRunner implements Runnable {
+	@Override
+	public void run() {
+		System.out.println("Hello, World!");
+	}
+}
+```
+
+### `addDeprecated`
+
+When a method is marked `@deprecated` in the Javadoc, but doesn't have the `@Deprecated` annotation, add the `@Deprecated` annotation.
+This only works if the compiler has been configured to mark deprecated methods without the deprecated annotation
+as an info/warning/error in the JDT settings.
+
+For example:
+
+```java
+/**
+ * Not used anymore, please stop using.
+ *
+ * @deprecated
+ */
+public boolean isAGoat() {
+	return false;
+}
+```
+
+becomes:
+
+```java
+/**
+ * Not used anymore, please stop using.
+ *
+ * @deprecated
+ */
+@Deprecated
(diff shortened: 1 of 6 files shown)


document/_java.learnMoreAboutCleanUps.md
```diff
@@ -0,0 +1,123 @@
+# Java Clean Ups
+
+Java clean ups are run on the current document whenever it's saved.
+They can fix code style and simple programming mistakes.
```
rgrunber, line 4: They can fix a wide variety of issues, from code style, programming mistakes, and even modernize the sources based on new Java language features.

document/_java.learnMoreAboutCleanUps.md
```diff
@@ -0,0 +1,123 @@
...
+Here is some information on the details of what each clean up does.
```
rgrunber, line 5: on the details of the supported cleanups and what they do.

package.json
```diff
@@ -980,6 +980,21 @@
           ],
           "markdownDescription": "Specify the Nullable annotation types to be used for null analysis. If more than one annotation is specified, then the topmost annotation will be used first if it exists in your project dependencies.",
           "scope": "window"
+        },
+        "java.cleanup.enabled": {
```
rgrunber, line 984: Don't forget to update this setting with the name we decide in JDT-LS. I think it needs to be named something that makes it more obvious it takes a list of items, and isn't a boolean.

rgrunber (commented):
Overall looks good to me. I think we just need to decide regarding the cleanups that can only be activated when their corresponding problem is set to error/warning.

package.json
```diff
@@ -980,6 +980,21 @@
           ],
           "markdownDescription": "Specify the Nullable annotation types to be used for null analysis. If more than one annotation is specified, then the topmost annotation will be used first if it exists in your project dependencies.",
           "scope": "window"
+        },
+        "java.cleanup.actionsOnSave": {
+          "type": "array",
+          "markdownDescription": "The list of clean ups to be run on the current document when it's saved. Clean ups can automatically fix code style or programming mistakes. [Click here](command:_java.learnMoreAboutCleanUps) to learn more about what each clean up does.",
+          "items": {
+            "type": "string",
+            "enum": [
```
rgrunber, line 989: This is a more of a comment, but in the future it probably won't be practical to have a very very large dropdown for all of these. For now we might be able to manage, but I feel like the best type would be some kind of text box that can auto-complete possible values. Is that even possible ?

package.json
```diff
@@ -980,6 +980,21 @@
...
+              "qualifyMembers",
+              "qualifyStaticMembers",
+              "addOverride",
+              "addDeprecated"
+            ]
+          },
+          "default": [],
+          "scope": "resource"
```
rgrunber, line 997: Shouldn't this be `window` ? I don't think we can set different behaviour dependening on the file. Certainly JDT has no way of storing resource-level scopes. At best we can retrieve it for the project.

package.json
datho7561: Oh, I didn't know it did that in the UI editor, that's neat! I've just been editing through "Preferences: Open User Settings (JSON)", where you can get autocomplete for the value and filter it through typing out the beginning of the name. It would be nice to do something similar in the UI, but I think that might need to be implemented in VS Code.

rgrunber:
In the README.md , there is an entry for "New in 1.13.0". Could you add this setting there ?

package.json
```diff
@@ -997,6 +997,21 @@
           "default": "disabled",
           "markdownDescription": "Specify how to enable the annotation-based null analysis.",
           "scope": "window"
+        },
+        "java.cleanup.actionsOnSave": {
+          "type": "array",
+          "markdownDescription": "The list of clean ups to be run on the current document when it's saved. Clean ups can automatically fix code style or programming mistakes. [Click here](command:_java.learnMoreAboutCleanUps) to learn more about what each clean up does.",
```
rgrunber, line 1003: https://github.com/redhat-developer/vscode-java/blob/2d99511c0a0559c7ea111545cd981bce318c0eef/package.json#L431

Small difference, but `[Click Here]` vs. `Click [HERE]` . May as well make it consistent with what we have.

package.nls.json
```diff
@@ -22,5 +22,6 @@
 	"java.action.showSupertypeHierarchy": "Show Supertype Hierarchy",
 	"java.action.showSubtypeHierarchy": "Show Subtype Hierarchy",
 	"java.action.changeBaseType": "Base on this Type",
-	"java.project.createModuleInfo.command": "Create module-info.java"
+	"java.project.createModuleInfo.command": "Create module-info.java",
+	"_java.learnMoreAboutCleanUps": "Learn more about Java Clean Ups"
```
rgrunber, line 26: I'm not sure this needs to be here for now. It looks like we don't do translation for any of the internal (eg. `_java`) commands. If you place it in here the idea is to place it in the other `package.json.nls-${locale}.json` files.

document/_java.learnMoreAboutCleanUps.md
```diff
@@ -0,0 +1,124 @@
+# Java Clean Ups
+
+Java clean ups are run on the current document whenever it's saved.
+They can fix a wide variety of issues, from code style to programming mistakes,
+and can even modernize the sources based on new Java language features.
+Here some information on the supported cleanups and the details of what they do.
```
rgrunber, line 6: _Here is some information .._
or
_Here, some information on the supported cleanups and details are listed :_

PR #2773, opened 2022-11-01
Support quickfix for gradle jpms projects

CsCherrYY:

related to https://github.com/eclipse/eclipse.jdt.ls/pull/2304

this PR removes unnecessary client command `_java.gradle.upgradeWrapper`, and for all upgrade gradle command, we use `java.project.upgradeGradle` instead, at the server side, we also have such delegate command for use.

This PR provides a quick fix to upgrade Gradle version to 7.0.1, it can mitigate some Gradle compability issues in JPMS projects (see: https://github.com/microsoft/vscode-java-pack/issues/1080). The following steps will help to reproduce these issues:

1. given sample project https://github.com/HanSolo/tilesfx.git
> Note: the Gradle plugin `biz.aQute.bnd.builder` introduced in that sample project does not support Gradle 7, so we might need to disable it first
2. open it in VS Code with auto-build enabled, you can find several errors telling some types is not accessable.
3. Since we detect the problems in classpath indicator and Gradle version, we provide a Quickfix to help you update the Gradle version
image
4. when you click the button, the extension will help you update the Gradle version defined in the wrapper file and re-import the project, after that the above errors get disappear.

### DIFF
--- a/src/standardLanguageClientUtils.ts
+++ b/src/standardLanguageClientUtils.ts
@@ -133,15 +133,14 @@ export async function upgradeGradle(projectUri: string, version?: string): Promi
 		title: "Upgrading Gradle wrapper...",
 		cancellable: true,
 	}, (_progress, token) => {
-		return commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, "java.project.upgradeGradle", projectUri, version, token);
+		return commands.executeCommand<string>(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.UPGRADE_GRADLE_WRAPPER, projectUri, version, token);
 	});
 	if (result) {
-		const propertiesFile = path.join(Uri.parse(projectUri).fsPath, "gradle", "wrapper", "gradle-wrapper.properties");
-		if (fse.pathExists(propertiesFile)) {
-			const content = await fse.readFile(propertiesFile);
+		if (await fse.pathExists(result)) {
+			const content = await fse.readFile(result);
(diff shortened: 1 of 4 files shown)


src/standardLanguageClientUtils.ts
```diff
@@ -133,11 +133,11 @@ export async function upgradeGradle(projectUri: string, version?: string): Promi
 		title: "Upgrading Gradle wrapper...",
 		cancellable: true,
 	}, (_progress, token) => {
-		return commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, "java.project.upgradeGradle", projectUri, version, token);
+		return commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.UPGRADE_GRADLE_WRAPPER, projectUri, version, token);
 	});
 	if (result) {
-		const propertiesFile = path.join(Uri.parse(projectUri).fsPath, "gradle", "wrapper", "gradle-wrapper.properties");
-		if (fse.pathExists(propertiesFile)) {
+		const propertiesFile = result as string;
```
jdneo, line 139: `commands.executeCommand` can accept a generic type as the return type, then you do not need to do the cast here

jdneo (approved):

jdneo (approved):
Verified that the change works according to the provided steps.

jdneo:
@rgrunber This change should belongs to the last milestone. Don't need to include it in the new changelog.

PR #2793, opened 2022-11-11
Support Gradle annotation processing

jdneo:
requires: https://github.com/eclipse/eclipse.jdt.ls/pull/2319

### DIFF
--- a/package.json
+++ b/package.json
@@ -328,6 +328,12 @@
           "description": "Setting for GRADLE_USER_HOME.",
           "scope": "window"
         },
+        "java.import.gradle.annotationProcessing.enabled": {
+          "type": "boolean",
+          "default": true,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -328,6 +328,12 @@
           "description": "Setting for GRADLE_USER_HOME.",
           "scope": "window"
         },
+        "java.import.gradle.annotationProcessing.enabled": {
```
jdneo, line 331: Will it be better if naming the setting without `import`?

`java.gradle.annotationProcessing.enabled`

snjeza (approved):

fbricon:
Can you test it against a micronaut project, that generates bytecode directly, not sources?
cc @graemerocher

graemerocher (approved):

PR #2795, opened 2022-11-11
Add missing import on paste

CsCherrYY:

https://user-images.githubusercontent.com/45906942/201302153-f53d0a54-95ea-41e9-9f35-c5bd735e3321.mp4

related to https://github.com/eclipse/eclipse.jdt.ls/pull/2320
this PR requests VS Code proposed API (documentPaste).

### DIFF
--- a/src/pasteEventHandler.ts
+++ b/src/pasteEventHandler.ts
@@ -1,0 +1,49 @@
+import { CancellationToken, commands, DataTransfer, DocumentPasteEdit, DocumentPasteEditProvider, DocumentPasteProviderMetadata, DocumentSelector, EndOfLine, ExtensionContext, languages, Range, TextDocument, TextEditor, window, WorkspaceEdit } from "vscode";
+import { LanguageClient } from "vscode-languageclient/node";
+import { Commands } from "./commands";
+import { JAVA_SELECTOR } from "./standardLanguageClient";
+
+const TEXT_MIMETYPE: string = "text/plain";
+
+const MIMETYPES: DocumentPasteProviderMetadata = {
+	pasteMimeTypes: [TEXT_MIMETYPE]
+};
+
+class StringPasteEditProvider implements DocumentPasteEditProvider {
+
+	private client: LanguageClient;
+	private copiedDocumentUri: string | undefined;
+	private copiedContent: string | undefined;
+
+	constructor(client: LanguageClient) {
+		this.client = client;
+	}
+
+	async prepareDocumentPaste(document: TextDocument, _ranges: readonly Range[], dataTransfer: DataTransfer, _token: CancellationToken): Promise<void> {
+		const copiedContent: string = await dataTransfer.get(TEXT_MIMETYPE).asString();
+		if (copiedContent) {
+			this.copiedContent = copiedContent;
+			this.copiedDocumentUri = document.uri.toString();
+		}
+	}
+
+	async provideDocumentPasteEdits(document: TextDocument, ranges: readonly Range[], dataTransfer: DataTransfer, _token: CancellationToken): Promise<DocumentPasteEdit> {
+		const insertContent: string = await dataTransfer.get(TEXT_MIMETYPE).asString();
+		if (!insertContent) {
+			return new DocumentPasteEdit("");
+		}
+		const documentPasteEdit = new DocumentPasteEdit(insertContent);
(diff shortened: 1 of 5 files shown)


CsCherrYY:
cc @fbricon @rgrunber , since we also have another PR using proposed paste API: https://github.com/redhat-developer/vscode-java/pull/2703, will it be OK if we publish with the proposed API? (to be specific, for Theia compatibility)

rgrunber:
I think @datho7561 confirmed the extension will install without issue, so there should be no compatibility problems. I think we just need to confirm that `enableProposedApi` also doesn't affect anything on Theia.

I've added David as reviewer since he recently worked with the API. It would be great if both changes took a similar approach.

datho7561:
One thing that I want to mention is that, at least from my testing, the copy/paste handlers only work on Insiders. I tried setting the launch flag to allow vscode-java to use them in regular VS Code, but this just silenced the warning about using the API and didn't allow the handler to work. Please let me know if you have found a way to use the copy/paste handler in regular VS Code.

rgrunber:
> One thing that I want to mention is that, at least from my testing, the copy/paste handlers only work on Insiders. I tried setting the launch flag to allow vscode-java to use them in regular VS Code, but this just silenced the warning about using the API and didn't allow the handler to work. Please let me know if you have found a way to use the copy/paste handler in regular VS Code.

@CsCherrYY , so I think we should be fine in Theia. Before we request an exception, is there a way to test the proposed API in a released VS Code instance ? How does one simulate the exception on a local instance. We can test in VS Code Insiders, but according to @datho7561 , there are issues getting it to run in a released version.

I would have thought if it can work in a released version, then the process to test should be the same as Insiders.

rgrunber:
I think there's going to be a bit of overlap between this change and https://github.com/redhat-developer/vscode-java/pull/2703 . What about cases where you have a snippet that has both missing imports and a string literal to be escaped ?  If both your providers are triggered independently, I'm not sure how they would combine that change. Do they all run in sequence or are they only given the original paste text ?

Seems like we should look to register a single provider and have the (currently 2) features be called on the resulting text. If both changes end up calling the language server, it might also reduce the number of requests/responses.

src/pasteEventHandler.ts
```diff
@@ -0,0 +1,49 @@
...
+		const insertContent: string = await dataTransfer.get(TEXT_MIMETYPE).asString();
+		if (!insertContent) {
+			return new DocumentPasteEdit("");
+		}
+		const documentPasteEdit = new DocumentPasteEdit(insertContent);
+		const workspaceEdit = await commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.ADD_IMPORTS_PASTE, document.uri.toString(), JSON.stringify(this.client.code2ProtocolConverter.asRange(ranges[0])), insertContent, this.copiedContent === insertContent ? this.copiedDocumentUri : null);
+		if (workspaceEdit) {
+			documentPasteEdit.additionalEdit = this.client.protocol2CodeConverter.asWorkspaceEdit(workspaceEdit);
+		}
+		return documentPasteEdit;
```
datho7561, line 40: I've been playing around with merging the changes in this PR and https://github.com/redhat-developer/vscode-java/pull/2703. When you register multiple paste handlers, VS Code only applies the changes from the first handler that returns a result. I think it would be a good idea to only return a result if there is an `additionalEdit` to apply.

Then again, I think that it might make sense to just have one workspace command for handling all text paste events (eg. that handles both adding imports and pasting into string literals). I have opened a draft PR with this idea: https://github.com/eclipse/eclipse.jdt.ls/pull/2349/files. I need to add a field in the parameters for the original file uri, but otherwise, please let me know what you think of the API.

CsCherrYY: The API LGTM. As for the field original file uri, it can be either of the two ways:

- we send a new request in `prepareDocumentPaste`, making server be responsible for recording the original files. Once paste happens, server can compare the uri and content, to choose correct original file and get the imports. 
  - It will remove the requirement of client from recording the original files and keep the current API clear, but add an addtional request.
- keep current impelmentation, and add a new field `originalFileUri` in the API.

WDYT?

datho7561: I think that adding a new field to the `handlePasteEvent` might be the better option. That way, we only have one new command on the language server. This also makes it easier to test the paste event handler, since the handler itself won't have any state.

datho7561: Okay. I've updated my PRs. When you get a chance, could you let me know what you think? Thanks!

Issue #2723, opened 2022-10-06
Are shared indexes on the roadmap?


DatGuyJonathan:
Large projects can take a really long time to index locally which leads to an unproductive dev flow. Intellij's solution to that seems to be shared indexes. Would that be useful for vscode-java extension as well? Or are there other plans to improve indexing?

jdneo:
There is actually a discussion on upstream side, see: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/110

PR #2811, opened 2022-11-22
Share indexes among workspaces

testforstephen:
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2341

Provide a setting to control the shared index location, by default, it uses the global extension cache as the common index location.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1040,10 +1040,21 @@
           "default": [],
           "scope": "window"
         },
+        "java.sharedIndexes.enabled": {
+          "type": "string",
+          "enum": [
+            "auto",
+            "on",
+            "off"
+          ],
+          "default": "auto",
+          "markdownDescription": "[Experimental] Specify whether to share indexes between different workspaces. When set to `auto`, shared indexes will be enabled in Visual Studio Code - Insiders.",
+          "scope": "window"
+        },
         "java.sharedIndexes.location": {
           "type": "string",
-          "markdownDescription": "Specifies a common index location for all workspaces. The default value '`$DEFAULT_CACHE`' uses the extension global cache directory as the shared index location. And leaving it blank disables the shared indexes.",
-          "default": "$DEFAULT_CACHE",
+          "markdownDescription": "Specifies a common index location for all workspaces. See default values as follows:\n \nWindows: `~\\.jdt\\index`\n \nmacOS: `~/Library/Caches/.jdt/index`\n \nLinux: `~/.jdt/index`",
+          "default": "",
           "scope": "window"
         }
       }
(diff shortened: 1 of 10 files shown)


rgrunber:
If it makes sense, this should probably close out https://github.com/redhat-developer/vscode-java/issues/2723 .

src/javaServerStarter.ts
```diff
@@ -165,6 +170,26 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
 	return params;
 }
 
+function resolveIndexCache(context: ExtensionContext) {
+	let sharedIndexLocation: string = getJavaConfiguration().get("sharedIndexes.location");
+	if (sharedIndexLocation === "$DEFAULT_CACHE") {
+		sharedIndexLocation = path.resolve(context.globalStoragePath, "index");
```
jdneo, line 176: `globalStoragePath` is deprecated. Maybe consider using `globalStorageUri` and get the path only when the scheme is `file`?

testforstephen:
Comments:
- Verify how to regenerate index when the library contents has changed but file path doesn't change
- Default cache location: something like `/Users/name/.cache/redhat.java/index`
- Provide a command to cleanup everything in the shared index location

src/javaServerStarter.ts
```diff
@@ -165,6 +170,50 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
...
+		}
+	}
+
+	return location;
+}
+
+export function getSharedIndexCache(): string {
+	let location: string = getJavaConfiguration().get("sharedIndexes.location");
+	if (!location) {
+		switch (process.platform) {
```
rgrunber, line 198: Is there a reason you abandoned the use of `context.globalStorageUri` ?

package.json
```diff
@@ -1040,6 +1040,23 @@
           },
           "default": [],
           "scope": "window"
+        },
+        "java.sharedIndexes.enabled": {
```
rgrunber, line 1044: I would just make sure to add this as new in 1.14.0 in the README.md

src/javaServerStarter.ts
testforstephen: In our last sync-up meeting, one comment is to save the index into the OS specific cache directory, some place like what Gradle does. That's why I changed the cache from the global storage location to some OS specific directory.

package.json
testforstephen: done.

src/javaServerStarter.ts
testforstephen: this is outdated. we have changed it to some OS specific cache directory.

PR #2818, opened 2022-11-28
Translate `jdt://` links in Hover to a command link

### LINKED ISSUE #2810: When viewing hover Javadoc, `{@link ...` that points to a binary class doesn't display as a link
If you hover over a Java element that has Javadoc (eg. class, method, static/non-static member) that has a `{@link org.example.ClassName}` in it, the `{@link` is rendered as a link and is clickable if the class is a source file, but not if it is a binary file. For example, if you put `{@link java.lang.String}` in a Javadoc, then the hover documentation doesn't display a clickable link.

##### Environment
- Operating System: RHEL 8.6
- JDK version: The bundled Linux JustJ
- Visual Studio Code version: 1.73.1
- Java extension version: v1.13.2022111903 (prerelease)

##### Steps To Reproduce
1. Make a class file:
```java
public class MyClass {

  /**
   * {@link java.lang.String}
   */
  public void myMethod() {
    myMethod();
  }
}
```
2. Hover over `myMethod()` to see the Javadoc

##### Current Result
The reference to `java.lang.String` is not clickable.

##### Expected Result
The reference to `java.lang.String` is a clickable link that opens `java.lang.String` in read-only mode.

##### Additional Informations
You can see in the extension output under the hover response that it tries to link to `java.lang.String` like this:
```
[java.lang.String](jdt://contents/java.base/java.lang/String.class?=quarkius/%5C/home%5C/davthomp%5C/.sdkman%5C/candidates%5C/java%5C/17.0.3-tem%5C/lib%5C/jrt-fs.jar%60java.base=/javadoc_location=/https:%5C/%5C/docs.oracle.com%5C/en%5C/java%5C/javase%5C/17%5C/docs%5C/api%5C/=/=/maven.pomderived=/true=/%3Cjava.lang%28String.class#140)
```

datho7561:
Translate all links to `jdt://` in hover documentation to a `command://` link that opens the corresponding document. This gets around a limitation of hover Markdown in VS Code, where custom schemes aren't supported in links.

Closes #2810

### DIFF
--- a/src/providerDispatcher.ts
+++ b/src/providerDispatcher.ts
@@ -1,15 +1,14 @@
 'use strict';
 
-import { DocumentSymbolRequest, SymbolInformation as clientSymbolInformation, DocumentSymbol as clientDocumentSymbol, HoverRequest, WorkspaceSymbolRequest } from "vscode-languageclient";
+import { CancellationToken, commands, DocumentSymbol, DocumentSymbolProvider, Event, ExtensionContext, Hover, HoverProvider, languages, MarkdownString, MarkedString, Position, Range, SymbolInformation, SymbolKind, TextDocument, TextDocumentContentProvider, Uri, workspace, WorkspaceSymbolProvider } from "vscode";
+import { DocumentSymbol as clientDocumentSymbol, DocumentSymbolRequest, HoverRequest, SymbolInformation as clientSymbolInformation, WorkspaceSymbolRequest } from "vscode-languageclient";
 import { LanguageClient } from "vscode-languageclient/node";
-import { ExtensionContext, languages, DocumentSymbolProvider, TextDocument, CancellationToken, SymbolInformation, DocumentSymbol, TextDocumentContentProvider, workspace, Uri, Event, HoverProvider, Position, Hover, WorkspaceSymbolProvider, Range, commands, SymbolKind } from "vscode";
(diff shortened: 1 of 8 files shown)


src/providerDispatcher.ts
```diff
@@ -170,4 +171,24 @@ function overwriteWorkspaceSymbolProvider(context: ExtensionContext): void {
...
 	});
+}
+
+const REPLACE_JDT_LINKS_PATTERN: RegExp = /(\[(?:[^\]])+\]\()(jdt:\/\/(?:(?:(?:\\\))|([^)]))+))\)/;
+function fixJdtSchemeHoverLinks(hover: Hover): Hover {
+	const newContents: (MarkedString|MarkdownString)[] = [];
+	for (const content of hover.contents) {
+		if (content instanceof MarkdownString) {
+			const newContent: string = (<MarkdownString>content).value.replace(REPLACE_JDT_LINKS_PATTERN, (_substring, group1, group2) => {
+				const uri = `command:java.open.file?${encodeURI(JSON.stringify([encodeURIComponent(group2)]))}`;
```
rgrunber, line 182: I would use `Commands.OPEN_FILE` here.

rgrunber (commented):
Overall this looks good, so we should probably merge.

I would just add some comment about why this is needed. if this is some behaviour that we are now implementing to work around a limitation imposed by VS Code, then it should be mentioned.

PR #2824, opened 2022-12-02
refactor: move utilities out of extension.ts

Eskibear:
benefits:
- avoid circular dependencies related to `extension.ts`
- shrink size of extension.ts for better readability


testforstephen (approved):
LGTM

PR #2827, opened 2022-12-05
Add settings for some more cleanups.

rgrunber:
- addFinalModifier: Add final modifier where possible
- instanceofPatternMatch: Use pattern matching for instanceof checks
- lambdaExpression: Convert anonymmous class usage to lambda expressions
- switchExpression: Convert switch statement to switch expression


rgrunber:
Starting to wonder if we should sort these ..

datho7561 (approved):
Looks good to me! Thanks, Roland!

PR #2863, opened 2023-01-09
Standardize all snippets

EmilySeville7cfg:
All snippets are named using the following convention:

- if snippet is not for some method (except `main`), class or interface, then:
  - first two letters from the first word used **and** one letter from a second word (if it exists) as a snippet prefix
- else
  - full method/class/interface name used as a snippet prefix

Motivation: make snippets short and standardized across different VS Code extensions (see my recent PRs for details).


EmilySeville7cfg:
All the snippets in server.json have another detailed implementation at server side, which can provide more context like referencing a variable nearby. So, after the language server is ready, the client side will retire all the snippets in the server.json and let server to serve them. See: 

Should I move all snippets to `server.json`?

> About the label of the snippets, I think most of them have referred the name in other popular Java IDE (Eclipse, IntelliJ IDEA). For example, the snippet for main method is named main or psvm in these two IDEs, and that's where the name comes from. By doing so, users can quickly find them according to their previous experience in other IDEs.

Labels like `psvm` are not readable. I'm gonna to address this issue in another IDEs too. My goal to provide both short and readable labels. 😄 I wanna easily predict what label kind of looks like for smth.

jdneo:
> Labels like psvm are not readable. I'm gonna to address this issue in another IDEs too.

I understand that. `psvm` was asked several times in our user survey before. Different people may have different opinions on that.

Personally, I think a more ideal way to address this issue is to allow users configure the snippets by their own (via settings). Other IDEs have such capability which it's not supported right now in VS Code Java.

EmilySeville7cfg:
It's possible to create custom snippets in VS Code. Is it what do u mean?

fbricon:
User snippets can already be defined as per https://code.visualstudio.com/docs/editor/userdefinedsnippets

I don't think we need to define more settings to encapsulate those, but it'd be nice to extend user defined snippets to be  more java-context aware.

PR #2895, opened 2023-01-20
Use commit SHA instead of branch name for third-party actions

andreaTP:
Hi!
Following the [GH Action Security Hardening](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) guide we should use the commit SHA instead of the `branch` or `tag` for any third-party untrusted action.


fbricon:
@andreaTP thanks for your contribution!

PR #2901, opened 2023-01-25
Return early to allow commands to be run sooner

### LINKED ISSUE #2900: "Check details" link ineffective on startup, until the server actually finished initialization
When opening a folder fo the 1st time, project import can take quite a while. While the "Opening Java projects: check details"  popup is opened, clicking on the "Check details" link is most of the time useless, as it doesn't seem to work until the server has finished initialization.

datho7561:
Closes #2900

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -76,7 +76,7 @@ function getHeapDumpFolderFromSettings(): string {
 }
 
 
-export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
+export function activate(context: ExtensionContext): Promise<Promise<ExtensionAPI> | ExtensionAPI> {
 	context.subscriptions.push(markdownPreviewProvider);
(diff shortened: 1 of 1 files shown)


datho7561:
@fbricon @rgrunber does this fix it for you? It seems to fix it for me.

What extensions use the vscode-java API? I don't think this change modifies the API, but I want to test a bit more to be sure (this doesn't seem to break vscode-microprofile).

rgrunber:
It still fails for me on large projects like eclipse.jdt.ls. Doesn't seem to improve things for me. Even if you activate from the command palette, `Java: Show Build Job Status`, it still doesn't show up immediately. Instead I see that the extension itself is attempting to be activated, but it doesn't happen.

datho7561:
Yep. Sorry, I should have tested this better.

PR #2905, opened 2023-01-27
Return the API before importing the projects

### LINKED ISSUE #2900: "Check details" link ineffective on startup, until the server actually finished initialization

datho7561:
This also has the side effect that extensions that depend on vscode-java can start up sooner.

Fixes #2900

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -243,12 +243,16 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			};
 
 			apiManager.initialize(requirements, serverMode);
+			resolve(apiManager.getApiInstance());
+			// the promise is resolved
+			// no need to pass `resolve` into any code past this point,
+			// since `resolve` is a no-op from now on
 
 			if (requireSyntaxServer) {
 				if (process.env['SYNTAXLS_CLIENT_PORT']) {
-					syntaxClient.initialize(requirements, clientOptions, resolve);
+					syntaxClient.initialize(requirements, clientOptions);
 				} else {
-					syntaxClient.initialize(requirements, clientOptions, resolve, prepareExecutable(requirements, syntaxServerWorkspacePath, getJavaConfig(requirements.java_home), context, true));
+					syntaxClient.initialize(requirements, clientOptions, prepareExecutable(requirements, syntaxServerWorkspacePath, getJavaConfig(requirements.java_home), context, true));
 				}
 				syntaxClient.start();
 				serverStatusBarProvider.showLightWeightStatus();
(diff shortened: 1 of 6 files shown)


src/syntaxLanguageClient.ts
```diff
@@ -111,15 +108,4 @@ export class SyntaxLanguageClient {
 		return this.languageClient;
 	}
 
-	public resolveApi(resolve: (value: ExtensionAPI) => void): void {
-		apiManager.getApiInstance().serverMode = ServerMode.lightWeight;
```
datho7561, line 115: oops can't get rid of all of this

src/extension.ts
```diff
@@ -243,12 +243,16 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 			};
 
 			apiManager.initialize(requirements, serverMode);
+			resolve(apiManager.getApiInstance());
```
datho7561, line 246: I'll open an issue to refactor `activate` to use `await`. It can get confusing what is happening and in what order, and the function is really long.

datho7561: https://github.com/redhat-developer/vscode-java/issues/2906

rgrunber:
Yup, the underlying issue here was definitely a startup issue. Wouldn't be surprised if we hit it now because VS Code is now properly waiting for extension activation. The early `resolve()` after the `apiManager` is initialized definitely fixed it for me.

rgrunber (approved):
Change works for me.

jdneo (approved):
LGTM.

Though the API instance is resolved early, plugins of JDT-LS can leverage and wait `serverReady()` during activation.

test/lightweight-mode-suite/publicApi.test.ts
```diff
@@ -31,7 +31,7 @@ suite('Public APIs - LightWeight', () => {
 
 	test('status should be correct', async function () {
 		const api: ExtensionAPI = extensions.getExtension('redhat.java').exports;
-		assert.equal(api.status, 'Started');
+		assert.equal(api.status, 'Starting');
```
rgrunber, line 34: Was it `Started` before because we didn't resolve the api until the language server was up ? Now we do it much sooner, hence `Starting` .

datho7561: Yes.

rgrunber:
@jdneo I'll merge this. I would just keep an eye on any extension tests that may depend on this, or any places that make any assumptions about the state of the language server through the api manager immediately on startup.

PR #2925, opened 2023-02-08
Register the paste handler when LS reports ServiceReady

### LINKED ISSUE #2924: Experimental paste handling blocks pasting while project loading
slow-paste-handler

```
[Trace - 11:43:42] Sending request 'workspace/executeCommand - (3)'.
Params: {
    "command": "java.edit.handlePasteEvent",
    "arguments": [
        "{\"location\":{\"range\":{\"start\":{\"line\":83,\"character\":0},\"end\":{\"line\":83,\"character\":0}},\"uri\":\"file:///home/rgrunber/git/lemminx/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java\"},\"text\":\"XMLLanguageServer\",\"formattingOptions\":{\"insertSpaces\":false,\"tabSize\":4}}"
    ]
}
...
...
[Trace - 11:43:56] Received response 'workspace/executeCommand - (3)' in 13986ms.
No result returned.
```

I tried inserting a breakpoint into the delegate to see if some part of the command was being blocked, but it seems like there's a good chunk of time where the language server hasn't received the request, yet the client is clearly waiting, so maybe this is the language client delaying the request until the server is ready.

CC @datho7561 , @CsCherrYY 

**Update:** Note that regular pasting does work if done before the language server starts up, which is probably prior to `onReady`. That would at least seem to imply that `onReady` is not enough, and we need a stronger condition when registering the paste provider. Maybe `ServiceReady` or `Started` ?

datho7561:
Fixes #2924

### DIFF
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -123,6 +123,7 @@ export class StandardLanguageClient {
 							});
 						}
 
+						registerPasteEventHandler(context, this.languageClient);
 						activationProgressNotification.hide();
 						if (!hasImported) {
 							showImportFinishNotification(context);
@@ -595,7 +596,6 @@ export class StandardLanguageClient {
 				context.subscriptions.push(languages.registerInlayHintsProvider(JAVA_SELECTOR, new JavaInlayHintsProvider(this.languageClient)));
 			}
 
-			registerPasteEventHandler(context, this.languageClient);
 		});
 	}
 


rgrunber (approved):
Seems to fix the issue for me. Only question I have is regarding style. Should we use our own apiManager to do this, or should we place this in the `ServiceReady` case where the notification first comes in. I notice we do have some code there anyways. My preference is to have all the registrations that rely on the language server being in a certain state to be in one spot.

Do we have any other registrations that rely on the server being up? I guess we have some commands but they get activated using the `javaLSReady` context.

PR #2928, opened 2023-02-09
Enable no-response bot to auto close inactive issues

testforstephen:

It auto closes those inactive issues:
- Marked as `"need info"`
- Inactive for 14 days

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -7,8 +7,7 @@ on:
   issue_comment:
     types: [created]
   schedule:
-    # every morning at 00:00 am
-    - cron: '0 0 * * *'
+    - cron: '0 10 * * *'
 
 jobs:
   noResponse:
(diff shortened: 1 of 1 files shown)


.github/workflows/no-response.yml
```diff
@@ -0,0 +1,24 @@
+name: No Response
+
+# **What it does**: Closes issues where the original author doesn't respond to a request for information.
+# **Why we have it**: To remove the need for maintainers to remember to check back on issues periodically to see if contributors have responded.
+# **Who does it impact**: Everyone that works on docs or docs-internal.
```
fbricon, line 5: that last line is irrelevant

.github/workflows/no-response.yml
```diff
@@ -0,0 +1,24 @@
...
+    types: [created]
+  schedule:
+    # every morning at 00:00 am
+    - cron: '0 0 * * *'
+
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lee-dohm/no-response@v0.5.0
```
fbricon, line 18: please use the SHA1 commit of that version, as per https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

fbricon:
I think we need to manually go over all of https://github.com/redhat-developer/vscode-java/issues?q=is%3Aopen+is%3Aissue+label%3A%22need+info%22 to check if users actually gave some info, and remove the label if needed, before we apply this action.

rgrunber:
I've gone over all the `need info` labels. I definitely learned some interesting things along the way and for the most part I've either :

- Removed `need info` because additional info was given, or I felt there was enough info that an investigation at a later time was worth it
- Closed the issue because it was either a duplicate, or it was very similar to issues we fixed in the past, etc.
- Left the issue because it had `need info`, no response / no meaningful information to make any progress

I've also bumped the "days inactive" for auto-close to 30, as 14 seemed a bit too short.

PR #2944, opened 2023-02-16
Avoid inserting duplicate contents for file creation refactoring

### LINKED ISSUE #2939: Generate the code repeatedly in new java file
##### Environment
- Operating System: Mac
- JDK version: 17.0.6
- Visual Studio Code version: 1.75.1

##### Steps To Reproduce
1. Open a java project, like https://github.com/redhat-developer/vscode-java/tree/master/test/resources/projects/eclipse/simple-app
1. Type the code `File f = new File("demo.txt");` in a main method > Code Actions > Select `Create class xxx`, or `Create interface xxx`, or `Create enum xxx`

double-showing-code

##### Current Result
Show the code repeatedly in the new java file.

##### Expected Result
Only show the code one time

testforstephen:
Fixes #2939

The client side listener should avoid inserting duplicated conents when the file creation event is triggered from a WorkspaceEdit.
Given that the VS Code API doesn't provide a way to distinguish the event source, a workaround is to wait 100ms and let WorkspaceEdit to take effect first, then check if the workingcopy is filled with content.

### DIFF
--- a/src/fileEventHandler.ts
+++ b/src/fileEventHandler.ts
@@ -56,67 +56,85 @@ async function handleNewJavaFiles(e: FileCreateEvent) {
         }
     }
 
-    const formatNumber = (num => num > 9 ? String(num) : `0${num}`);
-    for (let i = 0; i < emptyFiles.length; i++) {
-        const typeName: string = resolveTypeName(textDocuments[i].fileName);
-        const isPackageInfo = typeName === 'package-info';
-        const isModuleInfo = typeName === 'module-info';
-        const date = new Date();
-        const context: any = {
-            fileName: path.basename(textDocuments[i].fileName),
-            packageName: "",
-            typeName: typeName,
-            user: userInfo().username,
-            date: date.toLocaleDateString(undefined, {month: "short", day: "2-digit", year: "numeric"}),
-            time: date.toLocaleTimeString(),
-            year: date.getFullYear(),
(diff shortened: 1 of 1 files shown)


fbricon:
why did it work before then?

testforstephen:
> why did it work before then?

It might be some changes on VS Code that leads the different behavior. Since we have supported this file event for years, no idea when this starts breaking.

testforstephen:
A quick check that this starts not working from VS Code 1.75. After we merge this workaround, we need to open an issue in upstream VS Code to discuss the cause further.

fbricon:
Diff ignoring whitespaces: https://github.com/redhat-developer/vscode-java/pull/2944/files?w=1

fbricon:
With the new patch, there's no longer duplicate content, but the new file doesn't open automatically either, as it used to.
Feb-16-2023 15-54-36

testforstephen:
> With the new patch, there's no longer duplicate content, but the new file doesn't open automatically either, as it used to.

The opening issue is not caused by this PR. The server side refactoring has no permission to ask client to open file.

If you tried it in VS Code 1.74, it also didn't open the new file automatically. In VS Code 1.75, you observed it opened the new file accidentally because the regression issue requires to open editor to insert snippets.

rgrunber (approved):
Works for me. Empty files created on the client side continue to work.

If there's no other issues, I think we should merge.

PR #2967, opened 2023-02-27
Support change signature refactoring

### LINKED ISSUE #2104: [Feature Request] Method Signature Refactoring
All prevalent IDEs such as Eclipse, IntelliJ, Netbeans provides a way to refactor method signatures.
It would be nice to have such a feature implemented in VSCode since the methods are usually spread across the entire project,  and currently no such refactoring actions are supported(checked the documentation and opened/closed issues).
Visual Studio seems to have such an existing [feature](https://docs.microsoft.com/en-us/visualstudio/ide/reference/change-method-signature?view=vs-2019)

Thank you for the hard work.

CsCherrYY:
fix #2104 

related to https://github.com/eclipse/eclipse.jdt.ls/pull/2497

a simple demo:

https://user-images.githubusercontent.com/45906942/221733196-0613e759-d6b7-4fce-929e-85faff474f4b.mp4


package.json
```diff
@@ -1417,19 +1421,24 @@
...
-    "ts-loader": "^9.2.6",
+    "style-loader": "^3.3.1",
+    "ts-loader": "^9.4.2",
     "typescript": "^4.6.4",
     "webpack": "^5.28.0",
     "webpack-cli": "^4.6.0"
   },
   "dependencies": {
+    "@vscode/codicons": "^0.0.32",
+    "@vscode/webview-ui-toolkit": "^1.2.1",
```
Eskibear, line 1432: 1.2.2 is out, fixing some react related issues. Not sure if you want to update. BTW, according to my self experience this package sometimes has breakage on patch/minor version updates, suggest to use fixed version. 

https://github.com/microsoft/vscode-webview-ui-toolkit/releases/tag/v1.2.2

src/webview/changeSignature/App.tsx
```diff
@@ -0,0 +1,413 @@
...
+				<VSCodeTextArea className={"preview"} value={this.getPreview()} readOnly={true}>Method signature preview:</VSCodeTextArea>
+				<VSCodeCheckbox id="delegate" className={"delegate"} onClick={this.onClick}>Keep original method as delegate to changed method</VSCodeCheckbox>
+				<div className={"bottom-buttons"}>
+					<VSCodeButton className={"vsc-button-left"} appearance="primary" onClick={this.onClick} id={"refactor"}>Preview and Refactor</VSCodeButton>
+				</div>
+			</main >
+		);
+	};
+
+	private getSelectedRowNumber(id: string): number | undefined {
```
Eskibear, line 406: better to add some comments, noting the required format of element id.

src/webview/vscodeapiWrapper.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * A utility wrapper around the acquireVsCodeApi() function, which enables
+ * message passing and state management between the webview and extension
+ * contexts.
+ *
+ * This utility also enables webview code to be run in a web browser-based
+ * dev server by using native web browser features that mock the functionality
+ * enabled by acquireVsCodeApi.
+ */
+class VSCodeAPIWrapper {
```
Eskibear, line 12: following camel case to rename the file to "vscode**A**piWrapper"?

src/webview/vscodeapiWrapper.ts
```diff
@@ -0,0 +1,78 @@
...
+   * @remarks When running webview code inside a web browser, postMessage will instead
+   * log the given message to the console.
+   *
+   * @param message Abitrary data (must be JSON serializable) to send to the extension context.
+   */
+  public postMessage(message: unknown) {
+    if (this.vsCodeApi) {
+      this.vsCodeApi.postMessage(message);
+    } else {
+      console.log(message);
```
Eskibear, line 35: should print more readable message about it.

src/webview/vscodeapiWrapper.ts
```diff
@@ -0,0 +1,78 @@
...
+   * @remarks When running webview source code inside a web browser, getState will retrieve state
+   * from local storage (https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).
+   *
+   * @return The current state or `undefined` if no state has been set.
+   */
+  public getState(): unknown | undefined {
+    if (this.vsCodeApi) {
+      return this.vsCodeApi.getState();
+    } else {
+      const state = localStorage.getItem("vscodeState");
```
Eskibear, line 51: is the key `vscodeState` defined by vscode or by you? shall we make it a constant variable?

CsCherrYY: By VS Code. The file is simply copied from the toolkit samle: https://github.com/microsoft/vscode-webview-ui-toolkit-samples/blob/main/frameworks/hello-world-react-cra/webview-ui/src/utilities/vscode.ts and currently this method is not used (only `postMessage` is used).

Should we just remove the unused ones now or keep them?

Eskibear: state can be accessed by vscode's webview APIs. 
https://code.visualstudio.com/api/extension-guides/webview#getstate-and-setstate

No idea why the lib needs to access `localStorage` as a fallback solution. My question was, does vscode write things into localStorage `vscodeState`? Or the key can be anything else to store the state?

Anyway, you should remove it if it's not used at all.

CsCherrYY: >My question was, does vscode write things into localStorage vscodeState? Or the key can be anything else to store the state?

see https://github.com/redhat-developer/vscode-java/pull/2967/files#diff-833529db0fa7b608f5a550b714774cfda2d7b4a46fa6486679ff5041cb077eecR68-R72, things will be written into localStorage when running in a web browser (maybe simple browser in VS Code?), which is not the case here I think.

src/webview/changeSignature/App.tsx
CsCherrYY: commends added in https://github.com/redhat-developer/vscode-java/pull/2967/commits/48037a8ae34a620109f2222ca1f0167f7e86f9d3

src/refactoring/changeSignaturePanel.ts
```diff
@@ -0,0 +1,193 @@
...
+			* The current vscode doesn't provide a way for the extension to pre-select all changes.
+			*
+			* As a workaround, this extension would append a dummy text edit that needs a confirm,
+			* and then make all others text edits not need a confirm. This will ensure that
+			* the REFACTOR PREVIEW panel can be triggered and all valid changes pre-selected.
+			*/
+			if (preview) {
+				const textEditEntries = codeEdit.entries();
+				if (textEditEntries && textEditEntries.length) {
+					const dummyNodeUri: Uri = textEditEntries[textEditEntries.length - 1][0];
```
jdneo, line 180: Is it possible to use a non-exist uri for the dummy node?

src/webview/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { Uri, Webview } from "vscode";
+
+export function getUri(webview: Webview, extensionUri: Uri, pathList: string[]) {
+    return webview.asWebviewUri(Uri.joinPath(extensionUri, ...pathList));
+}
+
+export function getNonce() {
```
jdneo, line 7: In `markdownPreviewProvider.ts`, it also defines a `getNonce()` util. We can replace that with this one.

src/refactoring/changeSignaturePanel.ts
CsCherrYY: Using a non-exist uri here is possible. VS Code will infer the "file" and "path" itself:

image

But the problem here is that once the user checks this dummy node, the workspace edit will fail to apply. I'm afraid it's causing more issues.

src/webview/utils.ts
CsCherrYY: done in https://github.com/redhat-developer/vscode-java/pull/2967/commits/4705403b3c484ef1cdc0ef18c6812f9c5aa74796

src/refactoring/changeSignaturePanel.ts
jdneo: ok, then forget about it.

jdneo:
One more change request: Please add the description about this refactoring into https://github.com/redhat-developer/vscode-java/blob/master/document/_java.learnMoreAboutRefactorings.md.

And maybe other missing refactorings(e.g. extract interface)

> You may encounter a problem that the document will not appear, a workaround for testing can be found at https://github.com/redhat-developer/vscode-java/issues/2974

CsCherrYY:
description added in https://github.com/redhat-developer/vscode-java/pull/2967/commits/d419eab47d5bb70c3b4168897660d5f04cc0e40f

> And maybe other missing refactorings(e.g. extract interface)

Let me create another PR to add the missing refactoring descriptions.

jdneo:
Some feedback:

1. If set the access modifier to `package-private`, the signature preview will show `package-private`. Should be without any keyword.
2. Change the name of the parameter. After clicking `reset`, the name didn't reset to the original one.
3. Click the column header will highlight it, is this by design?    
image
4. I'm a little bit confused by the `Default value`, could be helpful if some tooltip can be provided.

jdneo:
I cannot change the added parameter name now

CsCherrYY:
> I cannot change the added parameter name now

My bad, just using same style for editing and non-editing parameter cell. https://github.com/redhat-developer/vscode-java/pull/2967/commits/1cc199b255f61c085fab96832e2be98f9972d58f fixes this.

jdneo:
Click `Keep original method as delegate to changed method`, then click `Reset`. Nothing happens.

Looks like something wrong with the status handling.

jdneo (approved):
Looks good to me. We can experiment it in the insider before the next month's stable release.

fbricon:
You can install the latest vscode-java pre-release

xieu90:
i dont really see vscode-java pre-release in extension search window

image

i tried to change the java pack to pre-release and dont see change signature refactoring either.

xieu90:
it works now, i can see it, thank you ^^

PR #2978, opened 2023-03-02
Fix regression in API impacting Codespaces

### LINKED ISSUE #2968: v1.15.x causes Java to be non-functional in GitHub Codespaces
Running and debugging Java projects stopped working in Codespaces with the release of v1.15.00. Clicking the Run button on a project while in the web-based Codespaces results in the panel below appearing and hanging in an endless loop. The project never runs.
Screenshot 2023-02-27 at 10 59 04 AM

Downgrading to v1.14.0 solves the problem. Upgrading to the pre-release does not resolve the problem.

Related issue: https://github.com/microsoft/vscode-java-debug/issues/1148

##### Environment
- Operating System: GitHub Codespaces
- JDK version: ??
- Visual Studio Code version: latest

##### Steps To Reproduce
1. Start a codespace from [https://github.com/LinkedInLearning/practice-it-java-3086189](https://github.com/LinkedInLearning/practice-it-java-3086189)
2. Go to folder `_01_02e` and open `Employee.java`.
3. Click the play button in the upper right corner to run Java.

##### Current Result
The above modal appears in the bottom right-hand corner and loops endlessly. Nothing else happens.

##### Expected Result
Java runs and output is produced in the terminal.

datho7561:
The API wouldn't report lightweight mode correctly when it was forced to start in lightweight due to vscode-java detecting it was being run in Codespaces. This broke the mechanism to switch to standard mode in Codespaces.

Fixes #2968

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,8 +5,8 @@ import * as fs from 'fs';
 import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import { CodeActionContext, CodeActionTriggerKind, commands, ConfigurationTarget, Diagnostic, env, EventEmitter, ExtensionContext, extensions, IndentAction, InputBoxOptions, languages, RelativePattern, TextDocument, UIKind, Uri, ViewColumn, window, workspace, WorkspaceConfiguration } from 'vscode';
-import { CancellationToken, CodeActionParams, CodeActionRequest, Command, DidChangeConfigurationNotification, ExecuteCommandParams, ExecuteCommandRequest, LanguageClientOptions, RevealOutputChannelOn, State } from 'vscode-languageclient';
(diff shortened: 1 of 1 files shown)


datho7561:
To test this PR:
1. Make sure you are in Codespaces
2. Use the default settings, which should mean that vscode-java forces LightWeight mode
3. Make sure that the java language server workspace is clean (I wasn't able to reproduce the issue if I had successfully opened the workspace in standard mode already)
4. Every time Codespaces restarts, it will uninstall all the `.vsix` that were manually installed from your computer, so you will need to install the `.vsix` at each launch

rgrunber (approved):
Looks like this is just restoring tbe part of https://github.com/redhat-developer/vscode-java/pull/2905/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626L375-L378 that updated the server mode.
Feel free to merge when ready.

PR #2985, opened 2023-03-07
Add support for .class files

### LINKED ISSUE #2679: Add support for .class files
Currently, you can't open .class files in vscode. To be clear I don't mean the "Go To Definition" action I mean plain .class files. If you try and open one then it'll give you a warning. I've found two extensions that sound like they solved the problem but don't:
1. [Java Decompiler](https://marketplace.visualstudio.com/items?itemName=dgileadi.java-decompiler) on "Go To Definition" decompiles the file if it's not already.
2. [Java CFR Decompiler](https://marketplace.visualstudio.com/items?itemName=perennialinnocence.java-cfr-decompiler) decompiles the file into a new .java file but doesn't open it in a read-only editor which means I can edit it and every time I want to close it I get a popup saying "Do you want to save changes?".

The conclusion is that there is no convenient way of opening a decompiled .class file in vscode.

JessicaJHee:
Fixes #2679 

Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2515

### DIFF
--- a/src/javaClassEditor.ts
+++ b/src/javaClassEditor.ts
@@ -1,0 +1,67 @@
+import path = require('path');
+import * as vscode from 'vscode';
+import { Uri, window, ExtensionContext} from "vscode";
+import { getNonce } from "./webviewUtils";
+
+class JavaClassDocument implements vscode.CustomDocument {
+	constructor(uri: Uri) { this.uri = uri; }
+	uri: Uri;
+	dispose(): void { }
+}
+
+export class JavaClassEditorProvider implements vscode.CustomReadonlyEditorProvider {
+
+	private context: ExtensionContext;
+
+	openCustomDocument(uri: Uri, openContext: vscode.CustomDocumentOpenContext, token: vscode.CancellationToken): JavaClassDocument {
+		return new JavaClassDocument(uri);
+	}
+
+	constructor (context: ExtensionContext) {
+		this.context = context;
+	}
+
+	public static readonly viewType = 'decompiled.javaClass';
+
+	async resolveCustomEditor(document: vscode.CustomDocument, webviewPanel: vscode.WebviewPanel, token: vscode.CancellationToken): Promise<void> {
+		const nonce: string = getNonce();
+		webviewPanel.webview.options = {
+			enableScripts: true,
+			localResourceRoots: [Uri.joinPath(Uri.parse(this.context.extensionPath), 'webview-resources')]
+		};
+		const classUri = Uri.parse((document.uri.toString()).replace(/^file/, "class"));
+		const styleUri = Uri.file(
+			path.join(this.context.extensionPath, 'webview-resources', 'button.css')
+		);
+		const style: string = `<link rel="stylesheet" type="text/css" href="${webviewPanel.webview.asWebviewUri(styleUri).toString()}">`;
+		webviewPanel.webview.html = `
+		<html lang="en">
+		<head>
+			<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'nonce-${nonce}'; style-src ${webviewPanel.webview.cspSource};">
+			${style}
+		</head>
+		<body>
+			<div class="center">
+				<p>This file is not displayed in the text editor because it is a Java class file. Click here to decompile and open.</p>
+				<button id="btn"><center>Decompile Class File</center></button>
+			<div>
+			<script nonce="${nonce}">
+				const vscode = acquireVsCodeApi();
+				document.getElementById("btn").addEventListener("click", decompiled);
+				function decompiled() {
+					vscode.postMessage({ command: 'decompiled' });
+				}
+			</script>
+		</body>
+		</html>
+		`;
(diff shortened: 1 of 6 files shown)


JessicaJHee:
Here is the current behavior of this PR: 

Peek 2023-03-10 18-03

src/javaClassEditor.ts
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	private static readonly viewType = 'decompiled.javaClass';
+
+	async resolveCustomEditor(document: vscode.CustomDocument, webviewPanel: vscode.WebviewPanel, token: vscode.CancellationToken): Promise<void> {
+		const nonce: string = this.getNonce();
+		webviewPanel.webview.options = {
+			enableScripts: true,
+		};
+		const classUri = Uri.parse((document.uri.toString()).replace(/^file/, "class").replace(/.class/, "_decompiled.class"));
```
rgrunber, line 32: Is there any reason not to map `file:/some/path/to/file.class` -> `class:/some/path/to/file.class` ? It looks cleaner when the custom editor (shown as file.class in the tab) gets replaced by the same file name.

src/javaClassEditor.ts
```diff
@@ -0,0 +1,77 @@
...
+		const styleUri = vscode.Uri.file(
```
rgrunber, line 33: This can just be `Uri.file(...)` right ?

src/javaClassEditor.ts
```diff
@@ -0,0 +1,77 @@
...
+		<html lang="en">
+		<head>
+			${style}
+		</head>
+		<body>
+			<div class="center">
+				<p>This file is not displayed in the text editor because it is a Java class file. Click here to decompile and open.</p>
+				<button onclick="decompiled()"><center>Decompile Class File</center></button>
+			<div>
+			<script nonce="${nonce}">
```
rgrunber, line 47: In order for this to be used, I think you need the following check : https://github.com/microsoft/vscode-extension-samples/blob/main/webview-sample/src/extension.ts#L196 .

Also maybe restricting the webview to only resources from a list of specific locations https://github.com/microsoft/vscode-extension-samples/blob/main/webview-sample/src/extension.ts#L43 .

src/providerDispatcher.ts
```diff
@@ -81,6 +84,29 @@ function createJDTContentProvider(options: ProviderOptions): TextDocumentContent
...
+		provideTextDocumentContent: async (uri: Uri, token: CancellationToken): Promise<string> => {
+			const languageClient: LanguageClient | undefined = await getActiveLanguageClient();
+
+			if (!languageClient) {
+				return '';
+			}
+
+			let decompiledContent: string;
+			try {
+				const originalUri = (uri.toString()).replace(/^class/, "file").replace(/_decompiled.class/, ".class");
```
rgrunber, line 99: `_decompiled.class` can also be removed here since we no longer map to it.

src/javaClassEditor.ts
JessicaJHee: Fixed! The idea before was to keep both files open

src/javaClassEditor.ts
```diff
@@ -0,0 +1,80 @@
...
+			switch (message.command) {
+				case 'decompiled':
+					webviewPanel.dispose();
+					window.showTextDocument(classUri, { preview: false });
+					return;
+			}
+		}, undefined, this.context.subscriptions);
+	}
+
+	private getNonce(): string {
```
rgrunber, line 69: Can you use https://github.com/redhat-developer/vscode-java/pull/2967/files#diff-f5df392ec8cf0691790829bce93e0d8e6231c602b5b3798e432b4ba82b66d1f6R7 (recently added) ?

src/providerDispatcher.ts
```diff
@@ -81,6 +84,29 @@ function createJDTContentProvider(options: ProviderOptions): TextDocumentContent
...
+				return '';
+			}
+
+			let decompiledContent: string;
+			try {
+				const originalUri = uri.toString().replace(/^class/, "file");
+				decompiledContent = await commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.GET_DECOMPILED_SOURCE, originalUri);
+				return decompiledContent;
+			} catch (error) {
+				console.log(`Error while getting decompiled source : ${error}`);
```
rgrunber, line 103: I would just replace `${error}` with `${originalUri}` (move it above the `try`).

src/javaClassEditor.ts
```diff
@@ -0,0 +1,67 @@
+import path = require('path');
+import * as vscode from 'vscode';
+import { Disposable, Uri, window, ExtensionContext} from "vscode";
```
rgrunber, line 3: `Disposable` doesn't seem to be used. Feel free to remove.

rgrunber (approved):
One very minor thing, and then just fix up the language-server side command at https://github.com/eclipse/eclipse.jdt.ls/pull/2515#pullrequestreview-1346508364 . Then I think we can merge.

PR #2991, opened 2023-03-09
Permit computation of method hierarchy through type hierarchy API.

rgrunber:
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2502

- Method hierarchy needs to pass along the method signature to be searched on each type
- Use data field to store the method signature being searched, if any
- Use SymbolKind.Null to denote a type in the hierarchy that does not implement the given method

method-hierarchy


rgrunber:
CC @CsCherrYY

CsCherrYY (approved):

PR #2992, opened 2023-03-09
Create .devcontainer/devcontainer.json

qapris:


rgrunber:
@testforstephen , has anyone played around with [this](https://code.visualstudio.com/docs/devcontainers/containers) ? Looking at https://containers.dev/implementors/json_reference/ , I think we'd probably need to support some port forwarding options (for debugging the language server), and environment variables to ensure the language server under development is used (eg. https://github.com/redhat-developer/vscode-java/blob/de77dc838325e2a75d91407d28eb8fb85252489c/.vscode/launch.json#L12).

PR #2996, opened 2023-03-13
Add API to track lsp perf on language client level

testforstephen:
This PR provides an API to easily trace the lsp request performance.
- Track `initialize` request on languageClient.start()
- Track common requests on `sendRequest(...)`

Here is a sample code on how to use this API:

```typescript
 const javaExt = vscode.extensions.getExtension("redhat.java");
 if (javaExt.exports.onDidRequestEnd) {
    javaExt.exports.onDidRequestEnd((traceEvent: any) => {
      console.log(`${traceEvent.type} - ${traceEvent.duration} ms - ${traceEvent.error}`);
    });
  }
```

### DIFF
--- a/src/TracingLanguageClient.ts
+++ b/src/TracingLanguageClient.ts
@@ -53,7 +53,7 @@ export class TracingLanguageClient extends LanguageClient {
 	}
 
 	private sendRequest0(method: any, ...args) {
-		if (!args) {
+		if (!args || !args.length) {
 			return super.sendRequest(method);
 		}
 
--- a/src/TracingLanguageClient.ts
+++ b/src/TracingLanguageClient.ts
@@ -1,3 +1,4 @@
+import { performance } from "perf_hooks";
 import { Event, EventEmitter } from "vscode";
 import { CancellationToken, LanguageClient, LanguageClientOptions, ProtocolRequestType, ProtocolRequestType0, RequestType, RequestType0, ServerOptions } from "vscode-languageclient/node";
 import { TraceEvent } from "./extension.api";
@@ -15,7 +16,7 @@ export class TracingLanguageClient extends LanguageClient {
 	start(): Promise<void> {
 		const isFirstTimeStart: boolean = !this.isStarted;
 		this.isStarted = true;
-		const startAt: number = Date.now();
+		const startAt: number = performance.now();
 		return super.start().then(value => {
 			if (isFirstTimeStart) {
 				this.fireTraceEvent("initialize", startAt);
@@ -41,7 +42,7 @@ export class TracingLanguageClient extends LanguageClient {
 	sendRequest<R>(method: string, token?: CancellationToken): Promise<R>;
 	sendRequest<R>(method: string, param: any, token?: CancellationToken): Promise<R>;
 	sendRequest(method: any, ...args) {
-		const startAt: number = Date.now();
+		const startAt: number = performance.now();
 		const requestType: string = this.getRequestType(method, ...args);
 		return this.sendRequest0(method, ...args).then(value => {
 			this.fireTraceEvent(requestType, startAt);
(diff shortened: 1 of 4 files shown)


testforstephen:
// @akaroml @Eskibear @CsCherrYY pls help a review on the API design.

src/TracingLanguageClient.ts
```diff
@@ -0,0 +1,98 @@
+import { Event, EventEmitter } from "vscode";
+import { CancellationToken, LanguageClient, LanguageClientOptions, ProtocolRequestType, ProtocolRequestType0, RequestType, RequestType0, ServerOptions } from "vscode-languageclient/node";
+import { TraceEvent } from "./extension.api";
+
+const requestEventEmitter = new EventEmitter<TraceEvent>();
+export const onDidRequestEnd: Event<TraceEvent> = requestEventEmitter.event;
+
+export class TracingLanguageClient extends LanguageClient {
```
Eskibear, line 8: maybe TraceableLanguguageClient is more proper?

src/TracingLanguageClient.ts
```diff
@@ -0,0 +1,98 @@
...
+	private getRequestType(method: any, ...args): string {
+		let requestType: string;
+		if (typeof method === 'string' || method instanceof String) {
+			requestType = String(method);
+		} else {
+			requestType = method?.method;
+		}
+
+		if (requestType === "workspace/executeCommand") {
+			if (args && args[0]?.command) {
```
Eskibear, line 82: can be simplified as `args?.[0]?.command`

src/TracingLanguageClient.ts
```diff
@@ -0,0 +1,98 @@
...
+	sendRequest<R>(method: string, token?: CancellationToken): Promise<R>;
+	sendRequest<R>(method: string, param: any, token?: CancellationToken): Promise<R>;
+	sendRequest(method: any, ...args) {
+		const startAt: number = Date.now();
+		const requestType: string = this.getRequestType(method, ...args);
+		return this.sendRequest0(method, ...args).then(value => {
+			this.fireTraceEvent(requestType, startAt);
+			return value;
+		}, reason => {
+			this.fireTraceEvent(requestType, startAt, reason);
```
Eskibear, line 50: here `reason` can be an Error type instead of a primitive string. While in definition of `TraceEvent` it's a string. Maybe an unexpected `toString()` will be called, not sure if this will be a problem. E.g. we want to know `Error.message`, but `Error.toString()` returns something like "[Object]".

testforstephen: Nice catch. To keep the API code clean, I would change the definition type of `error` in `TraceEvent` as `any`. The invoker should be responsible for how to consume the error information.

src/TracingLanguageClient.ts
testforstephen: I'd perfer to keep using Tracing since tracing is a common usage, similar to logging.

Eskibear: ok

src/TracingLanguageClient.ts
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	sendRequest<R, PR, E, RO>(type: ProtocolRequestType0<R, PR, E, RO>, token?: CancellationToken): Promise<R>;
+	sendRequest<P, R, PR, E, RO>(type: ProtocolRequestType<P, R, PR, E, RO>, params: P, token?: CancellationToken): Promise<R>;
+	sendRequest<R, E>(type: RequestType0<R, E>, token?: CancellationToken): Promise<R>;
+	sendRequest<P, R, E>(type: RequestType<P, R, E>, params: P, token?: CancellationToken): Promise<R>;
+	sendRequest<R>(method: string, token?: CancellationToken): Promise<R>;
+	sendRequest<R>(method: string, param: any, token?: CancellationToken): Promise<R>;
+	sendRequest(method: any, ...args) {
+		const startAt: number = Date.now();
```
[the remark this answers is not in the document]
testforstephen: thanks. Good to know the new API.

The data will like `textDocument/completion - 1359.5962999993935 ms` with performance.now api. Then the downstream tracer can decide whether to convert the floating point number to integer.

jdneo (approved):
LGTM

Issue #3005, opened 2023-03-18
Extension fails to load with pre-release version


gayanper:
Fail to start the extension

##### Environment
- Operating System: MacOS
- JDK version: 17
- Visual Studio Code version: 1.76.2
- Java extension version: pre-release

##### Steps To Reproduce
1. Install pre-release version
2. Reload vscode

##### Current Result

Extension loading fails with 
```
rejected promise not handled within 1 second: TypeError: expandHomeDir is not a function
stack trace: TypeError: expandHomeDir is not a function
	at Object.<anonymous> (/Users/gayanperera/Projects/vscode/vscode-java/dist/extension.js:80495:28)
	at Generator.next (<anonymous>)
	at fulfilled (/Users/gayanperera/Projects/vscode/vscode-java/dist/extension.js:80459:58)

```

jdneo:
fixed by https://github.com/redhat-developer/vscode-java/pull/3009

PR #3006, opened 2023-03-18
Fix extension loading #3005

gayanper:


gayanper:
@rgrunber @jdneo Let me know if there is another way to solve this

snjeza:
The issue happens when setting `java.jdt.ls.java.home`.
The PR fixes it.

snjeza (approved):

fbricon:
can anyone explain why this patch fixes the issue? It seems wrong, but I suck at JS/TS

gayanper:
I'm not an expert either, but i guess it could be something to do with how import and require load the JS module. 

I doubt if it could be due to a recent electron + NodeJS runtime change.

CsCherrYY:
> can anyone explain why this patch fixes the issue? It seems wrong, but I suck at JS/TS

This issue should be related to the config change of #2967, this PR introduces some project configurations related to webview (react) might be incompatible with existing statements.

I created a new PR for this #3009 , which might resolve this issue and avoid any potential issue in the future

jdneo:
To add more information: https://www.typescriptlang.org/tsconfig#esModuleInterop

jdneo:
With https://github.com/redhat-developer/vscode-java/pull/3009, we can close this PR.

PR #3009, opened 2023-03-20
Add separate tsconfig for webview

CsCherrYY:
With this PR, we have a basic `tsconfig.base.json`, two configurations `tsconfig.json` (for extension) and `tsconfig.webview.json` (for webview) specifies the source configuration and source directories.

will address the side effects from #2967 , which might cause configuration problems related to tsconfig rule `esModuleInterop`.


jdneo (approved):
LGTM, it makes sense to sperate the config for different source sets.

Dustin4444 (commented):
#3469

PR #3010, opened 2023-03-20
Trace API: give indicator of response success status

testforstephen:
One use case for this API is to trace completion performance. 

To support the feature [Auto complete after new keyword #2010](https://github.com/eclipse/eclipse.jdt.ls/pull/2010) ("new |"), we enabled completion trigger characters on whitespace. However, this also triggers many completion requests when typing whitespace for indentation. These requests are invalid and jdtls returns an empty array. We need a way to distinguish such noise data during tracing.

### DIFF
--- a/src/TracingLanguageClient.ts
+++ b/src/TracingLanguageClient.ts
@@ -19,12 +19,12 @@ export class TracingLanguageClient extends LanguageClient {
 		const startAt: number = performance.now();
 		return super.start().then(value => {
 			if (isFirstTimeStart) {
-				this.fireTraceEvent("initialize", startAt);
+				this.fireTraceEvent("initialize", startAt, undefined);
 			}
 			return value;
 		}, reason => {
 			if (isFirstTimeStart) {
-				this.fireTraceEvent("initialize", startAt, reason);
+				this.fireTraceEvent("initialize", startAt, undefined, reason);
 			}
 			throw reason;
 		});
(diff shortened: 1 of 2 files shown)


jdneo (approved):

src/TracingLanguageClient.ts
```diff
@@ -88,12 +88,21 @@ export class TracingLanguageClient extends LanguageClient {
...
 			type,
 			duration,
 			error: reason,
+			resultLength,
 		});
 	}
+
+	private getResultLength(value: any): number | undefined {
+		if (!value) {
+			return 0;
```
Eskibear, line 103: should we return `undefined` here?

testforstephen: If the response is undefined, it indicates that there are no results and returning 0 is acceptable. For me, returning `undefined` means that we have not yet determined whether the response contains valid results.

Eskibear: Ok, it makes sense. Then we probably need to revisit line 106. If value = `{ someKey: "some value" }`, it now returns undefined, is that expected?

If 0 is expected in above case, shall we append `?? 0` like below?
```ts
return value?.length ?? value?.items?.length ?? 0;
```

testforstephen: > Then we probably need to revisit line 106. If value = { someKey: "some value" }, it now returns undefined, is that expected?

Yes. We can return `undefined` for now. Maybe one day there will be a requirement to resolve some response values more accurately and we can improve it then.

Eskibear: OK, LGTM then.

PR #3013, opened 2023-03-21
Remove unexpected webview for goToDefinition requests

### LINKED ISSUE #3011: Dependency classes are not longer directly opened
Following https://github.com/redhat-developer/vscode-java/pull/2985, opening a non-project type no longer directly opens the source, but shows the intermediary "Decompile Class File" webview. This is a huge regression in DX.

##### Steps To Reproduce
1. In a Java file, Ctrl+Click on `String`

##### Current Result
"Decompile Class File" webview shows up instead of proper JDK source as before

##### Expected Result
Attached or decompiled sources should  shows up directly during code navigation/references/..., as before

##### Additional Informations
The "Decompile Class File" webview should only show up when manually opening a .class file from the IDE

JessicaJHee:
Fixes #3011

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,8 +63,7 @@
         "displayName": "Decompiled Java Class File",
         "selector": [
           {
-            "scheme": "file",
-            "filenamePattern": "*.class"
+            "filenamePattern": "file:/**/*.class"
           }
         ]
       }


rgrunber (commented):
This fixes the issue for me. @fbricon if you're fine with this.

fbricon:
Thanks @JessicaJHee & @rgrunber

Issue #659, opened 2018-10-01
Organize Imports on Save


Catscratch:
Hi,

how may I organize imports on file save?

The setting "java.saveActions.organizeImports": true does nothing.

##### Environment
- Operating System: Windows 10
- JDK version: OpenJDK 1.8.0_172
- Visual Studio Code version: 1.27.2 (user setup)
- Java extension version: please...what?

##### Steps To Reproduce
1. Go to settings
2. add "java.saveActions.organizeImports": true
3. Open a java file
4. insert a non used import like "import java.awt.color.*;"
5. Hit Ctrl + S
6. See that nothing happens
7. Hit Ctrl + Alt + O
8. See the import is cleaned

##### Current Result

Cleanup not performed on save.

##### Expected Result

Cleanup should be performed on save.

##### Additional Informations

I opened a Java Maven project (Eclipse project).

fbricon:
works for me. Try enabling verbose logging in your settings.json (`"java.trace.server": "verbose"`), and paste the logs here

btw 
> Java extension version: please...what?
<img width="602" alt="screen shot 2018-10-03 at 6 21 30 pm" src="https://user-images.githubusercontent.com/148698/46442885-33beb680-c739-11e8-8b8b-71750de19cf6.png">
-> 0.32.0 here

Catscratch:
Thanks for your reply. I got 0.31.0 and now upgraded to 0.32.0.

I also enabled verbose logging. But I got no log output on save.
(I used Help -> Switch Developer Tools -> Console -> All levels)

ThatNerdyPikachu:
bump, same issue here.

in the logs, the extension is actually receiving the message to organize the imports, but that isn't doing anything..

tan31989:
This issue is still seen in the latest vscode editor.

bundery:
This issue is still seen in the latest vscode editor.

fbricon:
@snjeza I believe this is what I saw in https://github.com/eclipse/eclipse.jdt.ls/pull/1415#issuecomment-617941928. My guess is the Compilation Unit is not reconciled properly at that point.

snjeza:
@bundery could you try the following settings:

```
"editor.codeActionsOnSave": {
		"source.organizeImports": true
    },
    "java.saveActions.organizeImports": false
```

msgeissler:
I have also never been able to get this settings to work. This includes two completely separate installations and maven projects.

The workaround / tip provided by @snjeza does work.

[...]

PR #3015, opened 2023-03-22
deprecate 'java.saveActions.organizeImports'

### LINKED ISSUE #2207: "java.saveActions.organizeImports" setting doesn't work on Windows
This issue is originally created by @alx1024 at https://github.com/microsoft/vscode-java-debug/issues/1071
> [provide a description of the issue]
> 
> ##### Environment
> * Operating System: Win 10
> * JDK version: 11
> * Visual Studio Code version: 1.62.0
> * Java extension version: 1.0.0
> * Java Debugger extension version: 0.36.0
> 
> ##### Steps To Reproduce
> 1. In any Java Program in any main-function type "File f = new File("/");"
> 2. Save File
> 
> ##### Current Result
> Even with checkbox checked in settings menu "Java > Save Actions: Organize Imports" (which results in
> 
> "java.saveActions.organizeImports": true
> 
> in settings.json): imports are not automatically organized
> 
> ##### Expected Result
> Imports are automatically organized on file save with setting checked
> 
> ##### Additional Informations
> if you modify the settings.json manually to:
> 
> "editor.codeActionsOnSave": { "source.organizeImports": true }
> 
> regardless if Java > Save Actions-option is checked, imports work, but you have to search for this solution on Stackoverflow

CsCherrYY:
fix #2207, #659 

A recommended way to organize imports in VS Code should be configuring generic `editor.codeActionsOnSave`. In this way, the java specific configuration can be deprecated to avoid any misunderstanding (it also seems to not work now).

### DIFF
--- a/package.json
+++ b/package.json
@@ -447,6 +447,7 @@
           "type": "boolean",
           "default": false,
           "description": "Enable/disable auto organize imports on save action",
+          "deprecationMessage": "This setting is deprecated, please configure 'editor.codeActionsOnSave' instead.",
           "scope": "window"
         },
         "java.import.exclusions": {


rgrunber:
Agreed. Not only that, but we also have `java.cleanup.actionsOnSave` which is LS supported, so it will work for clients other than VS Code. If you just add to https://github.com/eclipse/eclipse.jdt.ls/blob/702d59eb54f6bdd5e8bf1efbec3db2856b3a6339/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/cleanup/CleanUpRegistry.java#L42-L52 and are able to get organize imports to support the https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/cleanup/ISimpleCleanUp.java interface.

rgrunber:
Yup, something as simple as https://github.com/rgrunber/eclipse.jdt.ls/commit/2497722e9edac8f9e1a348723c94a9a8c3b55023 seems to work. You can then just set `"java.cleanup.actionsOnSave": ["organizeImports"]` in VS Code settings (it will complain because it isn't a known value but the LS will respect it).

We could definitely look at ways to integrate with `editor.codeActionsOnSave` since that works based on the code action id/kind and opens up possibilities to even more save actions. The main downside is the setting is only for VS Code.

binabdulkhalim-prog (changes requested):
- [ ]

fbricon:
Merged the deprecation change. It should discourage users from using it. It's gone from the settings UI and manual editing shows a warning:

Screenshot 2026-02-16 at 12 56 50

Other cleanups/refactorings could be done separately.

PR #3018, opened 2023-03-22
Support constructor access modifiers in snippet using choice values.

borellif:
This PR adds a new vscode snippet called 'pctor' - adding a private constructor. Complies with sonarlint(java:S1118)

### DIFF
--- a/snippets/java.json
+++ b/snippets/java.json
@@ -8,14 +8,14 @@
 		],
 		"description": "Public static main method"
 	},
-	"Public constructor": {
+	"Constructor": {
 		"prefix": "ctor",
 		"body": [
-			"public ${1:${TM_FILENAME_BASE}}($2) {",
-			"\t${3:super();}$0",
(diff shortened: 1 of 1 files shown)


snippets/java.json
```diff
@@ -17,6 +17,15 @@
 		],
 		"description": "Public constructor"
 	},
+	"Private constructor": {
+		"prefix": "pctor",
+		"body": [
+			"private ${1:${TM_FILENAME_BASE}}($2) {",
```
rgrunber, line 23: This seems very similar to the `ctor` snippet defined directly above the one you wish to introduce. Would it make more sense to re-use `ctor` and turn the "public" term into a [snippet choice](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_choice) ?

eg. `${1|public,protected,private|} ${2: ...`

rgrunber:
constructor-snippet-new

PR #3028, opened 2023-03-28
Recover when 'documentPaste' API is not properly registered.

rgrunber:
Certain VS Code based clients (eg. VSCodium, or any rebuild of VSCode with different product file) may not have registered the document paste event API, so we should continue gracefully.

```
2023-03-28 18:12:57.959 [error] [Extension Host] Notification handler 'language/status' failed with message: Extension 'redhat.java' CANNOT use API proposal: documentPaste.
Its package.json#enabledApiProposals-property declares: [] but NOT documentPaste.
 The missing proposal MUST be added and you must start in extension development mode or use the following command line switch: --enable-proposed-api redhat.java
```

CC: @CsCherrYY let me know if this makes sense.

### DIFF
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -128,8 +128,12 @@ export class StandardLanguageClient {
 							await onExtensionChange(extensions.all);
 						});
 					}
-
-					registerPasteEventHandler(context, this.languageClient);
+					try {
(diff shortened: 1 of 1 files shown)


CsCherrYY:
Tried in my PC, I can confirm that this patch can gracefully catch the exception and there is no such error in the extension host any more.

PR #3034, opened 2023-03-30
Define completionItem@resolveSupport

snjeza:
Related issues: 
- https://github.com/eclipse/eclipse.jdt.ls/pull/2535
- https://github.com/eclipse/eclipse.jdt.ls/pull/2453

### DIFF
--- a/src/TracingLanguageClient.ts
+++ b/src/TracingLanguageClient.ts
@@ -1,6 +1,6 @@
 import { performance } from "perf_hooks";
 import { Event, EventEmitter } from "vscode";
-import { CancellationToken, LanguageClient, LanguageClientOptions, ProtocolRequestType, ProtocolRequestType0, RequestType, RequestType0, ServerOptions } from "vscode-languageclient/node";
+import { CancellationToken, InitializeParams, LanguageClient, LanguageClientOptions, ProtocolRequestType, ProtocolRequestType0, RequestType, RequestType0, ServerOptions } from "vscode-languageclient/node";
 import { TraceEvent } from "./extension.api";
 
 const requestEventEmitter = new EventEmitter<TraceEvent>();
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -5,7 +5,7 @@ import { findRuntimes } from "jdk-utils";
 import * as net from 'net';
 import * as path from 'path';
 import { CancellationToken, CodeActionKind, commands, ConfigurationTarget, DocumentSelector, EventEmitter, ExtensionContext, extensions, languages, Location, ProgressLocation, TextEditor, Uri, ViewColumn, window, workspace } from "vscode";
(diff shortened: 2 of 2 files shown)


jdneo:
Hi @dbaeumer, could you confirm that when sending the initialize request, simply adding `textEdit` property to the completion's `resolveSupport` can allow us to update the `textEdit` during `completion/resolve` phase?

Or the extension should not modify that during initialization?

src/TracingLanguageClient.ts
```diff
@@ -13,6 +13,16 @@ export class TracingLanguageClient extends LanguageClient {
...
 	}
 
+	protected fillInitializeParams(params: InitializeParams) {
+		super.fillInitializeParams(params);
+		// https://github.com/eclipse/eclipse.jdt.ls/pull/2453#issuecomment-1429022121
+		const properties: string[] = params.capabilities.textDocument.completion.completionItem.resolveSupport.properties;
+		properties.push('textEdit');
+		properties.push('sortText');
+		properties.push('filterText');
+		properties.push('insertText');
```
testforstephen, line 23: This is a hack for me. The resolveSupport client capabilities should be determined by VS Code, not by the extension itself.

Eskibear:
> Or the extension should not modify that during initialization?

I don't think so. Even if you modify the value, vscode as a client still has no capability to resolve these fields.

src/TracingLanguageClient.ts
snjeza: I have updated the PR. Now, it uses `LanguageClient.registerFeature`

rgrunber: @testforstephen, so then should these changes eventually be hard-coded into vscode-languageserver-node client at https://github.com/microsoft/vscode-languageserver-node/blob/56c23c557e3568a9f56f42435fd5a80f9458957f/client/src/common/completion.ts#L84-L86 ? I think we just want to use these to accurately reflect that it's supported to the server.

Eskibear: And https://github.com/eclipse/eclipse.jdt.ls/issues/1864#issuecomment-923990085 roughly described why vscode doesn't support to resolve other fields like `textEdit`.

The ideal process of solving it in my mind would be:
- dig into vscode's impl, figure out current limitation from resolving other fields
- change vscode's impl to make sure it's working
- update hard-coded value of resolveSupport in vscode-language-client lib
- use the new lib here

testforstephen: @Eskibear has explained the reason well. VS Code is the consumer of languageclient and be responsible for this resolving behavior. If we really want to move forward on this direction, we need to go into the upstream VS Code implementation to see how to make it to support new lazy-resolvable properties.

dbaeumer:
No, that is not the idea. It is a client capability telling the server that a certain proper can be lazy so a server can leave it out on the `get` request.

PR #3045, opened 2023-04-04
Don't send settings with all telemetry events

### LINKED ISSUE #3044: Java settings don't need to be sent with all telemetry events
Java settings don't need to be sent with all telemetry events, but they are:
https://github.com/redhat-developer/vscode-java/blob/30cbd5bd23f39bc0a788608066058c7d05faa78f/src/telemetry.ts#L49

Does this make sense, from a typescript POV?
```typescript
const javaSettings = getNonDefaultJavaSettings(workspace.getConfiguration());

let properties: any;
if (eventName === STARTUP_EVT) {
	serverInitializedReceived = true;
	properties= { ...data, ...javaSettings };
} else {
	properties= { ...data};
}

return telemetryManager.send({
	name: eventName,
	properties
});
```

fbricon:
Fixes #3044

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -38,15 +38,19 @@ export namespace Telemetry {
 		if (!telemetryManager) {
 			throw new Error("The telemetry service for vscode-java has not been started yet");
 		}
+		const javaSettings = getJavaSettingsForTelemetry(workspace.getConfiguration());
+
+		let properties: any;
 		if (eventName === STARTUP_EVT) {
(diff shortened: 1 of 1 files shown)


fbricon:
> Change looks fine to me. Can you just rename `getNonDefaultJavaSettings` to something like `getJavaSettingsForTelemetry` ? Initially I may have wanted just the setting if they aren't the default, but it's probably easier to just emit the settings regardless of default or not.

done

Issue #2974, opened 2023-03-02
The refactoring document stops working


jdneo:
##### Environment
Version: 1.76.0 (user setup)
Commit: 92da9481c0904c6adfe372c12da3b7748d74bdcb
Date: 2023-03-01T10:22:44.506Z
Electron: 19.1.11
Chromium: 102.0.5005.196
Node.js: 16.14.2
V8: 10.2.154.26-electron.0
OS: Windows_NT x64 10.0.19045
Sandboxed: Yes

##### Steps To Reproduce
Click the refactor document link:
![image](https://user-images.githubusercontent.com/6193897/222312322-930c8ea6-7fd7-4112-94db-7430d5133bd0.png)

##### Current Result
Nothing happens

##### Expected Result
Show the refactoring document

##### Additional Informations
This should be a bug from upstream VS Code: https://github.com/microsoft/vscode/issues/175737.

Now we have two options for this issue:
1. Comment out https://github.com/redhat-developer/vscode-java/blob/85eedfe0e6b51c0393694be600b766b6e1bc25a6/src/codeActionProvider.ts#L35
This will make the document appear, but we lose the capability to directly show the related section based on current refactoring context.
3. Wait for VS Code to fix this.

jdneo:
Update: Upstream issue is fixed. We can remove the workaround once new VS Code is released.

PR #3052, opened 2023-04-10
Remove the workaround to show the refactoring document

jdneo:
fix https://github.com/redhat-developer/vscode-java/issues/2974

Now the document can go to the related section automatically.

refactor-doc

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "virtualWorkspaces": false
   },
   "engines": {
-    "vscode": "^1.74.0"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "virtualWorkspaces": false
   },
   "engines": {
-    "vscode": "^1.74.0"
+    "vscode": "^1.77.0"
```
jdneo, line 26: The upstream fix is in `1.77` so I update the field as well

PR #3053, opened 2023-04-10
Re-validate current document if the focus is switched to a Java file

testforstephen:
this requires https://github.com/eclipse/eclipse.jdt.ls/pull/2587

### DIFF
--- a/src/diagnostic.ts
+++ b/src/diagnostic.ts
@@ -6,7 +6,7 @@ import { validateAllOpenBuffersOnChanges } from "./settings";
 export function registerDocumentValidationListener(context: ExtensionContext, languageClient: LanguageClient) {
 	context.subscriptions.push(window.onDidChangeActiveTextEditor(textEditor => {
 		// Refresh the diagnostics when the focus is switched to a Java file.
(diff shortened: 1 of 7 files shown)


src/diagnostic.ts
```diff
@@ -0,0 +1,19 @@
+import { ExtensionContext, window } from "vscode";
+import { ValidateDocumentNotification } from "./protocol";
+import { LanguageClient } from "vscode-languageclient/node";
+import { validateAllOpenBuffersOnChanges } from "./settings";
+
+export function registerDocumentValidationListener(context: ExtensionContext, languageClient: LanguageClient) {
+	context.subscriptions.push(window.onDidChangeActiveTextEditor(textEditor => {
+		// Refresh the diagnostics when the focus is switched to a Java file.
+		if (textEditor.document.uri?.scheme === "file" && textEditor.document.languageId === "java") {
```
jdneo, line 9: `textEditor` might be undefined

export const onDidChangeActiveTextEditor: Event<TextEditor | undefined>;

testforstephen: nice catch.

jdneo (approved):
LGTM

PR #3057, opened 2023-04-11
Ignore .github/ and .gitignore from packaging

fbricon:

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -26,3 +26,7 @@ Jenkinsfile
 .eslintignore
 webpack.config.js
 .DS_Store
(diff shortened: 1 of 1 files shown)


rgrunber (approved):
This is fine to me. I don't see any files under `.github/` or `.gitignore` itself end up in the vsix when i package it though. Is it a platform specific issue or just for the sake of clarity that you have them there.

Eskibear:
LGTM to explicit exclude them. 
git related folders might already be excluded by default when packaging using vsce. But I still see `.sdkmanrc` there, which is not needed.

PR #3058, opened 2023-04-11
Record 'java.clean.workspace' invocations in telemetry

fbricon:

### DIFF
--- a/USAGE_DATA.md
+++ b/USAGE_DATA.md
@@ -17,6 +17,7 @@ vscode-java has opt-in telemetry collection, provided by [vscode-redhat-telemetr
  * Information about the following settings. In the case of settings that store a well defined value (eg. path/url/string), we simply collect whether the setting has been set.
    * `java.settings.url`, `java.format.settings.url`, `java.quickfix.showAt`, `java.symbols.includeSourceMethodDeclarations`, `java.completion.guessMethodArguments`, `java.completion.postfix.enabled`, `java.cleanup.actionsOnSave`, `java.sharedIndexes.enabled`, `java.inlayHints.parameterNames.enabled`, `java.server.launchMode`, `java.autobuild.enabled`
  * The extension name and the choice made when a recommendation to install a 3rd party extension is proposed
+ * The name of Java commands being manually executed, and any resulting errors
 
 ## What's included in the general telemetry data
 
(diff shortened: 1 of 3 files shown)


USAGE_DATA.md
```diff
@@ -17,6 +17,7 @@ vscode-java has opt-in telemetry collection, provided by [vscode-redhat-telemetr
  * Information about the following settings. In the case of settings that store a well defined value (eg. path/url/string), we simply collect whether the setting has been set.
    * `java.settings.url`, `java.format.settings.url`, `java.quickfix.showAt`, `java.symbols.includeSourceMethodDeclarations`, `java.completion.guessMethodArguments`, `java.completion.postfix.enabled`, `java.cleanup.actionsOnSave`, `java.sharedIndexes.enabled`, `java.inlayHints.parameterNames.enabled`, `java.server.launchMode`, `java.autobuild.enabled`
  * The extension name and the choice made when a recommendation to install a 3rd party extension is proposed
+ * The name of Java commands being manually executed
```
rgrunber, line 20: For the sake of completeness, we should probably put "and any resulting errors"

src/extension.ts
```diff
@@ -291,13 +292,16 @@ export function activate(context: ExtensionContext): Promise<ExtensionAPI> {
...
+				const data = {};
 				try {
 					cleanupLombokCache(context);
 					deleteDirectory(workspacePath);
 					deleteDirectory(syntaxServerWorkspacePath);
 				} catch (error) {
+					data['error'] = getMessage(error);
 					window.showErrorMessage(`Failed to delete ${workspacePath}: ${error}`);
 				}
+				await Telemetry.sendTelemetry('java.clean.workspace', data);
```
rgrunber, line 304: I would probably use `Commands.CLEAN_WORKSPACE` here.

rgrunber (commented):
Looks good overall, just some minor points.

I think once we get more of these, we should probably go with an approach similar to https://github.com/redhat-developer/vscode-quarkus/blob/0448e38c235ef0368057f7cc2ad021d52e577df3/src/commands/registerCommands.ts#L71 for commands.

fbricon:
> https://github.com/redhat-developer/vscode-quarkus/blob/0448e38c235ef0368057f7cc2ad021d52e577df3/src/commands/registerCommands.ts#L71

Sure, but clean workspace is special, for now, a the event is not sent on shutdown for some reason

Issue #2955, opened 2023-02-20
References to google-styleguide in formatter documentation


fvclaus:
# Problem
The official documentation page about linting has [a section](https://code.visualstudio.com/docs/java/java-linting#_applying-formatter-settings) about formatting which includes a link to the unmaintened/abandoned google/styleguide repo which might be misleading to some unfamiliar with the differences between google/styleguide and google-java-format. 

## A bit more explanation
The project, I currently work on uses the Intellij google-java-format plugin (like almost any other project, I have worked on in the past 6+ years). I know that the section above links to [google/styleguide](https://github.com/google/styleguide) repo, which is not the same as [google/google-java-format](https://github.com/google/google-java-format) repo. Not being overly familiar with the repos, I thought that the formatting output should be the same (or at least very similiar), especially, because the offical documentation links to it.

After spending a few hours trying to get a working solution, it seems to me, that the google/styleguide java style is abandoned (last updated 5 years ago), without an offical statement. I found [this issue](https://github.com/google/styleguide/issues/687) which funnily enough also mentions the vscode-java docs.

The formatting output is very different. Here is a small example:

google/styleguide:
![image](https://user-images.githubusercontent.com/2384738/220123502-744d98d0-5307-441f-aea5-b93d62f81437.png)

google/google-java-format:
![image](https://user-images.githubusercontent.com/2384738/220123602-1bf24cba-d515-42c4-b746-24ce10eb8f55.png)

# Suggestion
Can the reference to the google styleguide be removed? I think it would be bad idea to just mention that the formatter is unmaintained, because people don't read carefully enough, instead just copy the configuration.

I know the referenced documentation is not part of this repo, but I (as somebody unaffiliated with vscode-java) didn't want to open an issue on the vscode repo straight away. 

## google-java-format Solution
[This stackoverflow question](https://stackoverflow.com/questions/58697625/google-styleguide-java-format-for-vscode/75510069) lists different solutions.

There are a lot of plugins that are floating around (mostly forks of one plugin) where I tried a few, but none worked: https://marketplace.visualstudio.com/search?term=google%20java%20format&target=VSCode&category=All%20categories&sortBy=Relevance

CsCherrYY:
[...]

PR #3062, opened 2023-04-13
Support google-java-format 

### LINKED ISSUE #663: Use google-format-code instead of eclipse ?
Is there a way to use google-format-code instead of the eclipse formatter with vscode-java ?

At the moment i'm forced to use it with the SaveAndRun extension, but this has issues like me doing changes while the file is saved and having to reconcile the buffer as it tries to reload.

I could work on a PR for this if you give a few stating points on how to implement this, it would help ?

Thanks,

CsCherrYY:
fix #663, based on #3059, will rebase when #3059 gets merged.

sample to enable google style:
```
    "java.format.scheme": {
        "style": "google"
    },
```

#2955 can be updated once this PR gets merged.


ringin31:
what is that password

Issue #3046, opened 2023-04-06
Internal compiler error: java.lang.NoClassDefFoundError: org/eclipse/jdt/core/search/SearchPattern


Eskibear:
Open [this simple maven project](https://github.com/microsoft/vscode-maven/tree/38c3f06f579c4422179464250b5bb5a58db75df2/test/projects/multi-module) in vscode, I see two build errors, might be related to lombok support from detailed message.

After reloading window, errors are gone. I only copied one of the two diagnostics as below:
```
[{
	"resource": "/C:/Users/admin/git/microsoft/vscode-maven/test/projects/multi-module/module1/src/main/java/com/mycompany/app/App.java",
	"owner": "_generated_diagnostic_collection_name_#2",
	"code": "0",
	"severity": 8,
	"message": "Internal compiler error: java.lang.NoClassDefFoundError: org/eclipse/jdt/core/search/SearchPattern at java.base/java.lang.Class.getDeclaredMethods0(Native Method)",
	"source": "Java",
	"startLineNumber": 1,
	"startColumn": 1,
	"endLineNumber": 1,
	"endColumn": 2
}]
```

[jdtls-log.txt](https://github.com/redhat-developer/vscode-java/files/11165456/jdtls-log.txt)

[jdtls-log-after-reloading.txt](https://github.com/redhat-developer/vscode-java/files/11165474/jdtls-log-after-reloading.txt)

vscode-java@v1.17.2023040504

snjeza:
It is an upstream lombok issue.
Steps to reproduce
- enable lombok
- open [simple maven project](https://github.com/microsoft/vscode-maven/tree/38c3f06f579c4422179464250b5bb5a58db75df2/test/projects/multi-module)
- close all editors
- clean the workspace

> After reloading window, errors are gone

I can reproduce the issue even after reloading the window if close all editors.

@Eskibear could you check [VS Code 1.17.4](https://github.com/snjeza/vscode-test/raw/master/java-1.17.4.vsix)?
Please clean the workspace after updating to 1.17.4

Eskibear:
> could you check [VS Code 1.17.4](https://github.com/snjeza/vscode-test/raw/master/java-1.17.4.vsix)?

I tried reloading a couple of times, and the issue didn't occur. 
Because I was not able to stably reproduce it, I'm not sure if it indeed fixes the issue, or just due to good luck. I suggest that you open a PR so we can dig deeper and see if that's really related.

[...]

Issue #2887, opened 2023-01-18
lombok.config ignored in 1.14.0


FlipWarthog:
After updating the extension to 1.14.0, my `lombok.config` file is being ignored, specifically the `lombok.log.fieldName=LOG` setting.

##### Environment
- Operating System: Windows
- JDK version: 14
- Visual Studio Code version: 1.74.3
- Java extension version: 1.14.0

##### Steps To Reproduce
1. Extract attached project
2. Running `mvn clean package` from the commandline succeeds
3. Loading the project into vscode with the 1.14.0 extension shows the error

Sample reproducer:
[vscode-java-issue.zip](https://github.com/redhat-developer/vscode-java/files/10441007/vscode-java-issue.zip)

Issue example:
![vscode-java-issue](https://user-images.githubusercontent.com/83613837/213059307-15f58176-bf78-4ad2-b26e-bb2db486e9cb.jpg)

##### Current Result
vscode shows that `LOG` cannot be resolved to a variable, and `log` is not used.

##### Expected Result
`lombok.config` changes `log` to `LOG` so this error/warning combo should not be shown.

##### Additional Information
I did confirm that manually reverting back to 1.13.0 makes the issue go away.

FlipWarthog:
Lombok Configuration reference: https://projectlombok.org/features/configuration

schophil:
I have got the same issue with the @CustomLog annotation.

<img width="1007" alt="Screenshot 2023-01-18 at 08 03 09" src="https://user-images.githubusercontent.com/11485058/213105662-51215a8a-2517-4de6-94cc-7e5755e34b1f.png">

familygw:
Reverting back to version 1.13 resolve the issue (workaround)

vdkhvb:
Can confirm, I got these issues as well (CustomLog annotation fails & renaming of log field is not applied). It all boils down to not processing the lombok.config file.

snjeza:
This is an upstream lombok issue related to
- https://www.eclipse.org/eclipse/news/4.27/jdt.php#ecj-separated-from-core
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/181
Lombok can't load the ResourcePlugin class at https://github.com/projectlombok/lombok/blob/master/src/core/lombok/eclipse/EclipseAST.java#L191

@FlipWarthog Could you try [Lombok 1.18.26](https://github.com/snjeza/vscode-test/raw/master/lombok-1.18.26.jar)?

```
[...]

PR #3065, opened 2023-04-13
Revert PR#3031: Revert lombok@1.18.27 edge release to 1.18.24

testforstephen:
Due to a new regression #3046 in the lombok edge version, which seems to happen more frequently than the old regression #2887, I would like to revert to the previous version 1.18.24 until the issue is resolved.


jdneo (approved):

snjeza:
@testforstephen You can use
- lombok@1.18.26 or 
- lombok@1.18.27 edge with https://github.com/eclipse/eclipse.jdt.ls/pull/2594

PR #3072, opened 2023-04-18
Add a setting to control whether 'textEdit' of completion items can be resolved

jdneo:
Use a setting to control whether to append `textEdit` to the `resolveSupport` of completion.

This allows us to have a fallback when lazy-resolving textEdit no longer works in the future, while still comply with the protocol.

Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2600

### DIFF
--- a/package.json
+++ b/package.json
@@ -617,6 +617,12 @@
           "markdownDescription": "Specify whether to match case for code completion.",
           "scope": "window"
         },
+        "java.completion.lazyResolveTextEdit.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "[Experimental] Enable/disable lazily resolving text edits for code completion.",
+          "scope": "window"
+        },
         "java.foldingRange.enabled": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -617,6 +617,12 @@
           "markdownDescription": "Specify whether to match case for code completion.",
           "scope": "window"
         },
+        "java.completion.lazyResolveTextEdit.enabled": {
```
testforstephen, line 620: Since you have created a setting to control the textEdit lazy resolve behavior, why not pass it to the server directly? Then the server can determine the supported lazy resolve properties based on both the client capabilities and the settings. It can take effect right away without reloading the window.

On the server side, you can use false as the initial default value of this new preference, so that other clients will get the completion response following the spec by default.

jdneo: Makes sense

testforstephen (approved):
LGTM. A minor comment is to add the new setting to README as well.

PR #3082, opened 2023-04-21
Ensure meaningful information is displayed via progress reporter

jdneo:
In the progress report terminal, append detailed messages only when they are available. This is to avoid seeing tasks like:

`xxx Task: status [0/0]`

### DIFF
--- a/src/serverTaskPresenter.ts
+++ b/src/serverTaskPresenter.ts
@@ -95,7 +95,15 @@ class ServerTaskTerminal implements Pseudoterminal {
 			return;
 		}
 
-		this.onDidWriteEvent.fire(`${report.id.slice(0, 8)} ${report.task}: ${report.status} [${report.workDone}/${report.totalWork}]\r\n`);
+		let taskMsg = `${report.id.slice(0, 8)} ${report.task}`;
+		if (report.status) {
+			taskMsg += `: ${report.status}`;
+		}
+		if (report.totalWork && report.workDone >= 0) {
+			taskMsg += ` [${report.workDone}/${report.totalWork}]`;
+		}
+
+		this.onDidWriteEvent.fire(`${taskMsg}\r\n`);
 	}
 
 	open(initialDimensions: TerminalDimensions): void {


src/serverTaskPresenter.ts
```diff
@@ -95,7 +95,15 @@ class ServerTaskTerminal implements Pseudoterminal {
 			return;
 		}
 
-		this.onDidWriteEvent.fire(`${report.id.slice(0, 8)} ${report.task}: ${report.status} [${report.workDone}/${report.totalWork}]\r\n`);
+		let taskMsg = `${report.id.slice(0, 8)} ${report.task}`;
+		if (report.status) {
+			taskMsg += `: ${report.status}`;
+		}
+		if (report.totalWork && report.workDone >= 0) {
+			taskMsg += ` [${report.workDone}/${report.totalWork}]`;
```
rgrunber, line 103: Are we ok with messages like this  : `Refreshing workspace 0% [0/1000]` ?

servertasks-at-0

rgrunber (commented):
I don't think I came across instances where the report status was empty, but agree that this would probably fix the problem if it did occur.

jdneo:
> I don't think I came across instances where the report status was empty, but agree that this would probably fix the problem if it did occur.

If the progress report is provided by the progress monitor, then in most case it should be fine. The problem may occur when the related APIs are called manually, for example: https://github.com/eclipse/eclipse.jdt.ls/blob/06716e21d336f8acbb5c99d29e19f252cfffe83a/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/framework/protobuf/ProtobufSupport.java#L160-L163

The above case does not have problems, but if some of the fields are missing, then the display content at the client side looks wired. I didn't find any guidance about how the APIs should be called at the server side, I think the client side should do some check to ensure the display content looks in a good way.

src/serverTaskPresenter.ts
jdneo: 0 means the job is just started, so I think it should be fine.

Issue #2128, opened 2021-09-24
Code action to surround with try-with-resources


DING-Weibing:
try with resources feature looks like absent in vscode.

##### Environment
- Operating System: Windows 10 Pro 19043.1237
- JDK version: OpenJDK 11
- Visual Studio Code version: 1.60.2
- Java extension version: Extension Pack for Java v0.18.5

##### Steps To Reproduce
1. [step 1]
![image](https://user-images.githubusercontent.com/57584830/134701563-0088137a-29ff-420f-8e2e-4b857f1ca1f0.png)

2. [step 2]
![image](https://user-images.githubusercontent.com/57584830/134701600-3af41be5-ab08-41c4-a8ab-c45eca431407.png)

##### Current Result
![image](https://user-images.githubusercontent.com/57584830/134701621-3da3e568-4063-4f13-9705-14e597fb9cbe.png)

##### Expected Result
![image](https://user-images.githubusercontent.com/57584830/134701961-e36a3f1f-f7c2-46a0-882b-ca1d4697c74b.png)

fbricon:
@rgrunber see https://www.eclipse.org/eclipse/news/4.15/jdt.php#surround-with-try-with-resources and https://bugs.eclipse.org/bugs/show_bug.cgi?id=553113 upstream

rgrunber:
I think this is possible. `SurroundWithTryWithResourcesAction` is of course a UI class (relies on Action which is tied to SWT), but it's just using `SurroundWithTryWithResourcesRefactoring`. Doing a few quick changes, most things should resolve correctly :

- Move `AbstractExceptionAnalyzer`, `ExceptionAnalyzer`, `SurroundWithTryWithResourcesAnalyzer`, `SurroundWithTryWithResourcesRefactoring` from jdt.ui to jdt.core.manipulation.
- Use `LinkedProposalModelCore` instead of `LinkedProposalModel` in `SurroundWithTryWithResourcesRefactoring` and make sure clients can continue to work with the core version
- There's some calls to `QuickAssistProcessor` static methods but they all deal with ASTNode so I suspect these are utility methods that don't need to be in UI
- `SurroundWithTryWithResourcesRefactoring` does have a single constructor meant to take an `ITextSelection`. We can't move that so we'll need to split the class into a core and create a smaller class that just extends it and has the unsupportable constructor

rgrunber:
[...]

PR #3104, opened 2023-05-13
Add support for surround with try_catch.

### LINKED ISSUE #3103: Add support for Surround with TryCatch for Runtime Exceptions
Take the following code example and try to invoke Surround with Snippet, I would expect try catch snippet which will allow me to select the Exception types provided by jdt.ls which are thrown by the foo() method according to jdt.ls completion rules.

```
public class App {

  public void foo() throws IllegalStateException {
  }

     foo();
  }
```
##### Environment
- Operating System: Any
- JDK version: Any
- Visual Studio Code version: Any
- Java extension version: Any

gayanper:
The following changeset add a new snippet which will only available under Surround With Snippets for provide a try catch block surrounding the selected text.

### DIFF
--- a/snippets/java.json
+++ b/snippets/java.json
@@ -21,7 +21,7 @@
 		"prefix": "try_catch",
 		"body": [
 			"try {",
-			"\t$1",
+			"\t${TM_SELECTED_TEXT:$1}",
 			"} catch (${2:Exception} ${3:e}) {",
 			"\t$0// TODO: handle exception",
 			"}"


gayanper:
@jdneo @testforstephen @rgrunber Let me know what you think

rgrunber:
Is there a reason to not have this on the language server side, similar to https://github.com/redhat-developer/vscode-java/issues/2128 ? Currently we have our own copy of https://github.com/eclipse/eclipse.jdt.ls/blob/f32c0f6275d1cdec241e086841c8627010c2f714/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/corext/refactoring/surround/SurroundWithTryCatchRefactoring.java (maybe we could even move it into jdt.core.manipulation from ui ?), but we only use it when there's a diagnostic that requires using try/catch. We could maybe open it up (like in JDT UI) to an arbitrary selection.

gayanper:
@rgrunber Please check the new patch, it work both with and without selection.

rgrunber (approved):
With this change, the `try_catch` snippet shows up in the `Surround With Snippet` menu. It can probably be customized for easier access with a keybinding. The snippet continues to work as before when just accessed from a completion item.

surround-with-try-catch-snippet

I was thinking of adding it to `try_resources` but that has the added complication of whether the selected code represents the resource statement, or the body. So we can leave it for now.

PR #3114, opened 2023-05-25
Turn off text edit lazy resolving

jdneo:
Turn off the lazy resolve flag due to https://github.com/microsoft/vscode/pull/183224.

@rgrunber, I suggest that we arrange a pre-release only containing this change. Then we can observe the impact on the perf using that specific version. The data can also be used for some analysis and discussion.


rgrunber (approved):
We just merged some changes with performance implications but they would be included in [`1.19.2023052504`](https://redhat.gallerycdn.vsassets.io/extensions/redhat/java/1.19.2023052504/1685002458002/Microsoft.VisualStudio.Code.Manifest) (it includes the `java.edit.validateAllOpenBuffersOnChanges` property.

We could merge this for tomorrow (or even now) and it would be separate from those.

jdneo:
Thank you @rgrunber, looks good.

rgrunber:
Appears to be an infrastructure issue. It's waiting for an appropriate node but they're all offline :|

rgrunber:
`v1.19.2023053001` should contain the this change.

jdneo:
Thank you @rgrunber !

PR #3123, opened 2023-05-30
Display error status for OutOfMemory & consistent terminations

### LINKED ISSUE #2036: Progress reporter needs to response correctly when language server crashes
We need to make the progress reporter the status bar item response correctly when language server crashes.

Possible solution are: 

- close the progress reporter
- Update the status bar icon to reflect the current status.

hopehadfield:
Fixes #2036


JessicaJHee (commented):
Change works as expected!

PR #3126, opened 2023-06-01
Revert "Turn off text edit lazy resolving"

testforstephen:
Reverts redhat-developer/vscode-java#3114

As mentioned in microsoft/vscode#183426, VS Code team has reverted the change that disallowed textEdit in resolveCompletion. We plan to revert our change as well and keep the textEdit lazy resolve behavior enabled by default in Java extension. This is because we observe disabling it causes a significant performance downgrade for code completion. We don’t want users to experience a noticeable performance degradation with 1.19.0 release. We will ask VS Code team to maintain the current behavior until we find a better alternative solution for completion optimization.

Meanwhile, for the follow-up, we will continue to look for alternative improvements, such as:
- Adopting the https://github.com/eclipse-jdt/eclipse.jdt.core/pull/939  to optimize the new Object completion case.
- Using the signature help to simplify the textEdit calculation for method completion. For example, only insert an empty parameter list `method()` in textEdit, and let the signature help to hint the parameter list later.
- Working with VS Code team to relax the restriction that the additionalTextEdit cannot overlap with the primary edit.

I suggest adding this revert in the upcoming 1.19 release, @fbricon @rgrunber @jdneo WDYT?


rgrunber:
Makes sense to me, so I'll merge. As long as we're informed of any changes to text edit resolving (so it doesn't break completely) it should be fine.

PR #3128, opened 2023-06-01
Adjust Changelog due to re-enabling lazy text edit resolving.

rgrunber:


rgrunber:
It made sense to add back the JDT-LS performance issue now that we're taking advantage of it again.

PR #3132, opened 2023-06-02
Suppress extra error messages upon language server failure

### LINKED ISSUE #3131: Unnecessary pop-ups following multiple language server failures.
Upon five language server failures, up to five error messages appear, including the following:

"The Language Support for Java server crashed 5 times in the last 3 minutes. The server will not be restarted."
"Language Support for Java client: couldn't create connection to server."
"Connection to server got closed. Server will not be restarted."
"Language Support for Java (Syntax Server) client: couldn't create connection to server."
"Restarting server failed"

Looking to suppress some of the messages as the current amount is overkill.

hopehadfield:
Fixes #3131

### DIFF
--- a/src/clientErrorHandler.ts
+++ b/src/clientErrorHandler.ts
@@ -14,13 +14,15 @@ export class ClientErrorHandler implements ErrorHandler {
 		if (count && count <= 3) {
 			logger.error(`${this.name} server encountered error: ${_message}, ${_error && _error.toString()}`);
 			return {
-				action: ErrorAction.Continue
+				action: ErrorAction.Continue,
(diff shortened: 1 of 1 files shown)


rgrunber:
This would silence all the popups resulting from a failure after the client/server are already connected. So errors in initialization would still get through. It's an improvement, and I don't think we have a way to silence the initialization errors.

We should definitely merge this as it reduces excessive popups. I'll just double check whether suppressing (returning `false`) for more errors in the initialization handler might result in an improved outcome because we don't make a futile attempt to communicate on the disposed connection.

PR #3140, opened 2023-06-07
Use choice syntax as placeholder for method/field snippets.

hopehadfield:
Gets rid of the individual `public_method`, `private_method`, `protected_method`, `public_static_method`, and `private_static_method` snippets and instead replaces them with a single `method` snippet, with choices for public/private/protected and static/non-static.

Peek 2023-06-07 14-10

Additionally, `prf` snippet is removed and replaced with `field` snippet, with options for public/private/protected.

Peek 2023-06-07 14-28

### DIFF
--- a/snippets/java.json
+++ b/snippets/java.json
@@ -38,50 +38,14 @@
 			"}"
 		]
 	},
-	"private_method": {
-		"prefix": "private_method",
+	"method": {
+		"prefix": "method",
 		"body": [
-			"private ${1:void} ${2:name}($3) {",
+			"${1|public,protected,private|}${2| , static |}${3:void} ${4:name}($5) {",
 			"\t$0",
 			"}"
 		],
-		"description": "private method"
-	},
-	"Public method": {
-		"prefix": "public_method",
-		"body": [
-			"public ${1:void} ${2:name}(${3}) {",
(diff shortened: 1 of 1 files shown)


snippets/java.json
```diff
@@ -38,50 +38,14 @@
 			"}"
 		]
 	},
-	"private_method": {
-		"prefix": "private_method",
+	"method": {
+		"prefix": "method",
 		"body": [
-			"private ${1:void} ${2:name}($3) {",
+			"${1|public,private,protected|}${2| , static |}${3:void} ${4:name}($5) {",
```
rgrunber, line 44: I would place `protected` in the middle so it's order from least restrictive to most restrictive. See https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html .

rgrunber, line 44: Why not place a space between "$2" and "$3" so you don't have to carry it in the options ? Can one of the choices for $2 not be an empty space ?

hopehadfield: I tried to leave the option in $2 completely blank (no space), but that wasn't a valid option

hopehadfield: I found a zero-width space character, I can use that instead

Issue #2903, opened 2023-01-26
Add an option not to insert arguments on completion


felixxing:
[provide a description of the issue]

##### Environment
- Operating System: Windows
- JDK version: openjdk 17.0.5 2022-10-18 LTS
- Visual Studio Code version: 1.74.3
- Java extension version: N/A

##### Steps To Reproduce
1. Create any method that has arguments 
2. Type that method and use autocomplete
3. You can see that the name of argument is also fill in, which is unexpected

##### Current Result
<img width="560" alt="Screenshot 2023-01-26 152024" src="https://user-images.githubusercontent.com/101950667/214942166-5735335f-c870-4a0e-b152-31d23f7f054e.png">

##### Expected Result
<img width="562" alt="Screenshot 2023-01-26 152048" src="https://user-images.githubusercontent.com/101950667/214942178-24c70db7-0363-4310-9a4e-a909d2fb2094.png">

##### Additional Informations
If this is a feature, can I turn of this?

snjeza:
@felixxing You can try
```
"java.completion.guessMethodArguments": false
```

felixxing:
> @felixxing You can try
> 
> ```
> "java.completion.guessMethodArguments": false
> ```

That does not work.

jdneo:
It's currently by design that the method arguments will be filled according to the selected method's signature.

Looks like we need to expose this setting from JDT
![image](https://user-images.githubusercontent.com/6193897/215376436-b0bcfa44-75d4-4a0c-b1fe-bc035556cd32.png)

I remember it's not exposed now in VS Code Java, @snjeza is it correct?

snjeza:
> I remember it's not exposed now in VS Code Java, @snjeza is it correct?

Yes, it is correct.

thgcode:
I'm not sure if the images display the same issue, but it would be nice to have the option to not include the argument names on method calls. I use a screen reader (NVDA) / keyboard and deleting all the arguments when selecting a method call everytime makes using this feature a bit annoying, especially if the method has more than 2 arguments. It would be nice if we could have the option to select between the three modes, one for completely turning it off, one for argument names and other for guessing the method arguments.

[...]

PR #3143, opened 2023-06-08
Support 'off' mode when fill the argument during completion

jdneo:
- Change the option of 'java.completion.guessMethodArguments' from boolean to four options: "auto", "off", "insertParameterNames" and "insertBestGuessedArguments". Defaults to "auto", means that for VS Code Insiders, it will use "off", other platforms will use "insertBestGuessedArguments".

requires https://github.com/eclipse/eclipse.jdt.ls/pull/2694

fix https://github.com/redhat-developer/vscode-java/issues/2903


snjeza (approved):

StraySheepPO8:
What is the minimum version of the plug-in required for this function

PR #3152, opened 2023-06-12
JDT_LS_PATH env var to reference local JDT-LS build

mickaelistria:


rgrunber:
Nice. I'll give it a try. So basically this is a way for developers who may want to test in VS Code (or verify they aren't breaking VS Code) but they don't want to set NodeJS/NPM and perform that aspect of the build. This solution allows them to hijack an existing vsix, and just swap out JDT-LS.

Definitely a "user at your own risk" but it would reduce the barrier to entry for some.

src/javaServerStarter.ts
```diff
@@ -171,7 +171,9 @@ function prepareParams(requirements: RequirementsData, javaConfiguration, worksp
 		configDir = isSyntaxServer ? 'config_ss_linux' : 'config_linux';
 	}
 	params.push('-configuration');
-	if (startedFromSources()) { // Dev Mode: keep the config.ini in the installation location
+	if (process.env.JDT_LS_PATH !== undefined) {
```
rgrunber, line 174: Would it make sense to just combine as `startedFromSources() || process.env.JDT_LS_PATH !== undefined` ? It seems like we we should just have used `serverHome` (since that will contain the correct path in both cases) and not repeated `path.resolve(__dirname, '../server'` .

You could probably customize the `console.log` message also based on whether JDT_LS_PATH exists or not.

mickaelistria: Thanks. I've refactored it a bit, hopefully into something close enough to your recommendation.

CONTRIBUTING.md
```diff
@@ -176,6 +176,16 @@ While developing the language server and the extension, you don't need to deploy
 - You can modify `launch.json` to use a different port:
     - Modify `JDTLS_CLIENT_PORT` to specify the port VS Code should connect to.
 
+## C-3) Set location to a local copy of JDT-LS
+
+This approach is mainly targeting developers of JDT-LS who want to test their changes in VSCode without having to rebuild vscode-java and to deal with the npm and vsix toolchains.
+
+- Make sure a recent snapshot of VSCode-Java is installed in your VSCode instance (post mid-June 2023)
```
rgrunber, line 183: - (as of vscode-java 1.20.0)
- VSCode-Java -> vscode-java

CONTRIBUTING.md
```diff
@@ -176,6 +176,16 @@ While developing the language server and the extension, you don't need to deploy
...
+- Clear the workspace state: `$ rm -r /home/userId/.config/Code/User/workspaceStorage/0123456789abcdef01234567890abcdef/redhat.java/jdt_ws`
+- Start VSCode with the `JDT_LS_PATH` environment variable set to the local copy of JDT-LS you want to try: `$ JDT_LS_PATH=/home/userId/git/eclipse.jdt.ls/org.eclipse.jdt.ls.product/target/products/languageServer.product/linux/gtk/x86_64 code`
+
+Most other options such as `DEBUG_VSCODE_JAVA` are still usable with this approach.4
```
rgrunber, line 187: '4' is just a typo here and can be removed I think.

CONTRIBUTING.md
```diff
@@ -176,6 +176,16 @@ While developing the language server and the extension, you don't need to deploy
 - You can modify `launch.json` to use a different port:
     - Modify `JDTLS_CLIENT_PORT` to specify the port VS Code should connect to.
 
+## C-3) Set location to a local copy of JDT-LS
+
+This approach is mainly targeting developers of JDT-LS who want to test their changes in VSCode without having to rebuild vscode-java and to deal with the npm and vsix toolchains.
+
+- Make sure a recent snapshot of VSCode-Java is installed in your VSCode instance (post mid-June 2023)
+- Clear the workspace state: `$ rm -r /home/userId/.config/Code/User/workspaceStorage/0123456789abcdef01234567890abcdef/redhat.java/jdt_ws`
```
rgrunber, line 184: I would remove this for sake of simplicity, or mention it as an optional step. We don't really mention the need to remove the `workspaceStorage` in any of the other instructions and it won't prevent using the `JDT_LS_PATH` at all. Even `globalStorage` doesn't need to be deleted for it to work.

CONTRIBUTING.md
```diff
@@ -176,6 +176,16 @@ While developing the language server and the extension, you don't need to deploy
...
+- Start VSCode with the `JDT_LS_PATH` environment variable set to the local copy of JDT-LS you want to try: `$ JDT_LS_PATH=/home/userId/git/eclipse.jdt.ls/org.eclipse.jdt.ls.product/target/products/languageServer.product/linux/gtk/x86_64 code`
```
rgrunber, line 185: I would prefer just using `$HOME/` instead of `/home/userid/`

In fact, I'm able to run it with even `$HOME/git/eclipse.jdt.ls/org.eclipse.jdt.ls.product/target/repository/` (slightly simpler path but contains the config folders & plugins)

rgrunber (commented):
Works well for me! Some good news. Everything I said about needing to clean up the configuration being stored in the `globalStorage/` folder doesn't apply. We create the location in `resolveConfiguration(context, configDir)` but your code will never take that path, so we can leave the contents completely untouched and the user can go back to their normal setup when they remove the JDT_LS_PATH environment variable!

mickaelistria:
@rgrunber all suggestions should have been addressed now.

PR #3159, opened 2023-06-19
Smart semicolon autocorrect

### LINKED ISSUE #703: Smart semicolon autocorrect?
Intellij has this nice semicolon autocorrect, that when typed within parenthesis instead of end of the line, will automatically insert the semicolon at the end of the line. 
smart-semicolon

The VS Code marketplace has at [least one](https://marketplace.visualstudio.com/items?itemName=thousandtyone.autoend) extension, which automatically moves the semicolon at the end, but I find it very buggy and causing more harm than it helps (stops inserting all semicolons!). Others require a keyboard shortcut to activate (meh).

I'm opening this issue for consideration

snjeza:
Fixes #703 
Requires https://github.com/eclipse/eclipse.jdt.ls/pull/2710

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,13 +234,13 @@ The following settings are supported:
 * `java.configuration.maven.defaultMojoExecutionAction` : Specifies default mojo execution action when no associated metadata can be detected. Defaults to `ignore`.
 * `java.completion.lazyResolveTextEdit.enabled`: [Experimental] Enable/disable lazily resolving text edits for code completion. Defaults to `true`.
 * `java.edit.validateAllOpenBuffersOnChanges`: Specifies whether to recheck all open Java files for diagnostics when editing a Java file. Defaults to `false`.
-* `java.edit.smartSemicolonDetection.enabled`: Defines the `smart semicolon` detection. Defaults to `false`.
 
 New in 1.21.0
 * `java.editor.reloadChangedSources`: Specifies whether to reload the sources of the open class files when their source jar files are changed. Defaults to `ask`.
   - `ask`: Ask to reload the sources of the open class files
   - `auto`: Automatically reload the sources of the open class files
   - `manual`: Manually reload the sources of the open class files
+* `java.edit.smartSemicolonDetection.enabled`: Defines the `smart semicolon` detection. Defaults to `false`.
 
 Semantic Highlighting
 ===============
--- a/README.md
+++ b/README.md
@@ -234,6 +234,7 @@ The following settings are supported:
 * `java.configuration.maven.defaultMojoExecutionAction` : Specifies default mojo execution action when no associated metadata can be detected. Defaults to `ignore`.
 * `java.completion.lazyResolveTextEdit.enabled`: [Experimental] Enable/disable lazily resolving text edits for code completion. Defaults to `true`.
 * `java.edit.validateAllOpenBuffersOnChanges`: Specifies whether to recheck all open Java files for diagnostics when editing a Java file. Defaults to `false`.
+* `java.edit.smartSemicolonDetection.enabled`: Defines the `smart semicolon` detection. Defaults to `false`.
 
 New in 1.21.0
 * `java.editor.reloadChangedSources`: Specifies whether to reload the sources of the open class files when their source jar files are changed. Defaults to `ask`.
--- a/package.json
+++ b/package.json
@@ -1140,6 +1140,12 @@
           "markdownDescription": "Specifies whether to recheck all open Java files for diagnostics when editing a Java file.",
           "scope": "window"
         },
+        "java.edit.smartSemicolonDetection.enabled": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Defines the `smart semicolon` detection. Defaults to `false`.",
+          "scope": "window"
+        },
         "java.editor.reloadChangedSources": {
           "type": "string",
           "enum": [
@@ -1302,6 +1308,16 @@
         "command": "java.server.restart",
         "title": "%java.server.restart%",
         "category": "Java"
+      },
+      {
+        "command": "java.action.smartSemicolonDetection.command",
+        "title": "%java.action.smartSemicolonDetection%",
+        "category": "Java"
+      },
+      {
+        "command": "java.edit.smartSemiColonDetectionUndo",
+        "title": "%java.edit.smartSemiColonDetectionUndo%",
+        "category": "Java"
       }
     ],
     "keybindings": [
@@ -1319,6 +1335,16 @@
         "key": "ctrl+shift+v",
         "mac": "cmd+shift+v",
         "when": "javaLSReady && editorLangId == java"
+      },
+      {
+        "command": "java.action.smartSemicolonDetection.command",
+        "key": ";",
+        "when": "editorFocus && javaLSReady && editorLangId == java"
+      },
+      {
+        "command": "java.edit.smartSemiColonDetectionUndo",
+        "key": "backspace",
+        "when": "editorFocus && javaLSReady && editorLangId == java"
       }
     ],
     "menus": {
--- a/package.json
+++ b/package.json
@@ -1310,13 +1310,13 @@
         "category": "Java"
       },
       {
-        "command": "java.action.smartSemicolonDetection.command",
-        "title": "%java.action.smartSemicolonDetection%",
+        "command": "java.edit.smartSemicolonDetection.command",
+        "title": "%java.edit.smartSemicolonDetection%",
         "category": "Java"
       },
       {
-        "command": "java.edit.smartSemiColonDetectionUndo",
-        "title": "%java.edit.smartSemiColonDetectionUndo%",
+        "command": "java.edit.smartSemicolonDetectionUndo",
+        "title": "%java.edit.smartSemicolonDetectionUndo%",
         "category": "Java"
       }
     ],
@@ -1337,12 +1337,12 @@
         "when": "javaLSReady && editorLangId == java"
       },
       {
-        "command": "java.action.smartSemicolonDetection.command",
+        "command": "java.edit.smartSemicolonDetection.command",
         "key": ";",
         "when": "editorFocus && javaLSReady && editorLangId == java"
       },
       {
-        "command": "java.edit.smartSemiColonDetectionUndo",
+        "command": "java.edit.smartSemicolonDetectionUndo",
         "key": "backspace",
         "when": "editorFocus && javaLSReady && editorLangId == java"
       }
--- a/package.nls.json
+++ b/package.nls.json
@@ -24,5 +24,7 @@
 	"java.action.changeBaseType": "Base on this Type",
 	"java.project.createModuleInfo.command": "Create module-info.java",
 	"java.clean.sharedIndexes": "Clean Shared Indexes",
-	"java.server.restart": "Restart Java Language Server"
+	"java.server.restart": "Restart Java Language Server",
+	"java.action.smartSemicolonDetection": "Java Smart Semicolon Detection",
+	"java.edit.smartSemiColonDetectionUndo": "Java Smart Semicolon Detection Undo"
 }
--- a/package.nls.json
+++ b/package.nls.json
@@ -25,6 +25,6 @@
 	"java.project.createModuleInfo.command": "Create module-info.java",
 	"java.clean.sharedIndexes": "Clean Shared Indexes",
 	"java.server.restart": "Restart Java Language Server",
-	"java.action.smartSemicolonDetection": "Java Smart Semicolon Detection",
-	"java.edit.smartSemiColonDetectionUndo": "Java Smart Semicolon Detection Undo"
+	"java.edit.smartSemicolonDetection": "Java Smart Semicolon Detection",
+	"java.edit.smartSemicolonDetectionUndo": "Java Smart Semicolon Detection Undo"
 }
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -338,12 +338,12 @@ export namespace Commands {
     /**
     * Smart semicolon detection.
     */
-    export const SMARTSEMICOLON_DETECTION = "java.action.smartSemicolonDetection";
-    export const SMARTSEMICOLON_DETECTION_CMD = "java.action.smartSemicolonDetection.command";
+    export const SMARTSEMICOLON_DETECTION = "java.edit.smartSemicolonDetection";
+    export const SMARTSEMICOLON_DETECTION_CMD = "java.edit.smartSemicolonDetection.command";
 
     /**
     * Smart semicolon detection backspace.
     */
-    export const SMARTSEMICOLON_DETECTION_UNDO = "java.edit.smartSemiColonDetectionUndo";
+    export const SMARTSEMICOLON_DETECTION_UNDO = "java.edit.smartSemicolonDetectionUndo";
 
 }
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -335,4 +335,15 @@ export namespace Commands {
     */
     export const GET_DECOMPILED_SOURCE = "java.decompile";
 
+    /**
+    * Smart semicolon detection.
+    */
+    export const SMARTSEMICOLON_DETECTION = "java.action.smartSemicolonDetection";
+    export const SMARTSEMICOLON_DETECTION_CMD = "java.action.smartSemicolonDetection.command";
+
+    /**
+    * Smart semicolon detection backspace.
+    */
+    export const SMARTSEMICOLON_DETECTION_UNDO = "java.edit.smartSemiColonDetectionUndo";
+
 }
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -2,10 +2,11 @@
 
 import * as fs from 'fs';
 import * as path from 'path';
-import { commands, ConfigurationTarget, env, ExtensionContext, Position, Range, SnippetString, TextDocument, Uri, window, workspace, WorkspaceConfiguration, WorkspaceFolder } from 'vscode';
+import { commands, ConfigurationTarget, env, ExtensionContext, Position, Range, Selection, SnippetString, TextDocument, Uri, window, workspace, WorkspaceConfiguration, WorkspaceFolder } from 'vscode';
 import { Commands } from './commands';
 import { cleanupLombokCache } from './lombokSupport';
 import { ensureExists, getJavaConfiguration } from './utils';
+import { apiManager } from './apiManager';
 
 const DEFAULT_HIDDEN_FILES: string[] = ['**/.classpath', '**/.project', '**/.settings', '**/.factorypath'];
 const IS_WORKSPACE_JDK_ALLOWED = "java.ls.isJdkAllowed";
@@ -27,6 +28,9 @@ export const ORGANIZE_IMPORTS_ON_PASTE = 'actionsOnPaste.organizeImports'; // ja
 
 let oldConfig: WorkspaceConfiguration = getJavaConfiguration();
 const gradleWrapperPromptDialogs = [];
+
+let oldPosition: Position = null;
+let newPosition: Position = null;
 
 export function onConfigurationChange(workspacePath: string, context: ExtensionContext) {
 	return workspace.onDidChangeConfiguration(params => {
@@ -329,6 +333,10 @@ export function handleTextBlockClosing(document: TextDocument, changes: readonly
 		return;
 	}
 	if (lastChange.text !== '"""";') {
+		if (lastChange.text !== ';') {
+			oldPosition = null;
+			newPosition = null;
+		}
 		return;
 	}
 	const selection = activeTextEditor.selection.active;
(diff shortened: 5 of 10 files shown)


rgrunber:
A few things :

I noticed that for a larger project, there is a point where one can edit a source file, while the language server is still loading. Perhaps the LS has reached `ServiceReady` but the delegate commands are not responding. I think we had to deal with a very similar case in the paste delegate command. If a user types `;` during this time, the character is completely ignored for a few seconds.

```
[Info  - 14:02:19] Jun. 19, 2023, 2:02:18 p.m. >> document/foldingRange
[Trace - 14:02:19] Sending request 'workspace/executeCommand - (29)'.
Params: {
    "command": "java.action.smartSemicolonDetection",
    "arguments": [
        "{\"uri\":\"file:///home/rgrunber/git/lemminx/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java\",\"position\":{\"line\":328,\"character\":24}}"
    ]
}
...
...
(~2s later)
...
...
[Trace - 14:02:21] Received response 'workspace/executeCommand - (29)' in 2122ms.
Result: {
    "uri": "file:///home/rgrunber/git/lemminx/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java",
    "position": {
        "line": 328,
        "character": 51
    }
![multi-line-smart-insert-semicolon](https://github.com/redhat-developer/vscode-java/assets/1417342/c12ab9bb-4a45-4ced-9e84-e77e88589e76)

}
```

Some other odd cases. Can we handle any of these ?

text-block-smart-insert-semicolon

multi-line-smart-insert-semicolon

snjeza:
> I noticed that for a larger project, there is a point where one can edit a source file, while the language server is still loading. Maybe we need only register the capability under [ServiceReady](https://github.com/redhat-developer/vscode-java/blob/71f23cc16265c24ecc8a33a0097d15ff3af8b6ee/src/standardLanguageClient.ts#L135) (like we do for paste delegate command) ? What ends up happening is the delegate commands are sent but do not produce a response. If a user types ; during this time, the character is completely ignored for a few seconds.

Fixed

snjeza:
> Some other odd cases. Can we handle any of these ?

It is related to https://github.com/eclipse/eclipse.jdt.ls/pull/2710. Checking...

hopehadfield:
Peek 2023-06-22 16-57

I was trying it out and noticed this case where the semicolon doesn't autocorrect. Is this something we can handle?

snjeza:
> I was trying it out and noticed this case where the semicolon doesn't autocorrect. Is this something we can handle?

I will try.

snjeza:
> Some other odd cases. Can we handle any of these ?

@rgrunber Could you, please, check it again.

I have also implemented the backspace char the same way as in Eclipse.
smart2

snjeza:
The remaining issues:

smart3

snjeza:
> The remaining issues:

It has been fixed.
I have copied FastJavaPartitionScanner and dependent classes - https://github.com/eclipse/eclipse.jdt.ls/pull/2710/commits/d740d724beba618afc223f267cfe9ce8db3f1ad4
@fbricon @rgrunber We could move them from o.e.jdt.ui/o.e.jface.text to o.e.jdt.core.manipulation/o.e.text.

package.json
```diff
@@ -1292,6 +1308,16 @@
         "key": "ctrl+shift+v",
         "mac": "cmd+shift+v",
         "when": "javaLSReady && editorLangId == java"
+      },
+      {
+        "command": "java.action.smartSemicolonDetection",
+        "key": ";",
+        "when": "editorFocus && javaLSReady && editorLangId == java"
```
rgrunber, line 1315: Try to insert `;` in a `.class` file. The smart semicolon detection runs because the `java` language id includes files ending with `.class`. You'll see this stacktrace.

```
[Error - 16:05:26] Jul. 18, 2023, 4:05:26 p.m. Cannot invoke "org.eclipse.jdt.core.ICompilationUnit.getBuffer()" because "unit" is null
Cannot invoke "org.eclipse.jdt.core.ICompilationUnit.getBuffer()" because "unit" is null
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.core.ICompilationUnit.getBuffer()" because "unit" is null
	at org.eclipse.jdt.ls.core.internal.handlers.SmartDetectionHandler.getLocation(SmartDetectionHandler.java:65)
	at org.eclipse.jdt.ls.core.internal.JDTDelegateCommandHandler.executeCommand(JDTDelegateCommandHandler.java:180)
	at org.eclipse.jdt.ls.core.internal.handlers.WorkspaceExecuteCommandHandler$1.run(WorkspaceExecuteCommandHandler.java:230)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jdt.ls.core.internal.handlers.WorkspaceExecuteCommandHandler.executeCommand(WorkspaceExecuteCommandHandler.java:220)
	at org.eclipse.jdt.ls.core.internal.handlers.JDTLanguageServer.lambda$4(JDTLanguageServer.java:597)
	at org.eclipse.jdt.ls.core.internal.BaseJDTLanguageServer.lambda$0(BaseJDTLanguageServer.java:87)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(Unknown Source)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool.scan(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(Unknown Source)
```

Ideally we should find something for the `when` condition that includes only `.java` files.

README.md
```diff
@@ -233,6 +233,7 @@ The following settings are supported:
 * `java.configuration.maven.defaultMojoExecutionAction` : Specifies default mojo execution action when no associated metadata can be detected. Defaults to `ignore`.
 * `java.completion.lazyResolveTextEdit.enabled`: [Experimental] Enable/disable lazily resolving text edits for code completion. Defaults to `true`.
 * `java.edit.validateAllOpenBuffersOnChanges`: Specifies whether to recheck all open Java files for diagnostics when editing a Java file. Defaults to `false`.
+* `java.edit.smartSemicolonDetection`: Defines the `smart semicolon` detection. Defaults to `false`.
```
rgrunber, line 236: I would call it `java.edit.smartSemicolonDetection.enabled`.

package.nls.json
```diff
@@ -24,5 +24,7 @@
 	"java.action.changeBaseType": "Base on this Type",
 	"java.project.createModuleInfo.command": "Create module-info.java",
 	"java.clean.sharedIndexes": "Clean Shared Indexes",
-	"java.server.restart": "Restart Java Language Server"
+	"java.server.restart": "Restart Java Language Server",
+	"java.action.smartSemicolonDetection": "Jave Smart Semicolon Detection",
+	"java.action.smartSemicolonDetectionSemichar": "Jave Smart Semicolon Detection Backspace"
```
rgrunber, line 29: - Jave -> Java

src/commands.ts
```diff
@@ -335,4 +335,14 @@ export namespace Commands {
...
 
+    /**
+    * Smart semicolon detection.
+    */
+    export const SMARTSEMICOLON_DETECTION = "java.action.smartSemicolonDetection";
+
+    /**
+    * Smart semicolon detection backspace.
+    */
+    export const SMARTSEMICOLON_DETECTION_SEMICHAR = "java.action.smartSemicolonDetectionSemichar";
```
rgrunber, line 346: - `java.edit.smartSemicolonDetection.command` ? This would also ensure it doesn't clash with the server-side delegate command.
- `java.edit.smartSemicolonDetectionUndo.command` ?

src/settings.ts
```diff
@@ -349,3 +357,69 @@ export function handleTextBlockClosing(document: TextDocument, changes: readonly
...
+				editBuilder.insert(oldPosition, ";");
+				const delRange = new Range(newPosition, new Position(newPosition.line, newPosition.character + 1));
+				editBuilder.delete(delRange);
+				window.activeTextEditor.selections = [new Selection(oldPosition, oldPosition)];
+				oldPosition = null;
+				newPosition = null;
+			});
+			return;
+		}
+		window.activeTextEditor!.edit(editBuilder => {
```
rgrunber, line 402: I think you can remove the parameter if it's unused.

rgrunber (changes requested):
Overall, looks good. Just some questions of style and where to place things.

I also noticed the following 
inconsistent-semicolon

I'm open to merging even with this as I suspect it might be some misbehaviour in the partition logic on the server-side.

src/settings.ts
```diff
@@ -349,3 +357,69 @@ export function handleTextBlockClosing(document: TextDocument, changes: readonly
 		}
 	}
 }
+
+let serverReady = false;
```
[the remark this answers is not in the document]
snjeza: > Are you using this variable here because you're trying to correctly handle the semicolon in the case of the text blocks also ?

No, it isn't related to the text block.

> I would have really like to move this into 

I have tried it, but I didn't succeed.

src/commands.ts
snjeza: There isn't the `java.action.smartSemicolonDetectionSemichar` server command.

src/settings.ts
```diff
@@ -349,3 +357,69 @@ export function handleTextBlockClosing(document: TextDocument, changes: readonly
...
 	}
 }
+
+let serverReady = false;
+
+export function registerSmartSemicolonDetection(context: ExtensionContext) {
+	apiManager.getApiInstance().serverReady().then(() => {
+		serverReady = true;
+	});
+	context.subscriptions.push(commands.registerCommand(Commands.SMARTSEMICOLON_DETECTION, async () => {
```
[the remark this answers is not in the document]
snjeza: Fixed.

package.json
snjeza: I have fixed this issue in the code. See https://github.com/redhat-developer/vscode-java/pull/3159/commits/70ed480dd30b432e6caebb9602fc8c786be0b043#diff-bb57a3bd912abc3ec2e729cb8a743838487677a5517683d0c8913a3619ac296aR368

README.md
snjeza: Fixed.

package.nls.json
snjeza: Fixed.

src/settings.ts
snjeza: Fixed.

snjeza:
@rgrunber I have updated this PR and https://github.com/eclipse/eclipse.jdt.ls/pull/2710

package.json
rgrunber: Ok. This is fine. Only slight difference is the semicolon no longer triggers the "Cannot edit in read-only editor" tooltip since we're just silently ignoring it but it's minor.

src/commands.ts
rgrunber: My mistake. Could we just call it `java.edit.smartSemiColonDetectionUndo` ? It's not obvious from the current name that this is handling the "backspace" support to undo the smart semicolon insertion.

src/settings.ts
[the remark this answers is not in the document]
rgrunber: Is there an issue with the following ? Basically I just moved all the smart semicolon logic into a new file, `smartSemicolonDetection.ts`. I made the `oldPostion`/`newPosition` variables global in there as well and create a helper function that can be used to access them from the text block logic. I also called `registerSmartSemicolonDetection(..)` directly in the `standardLanguageClient.ts` code where we listen for the `ServiceReady` notification.

<details>
<summary>patch</summary>

```diff
diff --git a/src/extension.ts b/src/extension.ts
index 90fc982..d0de894 100644
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -24,7 +24,7 @@ import { initialize as initializeRecommendation } from './recommendation';
 import * as requirements from './requirements';
 import { runtimeStatusBarProvider } from './runtimeStatusBarProvider';
 import { serverStatusBarProvider } from './serverStatusBarProvider';
-import { ACTIVE_BUILD_TOOL_STATE, cleanWorkspaceFileName, getJavaServerMode, handleTextBlockClosing, onConfigurationChange, registerSmartSemicolonDetection, ServerMode } from './settings';
+import { ACTIVE_BUILD_TOOL_STATE, cleanWorkspaceFileName, getJavaServerMode, handleTextBlockClosing, onConfigurationChange, ServerMode } from './settings';
 import { snippetCompletionProvider } from './snippetCompletionProvider';
 import { JavaClassEditorProvider } from './javaClassEditor';
 import { StandardLanguageClient } from './standardLanguageClient';
@@ -35,6 +35,7 @@ import { Telemetry } from './telemetry';
 import { getMessage } from './errorUtils';
 import { TelemetryService } from '@redhat-developer/vscode-redhat-telemetry/lib';
 import { activationProgressNotification } from "./serverTaskPresenter";
+import { registerSmartSemicolonDetection } from './smartSemicolonDetection';
 
 const syntaxClient: SyntaxLanguageClient = new SyntaxLanguageClient();
 const standardClient: StandardLanguageClient = new StandardLanguageClient();
diff --git a/src/settings.ts b/src/settings.ts
index 8cd580f..091c8ee 100644
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -7,6 +7,7 @@ import { Commands } from './commands';
 import { cleanupLombokCache } from './lombokSupport';
 import { ensureExists, getJavaConfiguration } from './utils';
 import { apiManager } from './apiManager';
+import { setSmartSemiColonDetectionState } from './smartSemicolonDetection';
 
 const DEFAULT_HIDDEN_FILES: string[] = ['**/.classpath', '**/.project', '**/.settings', '**/.factorypath'];
 const IS_WORKSPACE_JDK_ALLOWED = "java.ls.isJdkAllowed";
@@ -29,9 +30,6 @@ export const ORGANIZE_IMPORTS_ON_PASTE = 'actionsOnPaste.organizeImports'; // ja
 let oldConfig: WorkspaceConfiguration = getJavaConfiguration();
 const gradleWrapperPromptDialogs = [];
 
-let oldPosition: Position = null;
-let newPosition: Position = null;
-
 export function onConfigurationChange(workspacePath: string, context: ExtensionContext) {
 	return workspace.onDidChangeConfiguration(params => {
 		if (!params.affectsConfiguration('java')) {
@@ -334,8 +332,7 @@ export function handleTextBlockClosing(document: TextDocument, changes: readonly
 	}
 	if (lastChange.text !== '"""";') {
 		if (lastChange.text !== ';') {
-			oldPosition = null;
-			newPosition = null;
+			setSmartSemiColonDetectionState(null, null);
 		}
 		return;
 	}
@@ -358,68 +355,3 @@ export function handleTextBlockClosing(document: TextDocument, changes: readonly
 	}
 }
 
-let serverReady = false;
-
-export function registerSmartSemicolonDetection(context: ExtensionContext) {
-	apiManager.getApiInstance().serverReady().then(() => {
-		serverReady = true;
-	});
-	context.subscriptions.push(commands.registerCommand(Commands.SMARTSEMICOLON_DETECTION_CMD, async () => {
-		if (!isSemichar() && window.activeTextEditor.document.fileName.endsWith(".java")) {
-			const params: SmartDetectionParams = {
-				uri: window.activeTextEditor.document.uri.toString(),
-				position: window.activeTextEditor!.selection.active,
-			};
-			const response: SmartDetectionParams = await commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.SMARTSEMICOLON_DETECTION, JSON.stringify(params));
-			if (response !== null) {
-				window.activeTextEditor!.edit(editBuilder => {
-					oldPosition = window.activeTextEditor!.selection.active;
-					editBuilder.insert(response.position, ";");
-					window.activeTextEditor.selections = [new Selection(response.position, response.position)];
-					newPosition = window.activeTextEditor!.selection.active;
-				});
-				return;
-			}
-		}
-		window.activeTextEditor!.edit(editBuilder => {
-			editBuilder.insert(window.activeTextEditor!.selection.active, ";");
-		});
-		newPosition = null;
-		oldPosition = null;
-	}));
-	context.subscriptions.push(commands.registerCommand(Commands.SMARTSEMICOLON_DETECTION_SEMICHAR, async () => {
-		if (isSemichar()) {
-			window.activeTextEditor!.edit(editBuilder => {
-				editBuilder.insert(oldPosition, ";");
-				const delRange = new Range(newPosition, new Position(newPosition.line, newPosition.character + 1));
-				editBuilder.delete(delRange);
-				window.activeTextEditor.selections = [new Selection(oldPosition, oldPosition)];
-				oldPosition = null;
-				newPosition = null;
-			});
-			return;
-		}
-		window.activeTextEditor!.edit(() => {
-			commands.executeCommand("deleteLeft");
-		});
-		oldPosition = null;
-		newPosition = null;
-	}));
-}
-
-interface SmartDetectionParams {
-	uri: String;
-	position: Position;
-}
-
-function isSemichar() {
-	const enabled = getJavaConfiguration().get<boolean>("edit.smartSemicolonDetection");
-	const isSemichar = serverReady && window.activeTextEditor.selections.length === 1 && enabled && oldPosition !== null && newPosition !== null;
-	if (isSemichar) {
-		const active = window.activeTextEditor!.selection.active;
-		const prev = new Position(active.line, active.character === 0 ? 0 : active.character - 1);
-		return newPosition.isEqual(prev);
-	}
-	return isSemichar;
-}
-
diff --git a/src/smartSemicolonDetection.ts b/src/smartSemicolonDetection.ts
new file mode 100644
index 0000000..fceb270
--- /dev/null
+++ b/src/smartSemicolonDetection.ts
@@ -0,0 +1,73 @@
+'use strict';
+
+import { commands, ExtensionContext, Position, Range, Selection, window } from 'vscode';
+import { Commands } from './commands';
+import { getJavaConfiguration } from './utils';
+
+let oldPosition: Position = null;
+let newPosition: Position = null;
+
+export function registerSmartSemicolonDetection(context: ExtensionContext) {
+	context.subscriptions.push(commands.registerCommand(Commands.SMARTSEMICOLON_DETECTION_CMD, async () => {
+		if (!isSemichar() && window.activeTextEditor.document.fileName.endsWith(".java")) {
+			const params: SmartDetectionParams = {
+				uri: window.activeTextEditor.document.uri.toString(),
+				position: window.activeTextEditor!.selection.active,
+			};
+			const response: SmartDetectionParams = await commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.SMARTSEMICOLON_DETECTION, JSON.stringify(params));
+			if (response !== null) {
+				window.activeTextEditor!.edit(editBuilder => {
+					oldPosition = window.activeTextEditor!.selection.active;
+					editBuilder.insert(response.position, ";");
+					window.activeTextEditor.selections = [new Selection(response.position, response.position)];
+					newPosition = window.activeTextEditor!.selection.active;
+				});
+				return;
+			}
+		}
+		window.activeTextEditor!.edit(editBuilder => {
+			editBuilder.insert(window.activeTextEditor!.selection.active, ";");
+		});
+		newPosition = null;
+		oldPosition = null;
+	}));
+	context.subscriptions.push(commands.registerCommand(Commands.SMARTSEMICOLON_DETECTION_SEMICHAR, async () => {
+		if (isSemichar()) {
+			window.activeTextEditor!.edit(editBuilder => {
+				editBuilder.insert(oldPosition, ";");
+				const delRange = new Range(newPosition, new Position(newPosition.line, newPosition.character + 1));
+				editBuilder.delete(delRange);
+				window.activeTextEditor.selections = [new Selection(oldPosition, oldPosition)];
+				oldPosition = null;
+				newPosition = null;
+			});
+			return;
+		}
+		window.activeTextEditor!.edit(() => {
+			commands.executeCommand("deleteLeft");
+		});
+		oldPosition = null;
+		newPosition = null;
+	}));
+}
+
+interface SmartDetectionParams {
+	uri: String;
+	position: Position;
+}
+
+function isSemichar() {
+	const enabled = getJavaConfiguration().get<boolean>("edit.smartSemicolonDetection");
+	const isSemichar = window.activeTextEditor.selections.length === 1 && enabled && oldPosition !== null && newPosition !== null;
+	if (isSemichar) {
+		const active = window.activeTextEditor!.selection.active;
+		const prev = new Position(active.line, active.character === 0 ? 0 : active.character - 1);
+		return newPosition.isEqual(prev);
+	}
+	return isSemichar;
+}
+
+export function setSmartSemiColonDetectionState(oldPos: Position, newPos: Position) {
+	oldPosition = oldPos;
+	newPos = newPos;
+}
\ No newline at end of file
diff --git a/src/standardLanguageClient.ts b/src/standardLanguageClient.ts
index 6b07f11..2b0f6a6 100644
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -40,6 +40,7 @@ import { getAllJavaProjects, getJavaConfig, getJavaConfiguration } from "./utils
 import { Telemetry } from "./telemetry";
 import { TelemetryEvent } from "@redhat-developer/vscode-redhat-telemetry/lib";
 import { registerDocumentValidationListener } from './diagnostic';
+import { registerSmartSemicolonDetection } from './smartSemicolonDetection';
 
 const extensionName = 'Language Support for Java';
 const GRADLE_CHECKSUM = "gradle/checksum/prompt";
@@ -137,6 +138,7 @@ export class StandardLanguageClient {
 						// clients may not have properly configured documentPaste
 						logger.error(error);
 					}
+					registerSmartSemicolonDetection(context);
 					activationProgressNotification.hide();
 					if (!hasImported) {
 						showImportFinishNotification(context);
```
</details>

src/commands.ts
snjeza: Fixed.

src/settings.ts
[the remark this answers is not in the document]
snjeza: It works fine. I have updated the PR.

package.json
```diff
@@ -1302,6 +1308,16 @@
...
         "title": "%java.server.restart%",
         "category": "Java"
+      },
+      {
+        "command": "java.action.smartSemicolonDetection.command",
+        "title": "%java.action.smartSemicolonDetection%",
+        "category": "Java"
+      },
+      {
+        "command": "java.edit.smartSemiColonDetectionUndo",
```
rgrunber, line 1318: I would change `SemiColon` to `Semicolon` so it matches everywhere else. Sorry if I confused things by naming the method `setSmartSemiColonDetectionState`. That should also be `setSmartSemicolonDetectionState`.

rgrunber (approved):
Overall, this looks good. Just some very minor things to handle. Then we can merge and address any other issues.

There's another small thing I noticed but I'm open to handling it in a separate PR to avoid complicating this one further.

We set `javaLSReady`in `Started` at https://github.com/redhat-developer/vscode-java/blob/a60b5e4940727b628f45fc643f4423b46980ce2b/src/standardLanguageClient.ts#L152-L155

However, we only register the semicolon detection commands in `ServiceReady` . For large projects there can be a gap between `Started` and `ServiceReady` so there is a moment where the commands may be activated (after `Started`) but before we have registerd the command (at `ServiceReady`). The solution is probably to create a separate context state that's triggered within `ServiceReady`.

src/commands.ts
```diff
@@ -335,4 +335,15 @@ export namespace Commands {
     */
     export const GET_DECOMPILED_SOURCE = "java.decompile";
 
+    /**
+    * Smart semicolon detection.
+    */
+    export const SMARTSEMICOLON_DETECTION = "java.action.smartSemicolonDetection";
+    export const SMARTSEMICOLON_DETECTION_CMD = "java.action.smartSemicolonDetection.command";
```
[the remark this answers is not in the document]
snjeza: I have fixed both PRs

src/smartSemicolonDetection.ts
```diff
@@ -0,0 +1,73 @@
...
+import { commands, ExtensionContext, Position, Range, Selection, window } from 'vscode';
+import { Commands } from './commands';
+import { getJavaConfiguration } from './utils';
+
+let oldPosition: Position = null;
+let newPosition: Position = null;
+
+export function registerSmartSemicolonDetection(context: ExtensionContext) {
+	context.subscriptions.push(commands.registerCommand(Commands.SMARTSEMICOLON_DETECTION_CMD, async () => {
+		if (!isSemichar() && window.activeTextEditor.document.fileName.endsWith(".java")) {
```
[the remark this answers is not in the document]
snjeza: Fixed.

package.json
snjeza: Fixed.

snjeza:

rgrunber:
There's another issue. Even when `java.edit.smartSemicolonDetection.enabled` is `false`, the delegate command still runs because `!isSemichar()` is still `false`. I think we need an additional variable to check if it's disabled and skip that delegate command if block. We probably should do the same for the 

Also it looks to me like the purpose of `isSemichar()` is to detect whether we applied the smart semicolon insertion, so that we know that the next backspace should undo the smart insertion. If so, I think it would be better to call it `didSmartSemicolonInsertion`

snjeza:
> There's another issue. Even when java.edit.smartSemicolonDetection.enabled is false, the delegate command still runs because !isSemichar() is still false. I think we need an additional variable to check if it's disabled and skip that delegate command if block.

Fixed.

> Also it looks to me like the purpose of isSemichar() is to detect whether we applied the smart semicolon insertion, so that we know that the next backspace should undo the smart insertion. If so, I think it would be better to call it didSmartSemicolonInsertion

Fixed.

rgrunber (approved):
Change looks fine to me. We can address some of the issues in later contributions. The important part is it's disabled by default for now and doesn't introduce any overhead to regular semicolon/backspace usage.

PR #3162, opened 2023-06-20
Allow 3rd party extensions to track errors in language client & server

testforstephen:
Track the language client & server errors to improve the reliability of the product.


jdneo (approved):

testforstephen:
let's use a separate PR to cover more errors on client side.

PR #3165, opened 2023-06-21
Fixes part of #2685 - end-to-end performance tracking for code completion requests

robstryker:


src/standardLanguageClient.ts
```diff
@@ -735,3 +736,23 @@ export async function applyWorkspaceEdit(workspaceEdit: WorkspaceEdit, languageC
 		return Promise.resolve(true);
 	}
 }
+
+export function registerCodeCompletionTelemetryListener(context: ExtensionContext, languageClient: LanguageClient) {
+	const javaExt = extensions.getExtension("redhat.java");
+	javaExt.exports?.onDidRequestEnd((traceEvent: any) => {
```
rgrunber, line 742: If you look at the one call site of `registerCodeCompletionTelemetryListener` we actually have direct access to `apiManager` there so there's no need to go through `extensions` and then `exports`. Just `apiManager.getApiInstance().onDidRequestEnd(...)`

rgrunber (commented):
Overall, looks good. Haven't tried it yet, but just some small things I observed.

src/standardLanguageClient.ts
```diff
@@ -735,3 +736,23 @@ export async function applyWorkspaceEdit(workspaceEdit: WorkspaceEdit, languageC
...
+		if (traceEvent.type === "textDocument/completion") {
+			// See https://github.com/redhat-developer/vscode-java/pull/3010
+			// to exclude the invalid completion requests.
+			if (!traceEvent.resultLength && traceEvent.type === "textDocument/completion") {
+				return;
+			}
+			const event: TraceEvent = traceEvent as TraceEvent;
+			const props = {
+				ms: Math.round(event.duration * 100) / 100,
```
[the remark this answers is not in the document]
robstryker: toFixed returns a string. I prefer to have it in number format.

src/standardLanguageClient.ts
```diff
@@ -735,3 +736,21 @@ export async function applyWorkspaceEdit(workspaceEdit: WorkspaceEdit, languageC
 		return Promise.resolve(true);
 	}
 }
+
+export function registerCodeCompletionTelemetryListener() {
+	apiManager.getApiInstance().onDidRequestEnd((traceEvent: any) => {
```
rgrunber, line 741: Based on the API for `onDidRequestEnd` I think you could change `any` to `TraceEvent` and avoid having the line that casts to the trace event below.

rgrunber (commented):
Overall looks good.

We're tied into the tracing language client (which is used by the standard language client) so we won't get any telemetry in lightweight (syntax) mode or while a project is switching over between syntax/standard on loading, but I guess we should be ok with that for now. I also confirmed this wouldn't pick up cancellations.

Wouldn't mind hearing @fbricon's thoughts.

rgrunber:
After speaking to Fred, we decided to merge this **after** the [end of June release](https://github.com/redhat-developer/vscode-java/milestone/116). Then we can decide how best to configure it.

rgrunber:
Can you also modify the `USAGE_DATA.md` file to mention the new telemetry.

"The number of results (eg. 20), whether an error occured (eg. false), and duration (in milliseconds) when code assist is activated"

robstryker:
updated usage_data file

robstryker:
Updated unnecessary cast

rgrunber (approved):
LGTM. I'll be merging soon.

PR #3200, opened 2023-07-13
Create Java grammar and add textblock highlight support

### LINKED ISSUE #2896: Text block highlighting incorrect when it contains the substring: "/*
Syntax highlighting for text blocks breaks when it contains the substring `"/*`. It seems like the highlighter thinks that that is the start of a multi-line comment.

##### Environment
- Operating System: Ubuntu 22.04
- JDK version: 17
- Visual Studio Code version: 1.74.3
- Java extension version: v1.14.0

##### Steps To Reproduce
1. Create a Java file with a text block containing the substring with the 3 characters: `"/*`
```java
public class Blah {
        String files = """
                "/*.java"
                """;
    }
}
```

##### Current Result
Everything after the /* gets highlighted as comments (green):
image

##### Expected Result
Text inside code block should not be treated as (start of a) comment.

##### Additional Informations
Note that the highlighting behaves correctly when there is no `"` character right before the `/*`

hopehadfield:
As a result of the [original Java grammar resource](https://github.com/atom/language-java/blob/master/grammars/java.cson) for VSCode being made read-only as of December 2022, we have agreed to own the grammar in vscode-java (see https://github.com/microsoft/vscode/pull/187725). 

This is copied from https://github.com/microsoft/vscode/blob/main/extensions/java/syntaxes/java.tmLanguage.json and also includes a change that fixes #2896.

### DIFF
--- a/language-support/java/java.tmLanguage.json
+++ b/language-support/java/java.tmLanguage.json
@@ -1,0 +1,1855 @@
+{
+	"information_for_contributors": [
+		"This file has been converted from https://github.com/atom/language-java/blob/master/grammars/java.cson",
+		"If you want to provide a fix or improvement, please create a pull request against the original repository.",
+		"Once accepted there, we are happy to receive an update request."
+	],
+	"version": "https://github.com/atom/language-java/commit/29f977dc42a7e2568b39bb6fb34c4ef108eb59b3",
+	"name": "Java",
+	"scopeName": "source.java",
+	"patterns": [
+		{
+			"begin": "\\b(package)\\b\\s*",
+			"beginCaptures": {
+				"1": {
+					"name": "keyword.other.package.java"
+				}
+			},
+			"end": "\\s*(;)",
+			"endCaptures": {
+				"1": {
+					"name": "punctuation.terminator.java"
+				}
+			},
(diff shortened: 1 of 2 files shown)


rgrunber:
CC'ing @fbricon @testforstephen for awareness. As per https://github.com/microsoft/vscode/pull/187725#issuecomment-1633956154 , this seems to be the way forward for now.

rgrunber (approved):
The change looks ready to merge. Thanks for splitting this up for easy tracking of the changes made since the initial copy from the vscode repo.

The grammar is MIT licensed, so there should be no problems. As a note, we also package https://github.com/redhat-developer/vscode-java/blob/master/language-support/properties/JavaProperties.tmLanguage.json , which comes from https://github.com/textmate/java.tmbundle/tree/master#license , which is also MIT licensed.

Based on the comments in the vscode platform PR/issue, there is a chance they may create a separate repo for this grammar. We should monitor for if/when that happens so we can make future changes directly there and avoid duplication. This would also ensure that whether you have vscode-java, or not, you get proper syntax support for Java files. Alternatively, if VSCode is open to pulling the grammar from our repo, that could also work.

PR #3207, opened 2023-07-18
Reload the cached classfile sources when new source attachment is updated for them

testforstephen:
This requires jdt.ls PR https://github.com/eclipse/eclipse.jdt.ls/pull/2764.

What this PR does:
- Respond to SourceInvalidatedEvent notification. If the event is triggered by user operation such as "Attach Source", then refresh the open editors for new source directly. If it's triggered by auto downloading maven source, then ask user whether to reload the new source. This refresh behavior is controlled by a setting `java.editor.reloadChangedSources`. Defaults to `ask`, you can change it to `auto` for refreshing automatically.
- Expose the SourceInvalidatedEvent as an extension API `onDidSourceInvalidate`. This will let other source consumers (e.g. Debugger) to refresh their cache as well. The debugger should update the line numbers of the relevant stack frames when the new source jar is available.

- Ask to refresh open editors for new maven source

https://github.com/redhat-developer/vscode-java/assets/14052197/39786781-cf14-4f2a-a152-66e3a069cb18

- Auto refresh open editors for manually attached source

https://github.com/redhat-developer/vscode-java/assets/14052197/930a1512-6848-4f77-8656-2da7d774d27c

### DIFF
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -22,7 +22,7 @@ import * as pasteAction from './pasteAction';
 import { registerPasteEventHandler } from './pasteEventHandler';
 import { collectBuildFilePattern, onExtensionChange } from "./plugin";
 import { pomCodeActionMetadata, PomCodeActionProvider } from "./pom/pomCodeActionProvider";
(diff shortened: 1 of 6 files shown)


src/standardLanguageClient.ts
```diff
@@ -652,6 +669,59 @@ export class StandardLanguageClient {
...
+						workspace.getConfiguration().update("java.editor.reloadChangedSources", "auto", ConfigurationTarget.Global);
+					} else if (choice !== "Yes") {
+						return;
+					}
+				}
+				affectedDocumentUris.forEach(classFileUri => {
+					jdtContentProviderEventEmitter.fire(classFileUri);
+				});
+			} else {
+				affectedDocumentUris.forEach(classFileUri => {
```
jdneo, line 715: This code is executed both at the end of the if and else blocks, we can put it out of the if-else block

testforstephen: nice.

jdneo (approved):
LGTM

PR #3220, opened 2023-07-25
Support WorkDoneProgress in progress reporting

hopehadfield:
Allows for progress reporting through `$/progress` messages from server.

### DIFF
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -82,13 +82,15 @@ export interface StatusReport {
 }
 
 export interface ProgressReport {
-    id: string;
-    task: string;
-    subTask: string;
-    status: string;
-    workDone: number;
-    totalWork: number;
+    token: string;
+    value: any;
     complete: boolean;
+}
+
+export enum ProgressKind {
+    begin = "begin",
+    report = "report",
+    end = "end"
 }
 
 export interface ActionableMessage {
@@ -107,8 +109,9 @@ export namespace StatusNotification {
     export const type = new NotificationType<StatusReport>('language/status');
 }
 
-export namespace ProgressReportNotification {
-    export const type = new NotificationType<ProgressReport>('language/progressReport');
+// See https://github.com/microsoft/vscode-languageserver-node/blob/release/client/8.1.0/jsonrpc/src/common/connection.ts#L53-L55
+export namespace ProgressNotification {
+	export const type = new NotificationType<ProgressReport>('$/progress');
 }
 
 export namespace ClassFileContentsRequest {
--- a/src/serverStatus.ts
+++ b/src/serverStatus.ts
@@ -2,6 +2,7 @@
 
 import { EventEmitter } from "vscode";
 import { serverTasks } from "./serverTasks";
+import { ProgressKind } from "./protocol";
 
 export enum ServerStatusKind {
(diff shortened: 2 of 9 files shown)


rgrunber (commented):
There's one issue I noticed with this change.

With the legacy progress support, you would see printed tasks like : "Import Maven Project : Import project foo ... while the new change just says "Import Maven Project" . I think part of it is that the subtask is ever send from JDT-LS, which is a bug that has always existed, but I think it's worth fixing. You can also see that as the subtask parameter to ProgressReport, you're currently just providing the title.

Everything else looks pretty good for now.

hopehadfield:
@rgrunber Fixed with changes to eclipse-jdtls/eclipse.jdt.ls#2776

src/serverStatus.ts
```diff
@@ -31,7 +31,7 @@ export namespace serverStatus {
 
 	export function initialize() {
 		serverTasks.onDidUpdateServerTask(tasks => {
-			isBusy = tasks.some(task => !task.complete);
+			isBusy = tasks.some(task => !(task.value.kind === "end"));
```
rgrunber, line 34: Instead of referencing "end" like this, I would make some class `ProgressKind`, whose values can be "end", "begin", "report", "end", and use that.

src/protocol.ts
```diff
@@ -107,8 +108,8 @@ export namespace StatusNotification {
     export const type = new NotificationType<StatusReport>('language/status');
 }
 
-export namespace ProgressReportNotification {
-    export const type = new NotificationType<ProgressReport>('language/progressReport');
+export namespace ProgressNotification {
```
rgrunber, line 111: I would just add a comment, "See https://github.com/microsoft/vscode-languageserver-node/blob/release/client/8.1.0/jsonrpc/src/common/connection.ts#L53-L55" so we're aware why we had to copy it.

src/serverTasks.ts
```diff
@@ -33,8 +33,8 @@ function organizeTasks() {
 		newArray = recycleTasks(tasks, suggestedTaskEntrySize);
 	}
 
-	// make sure in-progress items are always at the end
-	newArray.sort((a, b) =>  Number(b.complete) - Number(a.complete));
+	// make sure in-ProgressReport items are always at the end
```
rgrunber, line 36: This was just meant as "in-progress" . Doesn't need to be "in-ProgressReport"

src/protocol.ts
```diff
@@ -82,13 +82,14 @@ export interface StatusReport {
...
 export interface ProgressReport {
-    id: string;
-    task: string;
-    subTask: string;
-    status: string;
-    workDone: number;
-    totalWork: number;
-    complete: boolean;
+    token: string;
+    value: any;
```
rgrunber, line 86: Only thing I can think of to improve is if it's possible to add a function to `ProgressReport` called `complete` that just performs `this.value.kind === ProgressKind.end` . That would simplify most of the checks we do for that condition. Otherwise I think this is fine.

rgrunber (approved):
Works for me.

src/protocol.ts
hopehadfield: I added a `complete` property that is defined when the notification is received.

PR #3224, opened 2023-07-26
Fix highlighting on 'new' keyword for inner classes

### LINKED ISSUE #1195: Wrong highlighting on 'new' keyword on inner classes
creating a new instance of an inner class doesn't highlight the `new` keyword properly

##### Environment
- Operating System: Win10
- JDK version: 1.8
- Visual Studio Code version: 1.41
- Java extension version: 0.54.2

##### Steps To Reproduce
1. Create a class and an inner class
```java
// Outer.java
class Outer {
    class Inner {
    }
}

// Program.java
class Program {
        Outer outer = new Outer();
        Outer.Inner inner = outer.new Inner();
    }
}
```

##### Current Result
image

##### Expected Result

image

hopehadfield:
Fixes #1195 

Before:
Screenshot from 2023-07-26 16-03-41

After:
Screenshot from 2023-07-26 15-53-26

### DIFF
--- a/language-support/java/java.tmLanguage.json
+++ b/language-support/java/java.tmLanguage.json
@@ -1377,6 +1377,17 @@
 		"properties": {
 			"patterns": [
 				{
+					"match": "(\\.)\\s*(new)",
+					"captures": {
+						"1": {
+							"name": "punctuation.separator.period.java"
+						},
+						"2": {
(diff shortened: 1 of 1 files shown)


rgrunber (approved):
It definitely looks added in the right spot (with properties) since that seems to handle qualified types.

Issue #3228, opened 2023-07-27
Backspace in find window removes match in code instead of find window


av-elier:
Type: <b>Bug</b>

* open any (?) java project, .java file
* Control+F, any matching letters, cursor still in find widget
* press backspace

Expected: one letter in find removed
Actual: match in .java file removed

Extension version: 1.21.2023072504
VS Code version: Code 1.80.1 (74f6148eb9ea00507ec113ec51c489d6ffb4b771, 2023-07-12T17:22:25.257Z)
OS version: Linux x64 5.15.0-76-generic snap
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz (16 x 4173)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: disabled_off<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>video_decode: enabled<br>video_encode: disabled_software<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: disabled_off|
|Load (avg)|5, 5, 4|
|Memory (System)|31.12GB (15.67GB free)|
|Process Argv|--no-sandbox --force-user-env --unity-launch --crash-reporter-id 0a30a66d-5d21-4646-be98-d983a3f5f775|
|Screen Reader|no|
|VM|0%|
|DESKTOP_SESSION|ubuntu|
|XDG_CURRENT_DESKTOP|Unity|
|XDG_SESSION_DESKTOP|ubuntu|
|XDG_SESSION_TYPE|x11|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383:30185418
vspor879:30202332
vspor708:30202333
vspor363:30204092
vslsvsres303:30308271
vserr242cf:30382550
pythontb:30283811
vsjup518:30340749
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
vscorecescf:30445987
vscod805cf:30301675
binariesv615:30325510
bridge0708:30335490
bridge0723:30353136
vsaa593cf:30376535
pythonvs932:30410667
py29gd2263:30792226
vsclangdc:30486549
c4g48928:30535728
[...]

Issue #3226, opened 2023-07-27
The entire class name is deleted after pressing the backspace key on the keyboard


loralei28:
##### Environment
- Operating System: M1
- JDK version:17.0.4
- Visual Studio Code version:1.80.1-universal
- Java extension version:1.21.0

##### Steps To Reproduce
1. Open a java file 
2. Double click on the class name > Rename Symbol
3. Press the backspace key on the keyboard

##### Current Result

https://github.com/redhat-developer/vscode-java/assets/108506566/368d689a-2734-4675-888d-1c30c15c71b8

##### Expected Result

![1](https://github.com/redhat-developer/vscode-java/assets/108506566/b53acb3a-fa7e-454c-b004-a9b15dacb65e)

testforstephen:
This is a regression caused by https://github.com/redhat-developer/vscode-java/pull/3159, which intercepts the system `backspace` shortcut and results in rename refactoring broken. We have to revert that PR first.

snjeza:
Fixed via https://github.com/redhat-developer/vscode-java/pull/3229

PR #3229, opened 2023-07-27
Smart semicolon autocorrect

### LINKED ISSUE #703: Smart semicolon autocorrect?

snjeza:
Fixes #703 #3226 #3228

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ The following settings are supported:
   - `ask`: Ask to reload the sources of the open class files
   - `auto`: Automatically reload the sources of the open class files
   - `manual`: Manually reload the sources of the open class files
+* `java.edit.smartSemicolonDetection.enabled`: Defines the `smart semicolon` detection. Defaults to `false`.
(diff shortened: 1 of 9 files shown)


rgrunber (approved):
Nice. Works well for me. I just wonder whether `editorFocus` is even needed. `editorTextFocus` seems to be enough to cover that case. Can it be removed ?

snjeza:
> Can it be removed ?

Yes, it can. I have removed it.

PR #3230, opened 2023-07-27
Fix highlighting for multiline class

### LINKED ISSUE #1444: Multiline inner "record" not properly highlighted
2nd part of #1442: Inner Java 14's record, when the record keyword is on a separate line, is not properly colored, even with semantic highlighting turned on.

```java
public class Demo {
    record
    Foo(int i,int j){}
    Foo foo = new Foo(0, 1);
    System.out.println(foo);
  }
}
```
is displayed as:
Screen Shot 2020-05-12 at 5 12 08 PM

hopehadfield:
Fixes [this comment](https://github.com/redhat-developer/vscode-java/issues/1444#issuecomment-627437007) from #1444 

Before:
Screenshot from 2023-07-27 13-27-00

After:
Screenshot from 2023-07-27 13-09-16

### DIFF
--- a/language-support/java/java.tmLanguage.json
+++ b/language-support/java/java.tmLanguage.json
@@ -305,7 +305,7 @@
 			]
 		},
 		"class": {
-			"begin": "(?=\\w?[\\w\\s-]*\\b(?:class|(?<!@)interface|enum)\\s+[\\w$]+)",
+			"begin": "(?=\\w?[\\w\\s-]*\\b(?:class|(?<!@)interface|enum))",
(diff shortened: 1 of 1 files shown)


rgrunber:
This seems to negatively affect text blocks : 
multi-line-class-bug

rgrunber (commented):
This seems to negatively affect text blocks : 
multi-line-class-bug

rgrunber (approved):
This looks good to me. It doesn't complicate things too much and fixes the issue.

PR #3251, opened 2023-08-22
Automatically register VMInstalls for the JDKs installed on the local machine

testforstephen:
- Use to jdk-utils lib to detect the JDKs installed on the local machine and add them to the “java.configuration.runtimes” setting. The JVMs configured in settings.json have higher priority. The server will automatically register them as VMInstalls.
- If there are multiple JDKs with the same version on the local machine, use the following order to select one as the active JVM:
  `"[JDK_HOME, JAVA_HOME, PATH] > [SDKMAN, jEnv, jabba, asdf] > Common places > Others (~/.gradle/jdks)"`
- We chose to use the nodejs library to detect the JDK because it enables the client to offer a JDK configuration UI in the future, which would give more transparency about the JDK information.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4615,9 +4615,9 @@
       "dev": true
     },
     "jdk-utils": {
-      "version": "0.4.4",
-      "resolved": "https://registry.npmjs.org/jdk-utils/-/jdk-utils-0.4.4.tgz",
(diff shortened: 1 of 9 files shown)


fbricon:
it should probably disable the JDT built-in JDK detection mechanism then

snjeza (approved):
Works for me.

testforstephen:
> it should probably disable the JDT built-in JDK detection mechanism then

make sense, fixed.

PR #3273, opened 2023-09-06
Track the case of language server OutOfMemory

testforstephen:


jdneo (approved):

rgrunber:
Your contributions are fine. I think we're just being affected by https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/2837 in https://github.com/redhat-developer/vscode-java/actions/workflows/pr-verify.yml. When I run locally, I can see the test instance hangs on the super implementation test with the same errors mentioned in that issue.

PR #3278, opened 2023-09-07
Track the LSP request data from syntax server

testforstephen:
The syntax server lives until the standard server becomes service ready. The completion performance from syntax server is also part of the end-to-end experience of completion feature. So adding a new property to indicate whether the LSP track event originates from the syntax server.


testforstephen:
@rgrunber Could you consider whether you want to measure the completion performance separately for the syntax server and the standard server?

https://github.com/redhat-developer/vscode-java/blob/4352492f24acd0da086d6bc6e7dcd13567b82a70/src/standardLanguageClient.ts#L825-L840

rgrunber:
You're right, it makes sense to add a field in `props` for `fromSyntaxServer`

src/TracingLanguageClient.ts
```diff
@@ -9,7 +9,7 @@ export const onDidRequestEnd: Event<TraceEvent> = requestEventEmitter.event;
 export class TracingLanguageClient extends LanguageClient {
 	private isStarted: boolean = false;
 
-	constructor(id: string, name: string, serverOptions: ServerOptions, clientOptions: LanguageClientOptions, forceDebug?: boolean) {
+	constructor(id: string, name: string, serverOptions: ServerOptions, clientOptions: LanguageClientOptions, readonly isSyntaxServer: boolean, forceDebug?: boolean) {
```
rgrunber, line 12: `name` also contains `Syntax Server` in the syntax server case so that could avoid needing an extra parameter, but I guess you prefer not having to create an additional helper method to evaluate the name ?

src/extension.api.ts
```diff
@@ -94,6 +94,10 @@ export interface TraceEvent {
 	 * Additional data properties, such as the completion trigger context.
 	 */
 	data?: any;
+	/**
+	 * Whether the request is from the syntax server.
```
rgrunber, line 98: I guess this should be "Whether the response is from the syntax server" ?

rgrunber (commented):
Looks good overall. I guess the final thing is just to add it to the `sendTelemetry` call for `textCompletion`

src/extension.api.ts
testforstephen: nice.

src/TracingLanguageClient.ts
testforstephen: yes, we can use the name to check if it's syntax server.

testforstephen:
https://github.com/redhat-developer/vscode-java/blob/4352492f24acd0da086d6bc6e7dcd13567b82a70/src/standardLanguageClient.ts#L156

Another issue is that you only register the completion telemetry listener when the standard server becomes service ready. This means that it will not receive any events from the syntax server, because the syntax server is already closed at this stage.

rgrunber:
> https://github.com/redhat-developer/vscode-java/blob/4352492f24acd0da086d6bc6e7dcd13567b82a70/src/standardLanguageClient.ts#L156
> 
> Another issue is that you only register the completion telemetry listener when the standard server becomes service ready. This means that it will not receive any events from the syntax server, because the syntax server is already closed at this stage.

Yup. https://github.com/redhat-developer/vscode-java/pull/3165#pullrequestreview-1495650018 . I think we can just move the registration to somewhere outside the standard/syntax code. Maybe after https://github.com/redhat-developer/vscode-java/blob/fedf3f7852beea1f2cf09e9262e5140c912e5ed3/src/extension.ts#L276 since it's just prior to initializing the syntax server ?

PR #3279, opened 2023-09-07
Remove command 'java.edit.smartSemicolonDetectionUndo'

snjeza:
Fixes #3277

PR #3281, opened 2023-09-08
Fix JavaLSReady context

snjeza:
See https://github.com/redhat-developer/vscode-java/pull/3279#issuecomment-1710977474


rgrunber:
I'll probably merge this just after playing a bit to confirm it works as expected, but it makes sense. If a client command uses the `javaLSReady` context for activation, then setting it in `ServiceReady` makes a lot more sense.

PR #3291, opened 2023-09-13
Revert "Ignore the unsupported java versions when auto detecting the installed jdks"

rgrunber:
Reverts redhat-developer/vscode-java#3258


fbricon:
you also need to revert 6cc071a

rgrunber:
I was going to revert each one manually via Web UI, but I can just add the other on here.

PR #3310, opened 2023-09-18
Open Java Server Log always opens standard server log

### LINKED ISSUE #3309: Open Java Server Log always opens standard server log
Steps to reproduce:

- Set `"java.server.launchMode": "LightWeight",`
- Start VS Code
- Open the command palette (F1)
- Select `Java: Open Java Language Server Log File`

VS Code will open <your_workspace>/redhat.java/jdt_ws/.metadata/.log

snjeza:
Fixes #3309

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -328,9 +328,9 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 			}
 
 			// Register commands here to make it available even when the language client fails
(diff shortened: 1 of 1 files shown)


rgrunber (approved):
Makes sense. I wondered what would happen if I called the command while the client is `Hybrid` and switching from `Lightweight` to `Standard`, but it seems because the server mode is `Hybrid` on initialization, it would never show the syntax server logs. I guess that's fine. It's still an improvement.

PR #3339, opened 2023-10-12
Feature discovery - Provide quick fix to 'Configure static import...'

testforstephen:
static import is one of the popular missing features that users report to us, but it is already supported by the Java extension. You can enable it by specifying a whitelist setting `java.completion.favoriteStaticMembers`. However, this feature is not easy to discover. To make it more accessible, we propose to add a quickfix option called “Configure static import...” on import statements or undefined methods. When you click on this option, it will open the settings page where you can adjust the static import preferences.


testforstephen:
https://github.com/redhat-developer/vscode-java/assets/14052197/461fd587-1936-42b9-ac2b-cc24800be163

jdneo (approved):

PR #3348, opened 2023-10-20
Exclude `**/node_modules/**` from `.java` files search

gluxon:
This is a strange PR and I would have no qualms with it being rejected.

Similar to https://github.com/microsoft/vscode-java-debug/pull/1234, I'm hoping to ignore `node_modules` to prevent large CPU usage from this extension when my teammates open a Node.js project. There's details in the linked pull request, but happy to elaborate if there's questions. Thanks!

Specifically, we're seeing this process run for long periods of time. It's searching on `/{src, test}/**/*.java`.

```
❯ ps x | grep ripgrep
93718   ??  R      0:16.30 /Applications/Visual Studio Code.app/Contents/Resources/app/node_modules.asar.unpacked/@vscode/ripgrep/bin/rg --files --hidden --case-sensitive --no-require-git -g /{src, test}/**/*.java -g !**/.git -g !**/.svn -g !**/.hg -g !**/CVS -g !**/.DS_Store -g !**/Thumbs.db --no-ignore --no-config

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -918,7 +918,10 @@ async function getTriggerFiles(): Promise<string[]> {
 			}
 		}
 
-		const javaFilesUnderRoot: Uri[] = await workspace.findFiles(new RelativePattern(rootFolder, "*.java"), undefined, 1);
+		// Paths set by 'java.import.exclusions' will be ignored when searching trigger files.
+		const exclusionGlob = getExclusionBlob();
+
+		const javaFilesUnderRoot: Uri[] = await workspace.findFiles(new RelativePattern(rootFolder, "*.java"), exclusionGlob, 1);
 		for (const javaFile of javaFilesUnderRoot) {
 			if (isPrefix(rootPath, javaFile.fsPath)) {
 				openedJavaFiles.push(javaFile.toString());
@@ -926,7 +929,7 @@ async function getTriggerFiles(): Promise<string[]> {
 			}
 		}
 
-		const javaFilesInCommonPlaces: Uri[] = await workspace.findFiles(new RelativePattern(rootFolder, "{src, test}/**/*.java"), undefined, 1);
+		const javaFilesInCommonPlaces: Uri[] = await workspace.findFiles(new RelativePattern(rootFolder, "{src, test}/**/*.java"), exclusionGlob, 1);
 		for (const javaFile of javaFilesInCommonPlaces) {
 			if (isPrefix(rootPath, javaFile.fsPath)) {
 				openedJavaFiles.push(javaFile.toString());

### CODE REVIEW

**Comment by rgrunber** [comment]
This is actually pretty reasonable. We have such logic on the language server side.

https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/317c78e09b863ef48a1d5c63fa22af7ee43d08f6/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/Preferences.java#L680-L683
https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/317c78e09b863ef48a1d5c63fa22af7ee43d08f6/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/Preferences.java#L706

I was thinking of adding `.git/` in there but apparently [the improvement is negligible](https://github.com/microsoft/vscode-java-debug/pull/1234/files#r998900345).

**Comment by jdneo** [comment]
This utility method might help (there is a type of the method name, should be `glob` instead of `blob`):

https://github.com/redhat-developer/vscode-java/blob/4cb7a84c54151491e78e1dbf54ad617abaa28e02/src/utils.ts#L101-L111

**Comment by gluxon** [comment]
@jdneo Is the recommendation to read the `java.import.exclusions` config and use it here?

If so, I'd just want to double check a quick question before making changes in this PR. I think I could see a developer wanting certain `.java` files in `java.import.exclusions` to still be a trigger that activates the language server, but not be included in imports?

Thanks for the typo catch. Fixed in latest push.

**Comment by jdneo** [comment]
> I think I could see a developer wanting certain .java files in java.import.exclusions to still be a trigger that activates the language server, but not be included in imports?

The extension will always be activated if a Java file is opened, which is defined here: https://github.com/redhat-developer/vscode-java/blob/master/package.json#L42

My understanding of this case is: The trigger files is used as an indicator of the project importer: https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/7e07649d7a9aee1872e5494b19e650058a577f55/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/managers/InvisibleProjectImporter.java#L735. If the user has already added a path into that setting, it means the user does not want to import anything from that folder. So, using that setting should be ok.

**Comment by gluxon** [comment]
Thanks for the help @jdneo! The latest push now reuses the `getExclusionsBlob()` function. I attempted to describe why this change makes sense as a comment. Would appreciate a double check that comment makes sense.

As for testing, I double checked this code path still finds a `Test.java` file at the root of the project if `build.gradle` is opened.

Screenshot 2023-10-24 at 2 26 16 PM
Screenshot 2023-10-24 at 2 26 22 PM

File: src/extension.ts [hunk 1]
```diff
@@ -918,15 +918,20 @@ async function getTriggerFiles(): Promise<string[]> {
 			}
 		}
 
-		const javaFilesUnderRoot: Uri[] = await workspace.findFiles(new RelativePattern(rootFolder, "*.java"), undefined, 1);
+		// Trigger files determine what invisible projects to import. Exclude
+		// patterns set in the java.import.exclusions config when determining
+		// trigger files.
+		const exclusionGlob = getExclusionBlob();
```
gluxon, line 924: I realized the typo I made in the original commit (caught https://github.com/redhat-developer/vscode-java/pull/3348#issuecomment-1774416986) was because I was looking at other usages of `workspace.findFiles()` and saw `getExclusionBlob()` earlier.

Happy to rename the `getExclusionBlob` function to `getExclusionGlob` in a different PR.

src/extension.ts
```diff
@@ -918,15 +918,20 @@ async function getTriggerFiles(): Promise<string[]> {
 			}
 		}
 
-		const javaFilesUnderRoot: Uri[] = await workspace.findFiles(new RelativePattern(rootFolder, "*.java"), undefined, 1);
+		// Trigger files determine what invisible projects to import. Exclude
```
jdneo, line 921: What about just simply saying: `Paths set by 'java.import.exclusions' will be ignored when searching trigger files.`

gluxon: Can do! Updated in latest push.

jdneo (approved):
LGTM!

Issue #3282, opened 2023-09-11
OutOfMemoryError leads to the connection failure "Error: Header must provide a Content-Length property"


testforstephen:
When Java language server runs out of memory, it will create a heap dump file in the specified path. This is because we add the arguments `-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=<path>` to the vmArgs list of the jdtls. The Java process will also print a message as below to the standard stdio with the location of the heap dump file.

```
java.lang.OutOfMemoryError: Java heap space
Dumping heap to ***\java_pid15080.hprof ...
Heap dump file created [2683396 bytes in 0.013 secs]
```

However, since the stdio is used for jsonrpc communication between client and language server, the unexpected message breaks the jsonrpc parsing. You will see lots of connection error `"Language Support for Java server encountered error and will shut down: undefined, Error: Header must provide a Content-Length property."` being printed to the client logs.

To reproduce this issue, you can set a small memory value (e.g. -Xmx128m) to the `java.jdt.ls.vmargs` setting.

// cc: @rgrunber

rgrunber:
I'll try this out. Is this not the same as https://github.com/redhat-developer/vscode-java/issues/2577 ? In other words `-Xlog:disable` should be preventing this. Or does the JVM treat the OOM error differently, and bypass the `Xlog` setting.

testforstephen:
Looks like the `-Xlog:disable` option does not suppress all the stdio output. You can reproduce this issue by running the following example code with vmArgs `-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=<path> -Xmx128m -Xlog:disable`.

```java
    public static void main(String[] args) {
        System.setOut(new PrintStream(new ByteArrayOutputStream()));
        System.setErr(new PrintStream(new ByteArrayOutputStream()));
        int[] array = new int[100000000];
    }
```

rgrunber:
[...]

PR #3355, opened 2023-10-23
Support for TransportKind.pipe through java.transport setting

rgrunber:
Fixes https://github.com/redhat-developer/vscode-java/issues/3282

### DIFF
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -5,7 +5,7 @@ import * as net from 'net';
 import * as os from 'os';
 import * as path from 'path';
 import { ExtensionContext, version, workspace } from 'vscode';
-import { Executable, ExecutableOptions, StreamInfo } from 'vscode-languageclient/node';
+import { Executable, ExecutableOptions, StreamInfo, TransportKind } from 'vscode-languageclient/node';
 import { logger } from './log';
 import { addLombokParam, isLombokSupportEnabled } from './lombokSupport';
 import { RequirementsData } from './requirements';
@@ -48,6 +48,21 @@ export function prepareExecutable(requirements: RequirementsData, workspacePath,
 	executable.options = options;
 	executable.command = path.resolve(`${requirements.tooling_jre}/bin/java`);
 	executable.args = prepareParams(requirements, workspacePath, context, isSyntaxServer);
+	const transportKind = getJavaConfiguration().get('transport');
+	switch (transportKind) {
+		case 'pipe':
+			executable.transport = TransportKind.pipe;
+			break;
+		case 'stdio':
+			executable.transport = TransportKind.stdio;
+			break;
+		default:
+			const isInsider: boolean = version.includes("insider");
+			const javaExtVersion = context.extension.packageJSON?.version;
+			const isPreReleaseVersion = /^\d+\.\d+\.\d{10}/.test(javaExtVersion);
+			executable.transport = (isInsider || isPreReleaseVersion) ? TransportKind.pipe : TransportKind.stdio;
+			break;
+	}
 	logger.info(`Starting Java server with: ${executable.command} ${executable.args.join(' ')}`);
 	return executable;
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1158,7 +1158,22 @@
...
+            "stdio",
+            "pipe"
+          ],
+          "enumDescriptions": [
+            "Use standard input/output",
+            "Use a named pipe"
+          ],
+          "default": "pipe",
+          "markdownDescription": "Specifies the mode of transport for communication between the client and language server.",
+          "scope": "window"
```
testforstephen, line 1174: Hint that this setting will take effect only after a restart. This is similar to other settings that require a restart.

https://github.com/redhat-developer/vscode-java/blob/9e20e131ea9f4fdf20bf361968d3cb8f700b61b8/src/settings.ts#L132-L137

src/javaServerStarter.ts
```diff
@@ -48,6 +48,17 @@ export function prepareExecutable(requirements: RequirementsData, workspacePath,
...
 	executable.args = prepareParams(requirements, workspacePath, context, isSyntaxServer);
+	const transportKind = getJavaConfiguration().get('transport');
+	switch (transportKind) {
+		case 'pipe':
+			executable.transport = TransportKind.pipe;
+			break;
+		case 'stdio':
+			executable.transport = TransportKind.stdio;
+			break;
+		default:
```
testforstephen, line 59: Fall back to `stdio` in the default branch?

rgrunber: I think I noticed that leaving `transport` empty defaulted to `stdio` (ie. we haven't defined it until now and we're clearly doing `stdio`) but I guess we can be explicit now. I'll set the default to `stdio` in the setting for now.

testforstephen: The default value of the setting has a significant impact on the user experience, as most users are unlikely to customize it manually. Our ultimate goal is to make `pipe` the default value instead of `stdio` in the setting. However, if you are concerned about the potential issues of switching to pipe right away, we can introduce a new default value `auto` for the `java.transport` setting. At the moment, we can map `auto` to pipe for the vscode insider users (similar to how we handle android support below). This will allow us to use the insider users as testers for the new pipe support. If no issues arise for insider, we can then consider rolling out to all users later.

https://github.com/redhat-developer/vscode-java/blob/9e20e131ea9f4fdf20bf361968d3cb8f700b61b8/src/utils.ts#L194-L199

rgrunber: Given that we agree to leave the setting out of `package.json`, `auto` wouldn't make sense as users would never really get to set it. I've configured the "insider" case to trigger in the default case of the switch. In other words, if `java.transport` is not set, and it is an insiders release, then we switch to `pipe`.

src/javaServerStarter.ts
```diff
@@ -48,6 +48,19 @@ export function prepareExecutable(requirements: RequirementsData, workspacePath,
...
+			const isInsider: boolean = version.includes("insider");
```
testforstephen, line 60: Another implementation is to check if the extension version is pre-release version. This can cover VS Code stable and insider better.

```typescript
const javaExtVersion = javaExt.packageJSON?.version;
const isPreReleaseVersion = /^\d+\.\d+\.\d{10}/.test(javaExtVersion);
```

rgrunber: Nice, i'll make this change!

snjeza:
@rgrunber You would need to add `java.transport` to package.json.

snjeza:
> I've removed it based on https://github.com/redhat-developer/vscode-java/pull/3355#discussion_r1371876115 . We want to avoid having it as an exposed setting for users, as the transport is an internal mechanism. We just want a way to test it in insiders/pre-release first and maybe allow more advanced users to set it.

What about -Djava.transport=pipe ? We would avoid `Unknown Configuration Setting` in settings.json.

rgrunber:
> > I've removed it based on [#3355 (comment)](https://github.com/redhat-developer/vscode-java/pull/3355#discussion_r1371876115) . We want to avoid having it as an exposed setting for users, as the transport is an internal mechanism. We just want a way to test it in insiders/pre-release first and maybe allow more advanced users to set it.
> 
> What about -Djava.transport=pipe ? We would avoid `Unknown Configuration Setting` in settings.json.

@testforstephen @fbricon , let me know your thoughts on this. I don't mind making it a system property.

fbricon:
pseudo setting is fine by me. "Unknown Configuration Setting" is expected. This is a temporary measure anyway.

PR #3374, opened 2023-11-02
Delete client log files when deleting workspace

### LINKED ISSUE #3349: is it is expected to have 19 GB of 1.1 mb log files in worksapceStorage
I was running out of diskspace and found that my /Users/manderse/Library/Application Support/Cursor/User/workspaceStorage
contained 29.6 GB.

19.4 GB of that was redhat.java folder with 18442 client.log.<timestamp> files of about 1.1 mb 

whats weird is that its all from about the same sday 2023-09-26 and 2023-09-27.

as if something "bad" happend that day and no cleanup :)

snjeza:
Fixes #3349 

I can only reproduce the problem when I stop the o.e.jdt.ls.core plugin while debugging Java LS.
The PR deletes the client.log.* files when run `Java: Clean the Java Language Server Workspace`.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -29,7 +29,7 @@ import { snippetCompletionProvider } from './snippetCompletionProvider';
 import { JavaClassEditorProvider } from './javaClassEditor';
 import { StandardLanguageClient } from './standardLanguageClient';
 import { SyntaxLanguageClient } from './syntaxLanguageClient';
-import { convertToGlob, deleteDirectory, ensureExists, getBuildFilePatterns, getExclusionBlob, getInclusionPatternsFromNegatedExclusion, getJavaConfig, getJavaConfiguration, hasBuildToolConflicts } from './utils';
+import { convertToGlob, deleteClientLog, deleteDirectory, ensureExists, getBuildFilePatterns, getExclusionBlob, getInclusionPatternsFromNegatedExclusion, getJavaConfig, getJavaConfiguration, hasBuildToolConflicts } from './utils';
 import glob = require('glob');
 import { Telemetry } from './telemetry';
 import { getMessage } from './errorUtils';
@@ -110,7 +110,12 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 	if (!storagePath) {
 		storagePath = getTempWorkspace();
 	}
+	const workspacePath = path.resolve(`${storagePath}/jdt_ws`);
 	clientLogFile = path.join(storagePath, 'client.log');
+	const cleanWorkspaceExists = fs.existsSync(path.join(workspacePath, cleanWorkspaceFileName));
+	if (cleanWorkspaceExists) {
+		deleteClientLog(storagePath);
+	}
 	initializeLogFile(clientLogFile);
 
 	const telemetryService: Promise<TelemetryService> = Telemetry.startTelemetry(context);
@@ -137,7 +142,6 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 	}).then(async (requirements) => {
 		const triggerFiles = await getTriggerFiles();
 		return new Promise<ExtensionAPI>(async (resolve) => {
-			const workspacePath = path.resolve(`${storagePath}/jdt_ws`);
 			const syntaxServerWorkspacePath = path.resolve(`${storagePath}/ss_ws`);
 
 			let serverMode = getJavaServerMode();
@@ -254,7 +258,7 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 				errorHandler: new ClientErrorHandler(extensionName),
 				initializationFailedHandler: error => {
 					logger.error(`Failed to initialize ${extensionName} due to ${error && error.toString()}`);
-					if (error.toString().includes('Connection') && error.toString().includes('disposed')) {
+					if ((error.toString().includes('Connection')  && error.toString().includes('disposed')) || error.toString().includes('Internal error')) {
 						if (!initFailureReported) {
 							apiManager.fireTraceEvent({
 								name: "java.client.error.initialization",
@@ -315,7 +319,6 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 				}
 			}));
 
-			const cleanWorkspaceExists = fs.existsSync(path.join(workspacePath, cleanWorkspaceFileName));
 			if (cleanWorkspaceExists) {
 				const data = {};
 				try {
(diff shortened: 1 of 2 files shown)


rgrunber (commented):
This works, but has a possibly unwanted side-effect.

If you activate the command to clean up the language server, it seems to delete the log file that is in use by the starting language client. Would it be possible to move the workspace cleanup prior to https://github.com/redhat-developer/vscode-java/blob/31f7f3f8c0ae0eaa5d39c3d20d080801e92b20dd/src/extension.ts#L114 ?

That way the newly created log file is unaffected by the delete.

rgrunber:
I tried the following change to your PR :

```diff
diff --git a/src/extension.ts b/src/extension.ts
index 938cf62..41a028f 100644
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -110,7 +110,12 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
        if (!storagePath) {
                storagePath = getTempWorkspace();
        }
+       const workspacePath = path.resolve(`${storagePath}/jdt_ws`);
        clientLogFile = path.join(storagePath, 'client.log');
+       const cleanWorkspaceExists = fs.existsSync(path.join(workspacePath, cleanWorkspaceFileName));
+       if (cleanWorkspaceExists) {
+               deleteClientLog(storagePath);
+       }
        initializeLogFile(clientLogFile);
 
        const telemetryService: Promise<TelemetryService> = Telemetry.startTelemetry(context);
@@ -137,7 +142,6 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
        }).then(async (requirements) => {
                const triggerFiles = await getTriggerFiles();
                return new Promise<ExtensionAPI>(async (resolve) => {
-                       const workspacePath = path.resolve(`${storagePath}/jdt_ws`);
                        const syntaxServerWorkspacePath = path.resolve(`${storagePath}/ss_ws`);
 
                        let serverMode = getJavaServerMode();
@@ -321,7 +325,6 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
                                try {
                                        cleanupLombokCache(context);
                                        deleteDirectory(workspacePath);
-                                       deleteClientLog(storagePath);
                                        deleteDirectory(syntaxServerWorkspacePath);
                                } catch (error) {
                                        data['error'] = getMessage(error);
```

Let me know if that works. The reason I kept the client log deletion separate from the workspace deletions is because those deletions also trigger the "clean workspace" telemetry event, and since telemetry initialization is done asynchronously, we can't guarantee that event would succeed if we moved that whole "deletion" block up.

If you find a nicer way, I'm open to it.

rgrunber:
On last thing. We have https://github.com/redhat-developer/vscode-java/blob/31f7f3f8c0ae0eaa5d39c3d20d080801e92b20dd/src/extension.ts#L255-L270 & https://github.com/redhat-developer/vscode-java/blob/31f7f3f8c0ae0eaa5d39c3d20d080801e92b20dd/src/clientErrorHandler.ts

One should detect failures in establishing a connection and the other can detect errors during client/server communication. If there are many errors being logged (from calling "stop" on the jdt.ls.core bundle), we have the ability to detect them and maybe prevent the client from attempting to re-initialize.

snjeza:

rgrunber:
@snjeza , were you able to reproduce the issue. The best I could do is by calling `exit()` directly in the `initialize()` of the language server. Even so, we still attempt to restart 5 times, prior to stopping, but at least there's fewer error messages.

snjeza:
I have reproduced the issue in the following way;

- install the PDE Support extension
- open eclipse.jdt.ls in VS Code
- wait to Java LS download the PDE target
- exit VS Code
- open eclipse.jdt.ls in Eclipse
- launch jdt.ls.socket-stream
- open vscode-java in VS Code
- run Launch Extension JDTLS Client

Java LS stops the org.eclipse.jdt.ls.core plugin at https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/a31867d0ee8bd1fd348a7ba52469fc35e1da9369/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/BundleUtils.java#L168
org.eclipse.osgi.container.ModuleContainer.addDependents(Module, Map<ModuleRevision, ModuleWiring>, Set<Module>) adds the o.e.jdt.ls plugin as a dependent plugin. 
org.osgi.framework.wiring.FrameworkWiring.refreshBundles(Collection<Bundle>, FrameworkListener...) tries to restart it and throws NPE at https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/a31867d0ee8bd1fd348a7ba52469fc35e1da9369/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/BundleUtils.java#L266

rgrunber:
It's not reproducing the issue for me. After running `Launch Extension JDTLS Client`, should the folder opened for the child instance of VS Code contain eclipse.jdt.ls ? It's not causing the issue at all.

snjeza:
> It's not reproducing the issue for me. After running Launch Extension JDTLS Client, should the folder opened for the child instance of VS Code contain eclipse.jdt.ls ? It's not causing the issue at all.

@rgrunber Could you try the following in Eclipse
- add the o.e.team.core plugin when launching `jdt.ls.socket-stream`?
- clear the workspace
- set a breakpoint at https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/a31867d0ee8bd1fd348a7ba52469fc35e1da9369/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/JavaLanguageServerPlugin.java#L364

rgrunber:
Thanks, that's what was needed. I had to add `o.e.team.core`.

rgrunber (approved):
Works for me. Feel free to merge when ready.

I think we should keep the original issue open to better understand how this is happening for those other users. Do you think they have some extension on their system that brings in `o.e.team.core` ?

snjeza:
`o.e.team.core` adds `o.e.jdt.ls.core` as a dependent plugin. They have it or a similar plugin. `BundleUtils.refreshBundles` tries to refresh `o.e.jdt.ls.core` and crashes.

rgrunber:
So `org.eclipse.team.core` is an optional dependency of `org.eclipse.jdt.core`. It isn't present by default in our JDT-LS runtime product. Do I have it correct that if some extension introduces `org.eclipse.team.core` through the `bundles` setting, then `jdt.ls.core` will be restarted by the Equinox runtime ?

Issue #3371, opened 2023-10-30
Not Starting Up Properly Probably Due to Unix Domain Socket Length Limitations


tthornton3-chwy:
Good evening! I and a couple other people on the team can say now that after upgrading to the latest VSCode Insiders / vscode-java standard/preview versions, the extension is not starting up successfully.
The "spinning" wheel keeps spinning indefinitely, and if you go to "Show Build Status" you get an absolutely empty window with no logs/status, at any point.
There are no interesting errors in any of the typical "Output" windows, for anything java. However, a `Java: Open all Log Files` gives at least this:

```
!SESSION 2023-10-30 16:02:54.536 -----------------------------------------------
eclipse.buildId=unknown
java.version=17.0.3
java.vendor=Eclipse Adoptium
BootLoader constants: OS=macosx, ARCH=aarch64, WS=cocoa, NL=en_US
Framework arguments:  --pipe=/var/folders/c4/f8v01p_s2sn9_hrx6y18dlxr0000gq/T/vscode-b4e4949e2fde772119c33ebd3ad375f0e8b742abf2.sock
Command-line arguments:  -data /Users/[redacted]/Library/Application Support/Code - Insiders/User/workspaceStorage/e0a072670fac3ceed984d299690f5fdc/redhat.java/jdt_ws --pipe=/var/folders/c4/f8v01p_s2sn9_hrx6y18dlxr0000gq/T/vscode-b4e4949e2fde772119c33ebd3ad375f0e8b742abf2.sock

!ENTRY org.eclipse.jdt.ls.core 1 0 2023-10-30 16:02:56.397
!MESSAGE class org.eclipse.jdt.ls.core.internal.JavaLanguageServerPlugin is started

!ENTRY org.eclipse.jdt.ls.core 4 0 2023-10-30 16:02:56.474
!MESSAGE Unix domain path too long
!STACK 0
java.net.SocketException: Unix domain path too long
	at java.base/sun.nio.ch.UnixDomainSockets.connect0(Native Method)
	at java.base/sun.nio.ch.UnixDomainSockets.connect(UnixDomainSockets.java:148)
	at java.base/sun.nio.ch.UnixDomainSockets.connect(UnixDomainSockets.java:144)
	at java.base/sun.nio.ch.SocketChannelImpl.connect(SocketChannelImpl.java:851)
[...]

PR #3375, opened 2023-11-03
Fix unix socket path for MacOS on Apple Silicon.

rgrunber:
As a pre-requisite to using the "overrides" entry (which requires npm 8.x), I've updated to Node 18. Let me know if that looks ok.

Fixes https://github.com/redhat-developer/vscode-java/issues/3371 & https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/2932
See https://github.com/microsoft/vscode-languageserver-node/issues/1346
Fix is at https://github.com/microsoft/vscode-languageserver-node/commit/d01d3edef36994bc0a2c8d9044fb1a1615608f1b

- vscode-jsonrpc now reacts to IPC path limits on platforms
- Update vscode-languageclient from 8.1.0 to 8.2.0-next.3
- Use vscode-jsonrpc 8.2.1-next.1 (vscode-languageserver-protocol 3.17.6-next.1 requires it)
- Requires engine bump from 1.67 -> 1.79

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Set Up NodeJS
         uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '18'
       - name: Set Up Java
         uses: actions/setup-java@v2
         with:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ The following will be a start to finish guide to get the entire language server 
 
 1. Install the required software:
    * latest [Visual Studio Code](https://code.visualstudio.com/)
-   * [Node.js](https://nodejs.org/) v4.0.0 or higher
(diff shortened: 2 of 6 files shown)


datho7561 (changes requested):
I think it might be a good idea to update the [CONTRIBUTING documentation](https://github.com/redhat-developer/vscode-java/blob/master/CONTRIBUTING.md#a-client-side-setup) (I don't know if this is the right doc) to specify that Node 18 is required, and specify what version of `npm` is required (I think the npm version is any 9.y.z). The format of `package-lock.json` changed significantly twice between npm 6 and npm 9, so if you `npm i` using the wrong version of `npm` it will fail or rewrite the file.

Otherwise, the code looks good! Unfortunately don't have access to Apple Silicon to test.

rgrunber:
Good catch. Yeah we definitely need to update the contributing document to indicate the Node version required.

testforstephen (approved):
LGTM. We tested it on a macOS M2 machine, the domain socket can be setup with this PR.

PR #3377, opened 2023-11-07
Add textmate definition for kotlin dsl

### LINKED ISSUE #3334: Host textmate definition for kotlin language
Buildship recently added experimental feature -- a language server for the *.kts files, which works for the gradle kts scripts.

It introduces a textmate definition for kotlin language: https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.kotlindsl.provider/kotlin.tmLanguage.json

Should we consider hosting this file in the extension? Currently there is no hightlighting when user opens *.kts files.

jdneo:
fix #3334

Before:
before

After:
after

### DIFF
--- a/language-support/kotlin/kotlin.tmLanguage.json
+++ b/language-support/kotlin/kotlin.tmLanguage.json
@@ -1,0 +1,571 @@
+{
+    "information_for_contributors": [
+		"This file has been copied from https://github.com/eclipse/buildship/blob/b848d9a08283b68860671a73ceec3c99bdab27c2/org.eclipse.buildship.kotlindsl.provider/kotlin.tmLanguage.json"
+	],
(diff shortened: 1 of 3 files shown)


testforstephen (approved):

fwcd:
Hm, while it's nice to have `*.gradle.kts` support in the upstream Java extension, it potentially also overrides the grammar of other Kotlin extensions (disclaimer: I'm the maintainer of [this Kotlin extension](https://marketplace.visualstudio.com/items?itemName=fwcd.kotlin)). This is a bit frustrating since [bugs in the grammar](https://github.com/redhat-developer/vscode-java/pull/3407#discussion_r1451831954) can potentially no longer be addressed by the Kotlin extensions themselves.

Perhaps, as an intermediate fix, the language support for Kotlin via this extension could be scoped to `.gradle.kts` files, so it doesn't affect other Kotlin files (which aren't handled by this language server anyway)?

PR #3397, opened 2023-11-22
DRAFT: Add SQL block highlighting support

### LINKED ISSUE #831: Support syntax highlight for embedded SQL languages
When embedding SQL into the Java code, it will be helpful to show the syntax highlight for the SQL.

VS Code provides a guide about how to support embedded languages. https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#embedded-languages

There is a VS Code extension already providing SQL support.
https://github.com/mydesireiscoma/es6-string-html/issues/10
https://marketplace.visualstudio.com/items?itemName=hjb2012.vscode-es6-string-html

aceArt-GmbH:
Fixes #831

Before:
image

After:
image

I am not familiar with vscode extensions, but here a PoC


rgrunber:
Nice. I definitely like that this part of the grammar can be injected into the necessary scope.

aceArt-GmbH:
Please review the regex

@fbricon is this solution in scope of this extension?

aceArt-GmbH:
Thanks @fbricon for testing. I updated the code accordingly.

Having other words in a comment breaking highlighting is somewhat intentional. We don't want to highlight any string, just because a comment mentions "sql".
But I think now with `language=sql` it probably won't happen on accident.

image

package.json
```diff
@@ -211,6 +211,16 @@
...
+      },
+      {
+        "injectTo": [
+          "source.java"
+        ],
+        "scopeName": "inline.sql",
+        "path": "./language-support/sql/inline-sql.json",
+        "embeddedLanguages": {
+          "meta.embedded.block.sql": "sql"
+        }
```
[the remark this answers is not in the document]
aceArt-GmbH: You are right. Sad.
Maybe https://github.com/microsoft/vscode/issues/133397 is the problem here?

Folding doesn't work either (but this is [intended?](https://github.com/microsoft/vscode/issues/176034#issuecomment-1505188746))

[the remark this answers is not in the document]
rgrunber: Ok, then now that I think of it, maybe it's best to just keep the settings that enable it there (as you have it), and if it's ever fixed, we'll have it start working for free. We've documented the limitations here for reference.

language-support/sql/inline-sql.json
```diff
@@ -0,0 +1,47 @@
+{
+	"fileTypes": ["java"],
+	"injectionSelector": "L:source.java -comment -string ",
```
rgrunber, line 3: My understanding of `L:source.java -comment -string` is it's supposed to inject into the `source.java` scope but exclude `comment.*` & `string.*` scopes. I'm not sure why we would avoid injecting into the `string.*` scopes because that's definitely what we want. We want to highlight text blocks that match `language-sql` pattern.

As for `comment.*` scopes, from the below example, that seems like  a comment block, but the sql syntax still is activated.

I guess my only point is if it's safe to make this `"injectionSelector": "L:source.java",` without changing any existing functionality, I would do that.

sql-inside-comments

rgrunber (approved):
I think this is ready to merge. I just had some questions about whether certain parts can be removed (ie. redundant/not working) to make this easier to maintain.

language-support/sql/inline-sql.json
aceArt-GmbH: There seems to be no [real multi line support](https://github.com/Microsoft/vscode-textmate/issues/41#issuecomment-344294298) available.
In the current state of this PR, there are 2 separate regex selectors.
This means, adding a `/* language=sql */` anywhere in the file, will destroy color highlighting until the second regex matches a `"`.
This is really bad and also explains the comment situation in your example.
"es6-string-html" has currently the same fate :/
https://macromates.com/manual/en/regular_expressions mentions `?m` should add multi line support, but I think VSCode does not implement it.

rgrunber: Alright, I guess we can live with that. Can you remove ` -comment -string ` as well as `fileTypes` without it breaking anything ? If so, feel free to do so, and I think we can merge this. If there's something we lose because of it, we can keep it. I just want to ensure everything added is needed (for the sake of easier maintenance).

aceArt-GmbH: With `-comment -string` the following two examples look like strings/comments.
Without it, the variable would look like a string/comment, but the value would have color highlighting, which I find odd

```java
String x = """
  String ONE_LINE_SQL = // language=sql
  "SELECT * FROM TABLE";
""";

/*
String ONE_LINE_SQL = // language=sql
"SELECT * FROM TABLE";
*/
  ```

rgrunber (approved):
There's definitely some oddities, but given that everything is guarded off by "language=sql" and not leaking into regular usage, we can go ahead with this and propose additional fixes as needed.

language-support/sql/inline-sql.json
rgrunber: Ok, so then it does help under such cases. The example I had a few comments above, was just one edge case not handled by the exclusions.

PR #3398, opened 2023-11-22
Support add/remove imported projects

jdneo:
- Reuse the 'BuildFilesSelector' and enable change the imported projects after the activation finishes.

requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2972

### DIFF
--- a/src/apiManager.ts
+++ b/src/apiManager.ts
@@ -18,6 +18,7 @@ class ApiManager {
     private onDidClasspathUpdateEmitter: Emitter<Uri> = new Emitter<Uri>();
     private onDidServerModeChangeEmitter: Emitter<ServerMode> = new Emitter<ServerMode>();
     private onDidProjectsImportEmitter: Emitter<Uri[]> = new Emitter<Uri[]>();
+    private onDidProjectsDeleteEmitter: Emitter<Uri[]> = new Emitter<Uri[]>();
     private traceEventEmitter: Emitter<any> = new Emitter<any>();
     private sourceInvalidatedEventEmitter: Emitter<SourceInvalidatedEvent> = new Emitter<SourceInvalidatedEvent>();
     private serverReadyPromiseResolve: (result: boolean) => void;
@@ -46,6 +47,7 @@ class ApiManager {
         const onDidClasspathUpdate = this.onDidClasspathUpdateEmitter.event;
         const onDidServerModeChange = this.onDidServerModeChangeEmitter.event;
         const onDidProjectsImport = this.onDidProjectsImportEmitter.event;
+        const onDidProjectsDelete = this.onDidProjectsDeleteEmitter.event;
         const traceEvent = this.traceEventEmitter.event;
 
         const serverReadyPromise: Promise<boolean> = new Promise<boolean>((resolve) => {
@@ -69,6 +71,7 @@ class ApiManager {
             serverMode,
             onDidServerModeChange,
             onDidProjectsImport,
+            onDidProjectsDelete,
             serverReady,
             onWillRequestStart,
             onDidRequestEnd,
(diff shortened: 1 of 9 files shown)


jdneo:
Some follow-ups per the online meeting:

1. Benchmark the perf of the file searching when it comes to a big project containing many many modules.
2. Test the behavior with Eclipse projects.

I'll share the result after finishing those tests.

jdneo:
> Benchmark the perf of the file searching when it comes to a big project containing many many modules.

I cloned this repo: `https://github.com/Azure/azure-sdk-for-java`, which has 690 `pom.xml` files. The time takes to find all pom files on my machine is `224 ms`, it's a windows with Intel i7-10700.

> Test the behavior with Eclipse projects.

Ok, it turns out the answer is simple. Picking Eclipse project is not supported now, because currently we only registered Maven and Gradle. (See: https://github.com/redhat-developer/vscode-java/blob/86bf3ae02f4f457184e6cc217f20240f9882dde9/package.json#L75-L84)

It should be fine to add Eclipse(`.project`) if we find it's necessary or asked by users.

testforstephen:
I have a play on the user experience, here are the problems I have:
1. The project picker should mark the imported projects differently, otherwise you will get lost when you click the picker item a few times.
2. Java projects are not refreshed automatically when new projects are imported.
3. When I try to remove all imported projects, nothing happens.
4. Check the parent pom.xml, it will actually import all sub modules on server side. When you go back the client project picker, the imported project list is not consistent with the server.
image

jdneo:
@testforstephen  Updated.

Now there is one case that cannot be handled.

If a parent pom has been imported, and some submodules are not. Selecting the submodule will not import them. Looks like its due to the implementation of maven importer: https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/9de85d52b9a8b3d51ec60112646e043ff222533d/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/managers/MavenProjectImporter.java#L177

Here it will find the project for the parent pom, thus, the submodule's pom will not be added to the `toImport` set.

------

Meanwhile, a new event is added for project deletion. https://github.com/microsoft/vscode-java-dependency/pull/802 consumes that to make sure the explorer gets refreshed on project deletion.

testforstephen:
The workflow works better now. Just some issues with the state management.
1. Change the setting to manual import mode, and then open a multimodule workspace (e.g. https://github.com/spring-guides/gs-multi-module/tree/main/initial). Follow the action to select one project `application` to import. Then use the command "**Java: Import Java Projects into Workspace**" to remove all imported projects. Check the Projects view and verified that no projects are present. Reload the window and you will see the deleted projects are present in Projects view. It looks like the deleted projects are imported automatically after reload.

2. Open a multimodule workspace (e.g. https://github.com/spring-guides/gs-multi-module/tree/main/initial) and import all projects first. Then run the command "**Java: Import Java Projects into Workspace**" to remove all imports project. Here is what it displays in Java Projects view. Do we change the hint to "**Import Projects...**"?
image

3. Open a multimodule workspace and import all projects first. Then run the command "**Java: Import Java Projects into Workspace**" to remove all imported projects. After that, run the command again and select one project to import. When the import is finished and go to Java Projects view, you will find that the **Refresh** context menu is lost.
image

jdneo:
> 1

Should be fixed in the latest commit

> 2

I expect to thoroughly improve it in project manager extension later. Current state management looks a little bit fragmented which is not easy to tune.

> 3

Should be fixed in the appended commit in https://github.com/microsoft/vscode-java-dependency/pull/802

testforstephen (approved):
Works for me now.

Issue #3323, opened 2023-09-28
Allow pasting a Java file into the file explorer


Artur-:
A handy way to get a java file from an external source into the project is to copy the source (not the file) and then paste it in the file tree. This has "always" worked in Eclipse and it would be great to have it working also in VS Code.

When pasting into a Java source folder, the file should be placed in the correct package (not the folder it is pasted in) and named according to a public Java class in the snippet. 

So pasting e.g. this
```
package foo.bar.baz;

public class HelloWorld {
}
```

into `src/main/java` in a Maven project should create `src/main/java/foo/bar/baz/HelloWorld.java`

##### Environment
- Operating System: mac
- JDK version: 21
- Visual Studio Code version: 1.82.2
- Java extension version: 1.22.1

##### Steps To Reproduce
1. Copy
2. Paste

##### Current Result
Nothing happens when you paste a Java snippet into the explorer

##### Expected Result
As described above

rgrunber:
Agreed. I really like that feature. Especiall that it creates the folder structure defined by the `package ...` declaration of the code sample (if present). I don't think we can use VS Code's experimental document paste API. That requires a document to be opened. Maybe we could create some key binding around <kbd>ctrl+v</kbd> or <kbd>ctrl+shift+v</kbd> and accessing https://code.visualstudio.com/api/references/vscode-api#Clipboard . Once we have the file, the invisible project support should be enough to handle the rest.

hopehadfield:
This feature is currently in the works. When a java text is pasted in the file explorer using `ctrl + shift + v`, a file is created containing that text and named using the declared type name. It is also added to the appropriate package based on the package name at the top.

![Peek 2023-11-10 14-22](https://github.com/redhat-developer/vscode-java/assets/115827695/fc1b73cc-d103-4175-9b31-0ac47e16255c)

This feature extends to when the language server is not already running, though it works with some delay for the server to get started.

![Peek 2023-11-10 14-26](https://github.com/redhat-developer/vscode-java/assets/115827695/b6c52a48-57f2-46fb-b0bd-c25560ce61ef)

When there is no project currently open, the file is placed in a tmp folder in the user's home directory. The following gif also exemplifies the creation of a folder according to the package name if it does not already exist.

![Peek 2023-11-10 14-24](https://github.com/redhat-developer/vscode-java/assets/115827695/bd864ee2-176e-4a8c-b0c0-e0af2515d938)

[...]

PR #3404, opened 2023-11-29
Support for pasting a java file in the file explorer

hopehadfield:
Addresses #3323

Introduces new keybinding and command for pasting a java file into the file explorer. Relies on delegate command introduced in https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2981.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1487,6 +1487,11 @@
         "command": "java.server.restart",
         "title": "%java.server.restart%",
         "category": "Java"
+      },
+      {
+        "command": "java.action.filesExplorerPasteAction",
+        "title": "%java.action.filesExplorerPasteAction%",
+        "category": "Java"
       }
     ],
     "keybindings": [
@@ -1504,6 +1509,12 @@
         "key": "ctrl+shift+v",
         "mac": "cmd+shift+v",
         "when": "javaLSReady && editorLangId == java"
+      },
+      {
+        "command": "java.action.filesExplorerPasteAction",
+        "key": "ctrl+shift+v",
+        "mac": "cmd+shift+v",
+        "when": "explorerViewletFocus && config.editor.pasteAs.enabled"
       }
     ],
     "menus": {
@@ -1622,6 +1633,10 @@
         {
           "command": "java.server.restart",
           "when": "javaLSReady"
+        },
+        {
+          "command": "java.action.filesExplorerPasteAction",
+          "when": "false"
         }
       ],
       "view/title": [
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -37,6 +37,7 @@ import { TelemetryService } from '@redhat-developer/vscode-redhat-telemetry/lib'
 import { activationProgressNotification } from "./serverTaskPresenter";
 import { loadSupportedJreNames } from './jdkUtils';
 import { BuildFileSelector, PICKED_BUILD_FILES, cleanupProjectPickerCache } from './buildFilesSelector';
+import { pasteFile } from './pasteAction';
 
 const syntaxClient: SyntaxLanguageClient = new SyntaxLanguageClient();
 const standardClient: StandardLanguageClient = new StandardLanguageClient();
@@ -93,6 +94,14 @@ function getHeapDumpFolderFromSettings(): string {
 
 export async function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 	await loadSupportedJreNames(context);
+	context.subscriptions.push(commands.registerCommand(Commands.FILESEXPLORER_ONPASTE, async () => {
+		const originalClipboard = await env.clipboard.readText();
+		// Hack in order to get path to selected folder if applicable (see https://github.com/microsoft/vscode/issues/3553#issuecomment-1098562676)
+		await commands.executeCommand('copyFilePath');
+		const folder = await env.clipboard.readText();
+		await env.clipboard.writeText(originalClipboard);
+		pasteFile(folder);
+	}));
 	context.subscriptions.push(markdownPreviewProvider);
 	context.subscriptions.push(commands.registerCommand(Commands.TEMPLATE_VARIABLES, async () => {
 		markdownPreviewProvider.show(context.asAbsolutePath(path.join('document', `${Commands.TEMPLATE_VARIABLES}.md`)), 'Predefined Variables', "", context);
@@ -650,7 +659,7 @@ function enableJavadocSymbols() {
 	});
 }
 
-function getTempWorkspace() {
+export function getTempWorkspace() {
 	return path.resolve(os.tmpdir(), `vscodesws_${makeRandomHexString(5)}`);
 }
 
--- a/src/pasteAction.ts
+++ b/src/pasteAction.ts
@@ -1,9 +1,11 @@
 'use strict';
 
-import { commands, env, ExtensionContext, Range, TextEditor, window } from 'vscode';
+import { TextEncoder } from 'util';
+import { commands, env, ExtensionContext, Range, TextEditor, Uri, window, workspace } from 'vscode';
 import { LanguageClient } from 'vscode-languageclient/node';
-
+import { apiManager } from './apiManager';
 import { Commands } from './commands';
+import fs = require('fs');
 
 export function registerCommands(languageClient: LanguageClient, context: ExtensionContext) {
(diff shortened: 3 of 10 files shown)


src/pasteAction.ts
```diff
@@ -48,4 +53,34 @@ export async function registerOrganizeImportsOnPasteCommand(): Promise<void> {
...
+		// If given path to selected folder is invalid (no folder is selected)
+		} else if (filePath === clipboardText || stats.isFile()) {
+			filePath = workspace.workspaceFolders[0].uri.path;
+		}
+		if (!serverReady) {
+			await apiManager.getApiInstance().serverReady().then( async () => {
+				serverReady = true;
+			});
+		}
+		const fileUri = Uri.file(await commands.executeCommand(Commands.RESOLVE_PASTED_TEXT, filePath, clipboardText));
```
rgrunber, line 80: I would prefer if you did

```typescript
await commands.executeCommand(Commands.EXECUTE_WORKSPACE_COMMAND, Commands.RESOLVE_PASTED_TEXT, filePath, clipboardText)
```

Although any delegate command defined in JDT-LS can be called on the client side directly, the mechanism is a bit obscured, and I think by convention we pretty much call it through `Commands.EXECUTE_WORKSPACE_COMMAND` everywhere.

package.json
```diff
@@ -1504,6 +1514,12 @@
         "key": "ctrl+shift+v",
         "mac": "cmd+shift+v",
         "when": "javaLSReady && editorLangId == java"
+      },
+      {
+        "command": "java.action.filesExplorerPasteAction",
+        "key": "ctrl+shift+v",
+        "mac": "cmd+shift+v",
+        "when": "(explorerViewletFocus || (workspaceFolderCount == 0 && explorerViewletVisible)) && config.java.pasteActions.fileExplorer"
```
rgrunber, line 1522: I think we should replace `java.pasteActions.fileExplorer` with `editor.pasteAs.enabled` (seems to be enabled by default), which is part of VS Code.

The option controls all the paste event handling actions. The only reason I could think of having the extra option is in case this breaks really badly and we want to give users a way to disable it without also disabling the paste event handling, but I just don't see it being worth it to carry the extra option.

src/extension.ts
```diff
@@ -93,6 +94,14 @@ function getHeapDumpFolderFromSettings(): string {
 
 export async function activate(context: ExtensionContext): Promise<ExtensionAPI> {
 	await loadSupportedJreNames(context);
+	context.subscriptions.push(commands.registerCommand(Commands.FILESEXPLORER_ONPASTE, async () => {
+		const originalClipboard = await env.clipboard.readText();
+		// Hack in order to get path to selected folder (if applicable)
```
rgrunber, line 99: I would link to the issue / PR that explains the hack https://github.com/microsoft/vscode/issues/3553 . If it ever stops working at least we would know where request more official support. I don't mind if we lose this capability as this feature is still valuable without it, but it's nice to have.

src/pasteAction.ts
```diff
@@ -48,4 +53,34 @@ export async function registerOrganizeImportsOnPasteCommand(): Promise<void> {
 			}
 		}
 	});
+}
+
+let serverReady = false;
+
+export async function pasteFile(context: ExtensionContext, folder: fs.PathLike): Promise<void>  {
```
rgrunber, line 60: `context` doesn't seem to be used inside here.

src/pasteAction.ts
```diff
@@ -48,4 +53,34 @@ export async function registerOrganizeImportsOnPasteCommand(): Promise<void> {
...
+	const clipboardText: string = await env.clipboard.readText();
+	let filePath = folder.toString();
+	fs.stat(folder, async (err, stats) => {
+		// If no project is open in the workspace
+		if (filePath === "." || workspace.workspaceFolders === undefined) {
+			filePath = os.homedir();
+			filePath += "/tmp";
```
rgrunber, line 67: The path separator character is different on Windows, so this is going to cause issues. Even if `fs` can translate, you're better off using a function that does it in an OS-agnostic way. Look around the codebase for `os.homedir()` and see if you can find a better way.

However, I'd rather just use https://github.com/redhat-developer/vscode-java/blob/aea6d6f500b0e26afd8e6357a0197ec119c2d1a9/src/extension.ts#L655-L657

package.json
```diff
@@ -1487,6 +1492,11 @@
         "command": "java.server.restart",
         "title": "%java.server.restart%",
         "category": "Java"
+      },
+      {
+        "command": "java.action.filesExplorerPasteAction",
+        "title": "Paste a file",
```
rgrunber, line 1498: - This command shows up in the command palette, even though it's not really useful that way. There's a way to hide the command from the palette and we do similar with some other commands.
- For the sake of accuracy I'd call it "Paste clipboard text into a file", or something like that.

rgrunber (changes requested):
Looks good overall, but I have yet to test, and will look more closely at `pasteFile`.

package.json
```diff
@@ -1487,6 +1487,11 @@
         "command": "java.server.restart",
         "title": "%java.server.restart%",
         "category": "Java"
+      },
+      {
+        "command": "java.action.filesExplorerPasteAction",
+        "title": "Paste clipboard text into a file",
```
rgrunber, line 1493: Have a look at https://github.com/redhat-developer/vscode-java/blob/master/package.nls.json to understand how the internationalization works (see above examples). After that just translate them to the corresponding languages.

rgrunber (approved):
Overall, change looks good on this end is ready to merge. I just need to verify the language server part.

src/pasteAction.ts
```diff
@@ -48,4 +54,34 @@ export async function registerOrganizeImportsOnPasteCommand(): Promise<void> {
...
+let serverReady = false;
+
+export async function pasteFile(folder: fs.PathLike): Promise<void>  {
+	const clipboardText: string = await env.clipboard.readText();
+	let filePath = folder.toString();
+	fs.stat(folder, async (err, stats) => {
+		// If no project is open in the workspace
+		if (filePath === "." || workspace.workspaceFolders === undefined) {
+			filePath = getTempWorkspace();
+			if(!fs.existsSync(filePath)){
```
rgrunber, line 68: You may want to confirm you have permission to write out the file, or write within the directory. If I set the directory to read-only where I might expect a paste I get :

```
Exception has occurred: NoPermissions (FileSystemError): Error: EACCES: permission denied, open '/home/rgrunber/sample-projects/java-textblocks/org/example/TextBlocks1.java'
t y.e (/usr/share/code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:127:27011)
	at Object.writeFile (/usr/share/code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:127:25448)
```

package.json
```diff
@@ -1504,6 +1509,12 @@
         "key": "ctrl+shift+v",
         "mac": "cmd+shift+v",
         "when": "javaLSReady && editorLangId == java"
+      },
+      {
+        "command": "java.action.filesExplorerPasteAction",
+        "key": "ctrl+shift+v",
+        "mac": "cmd+shift+v",
+        "when": "(explorerViewletFocus || (workspaceFolderCount == 0 && explorerViewletVisible)) && config.editor.pasteAs.enabled"
```
rgrunber, line 1517: After some discussion, we've decided it would be best to leave out the support for "no folders opened". Partly because of the language server startup requirement, partly to avoid "overrreaching", and partly because the "New Text File" comes close enough.

src/pasteAction.ts
```diff
@@ -48,4 +54,32 @@ export async function registerOrganizeImportsOnPasteCommand(): Promise<void> {
...
+
+let serverReady = false;
+
+export async function pasteFile(folder: fs.PathLike): Promise<void>  {
+	const clipboardText: string = await env.clipboard.readText();
+	let filePath = folder.toString();
+	fs.stat(folder, async (err, stats) => {
+		// If given path to selected folder is invalid (no folder is selected)
+		if (filePath === clipboardText || stats.isFile() || (filePath === "." && workspace.workspaceFolders !== undefined)) {
+			filePath = workspace.workspaceFolders[0].uri.path;
```
rgrunber, line 67: The stacktrace I mentioned from https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2981#issuecomment-1854760809 might actually be coming from the `uri` here. Can you try using `.uri.fsPath` ? That's a lot more common.

PR #3412, opened 2023-12-04
Fix escape characters within textblock

### LINKED ISSUE #3384: Ending text blocks / multiline strings in quotes ruins syntax highlightion
This has been an issue with the extension for a very long time. I just hadn't reported it yet.
It has existed since at least `v1.18.0`. It occurs only when the text block / multiline string ends with a `\"` on the same line as the `"""` that Java uses to mark the end of said blocks/strings.

image

##### Environment
- Operating System: KDE Neon
- JDK version: `jdk-17.0.7+7`
- Visual Studio Code version: `1.84.1`
- Java extension version: `v1.18.0`

hopehadfield:
Fixes #3384 

Before:
Screenshot from 2023-12-04 15-41-28

After:
Screenshot from 2023-12-04 15-39-43

### DIFF
--- a/language-support/java/java.tmLanguage.json
+++ b/language-support/java/java.tmLanguage.json
@@ -1596,11 +1596,7 @@
 					"name": "string.quoted.triple.java",
 					"patterns": [
 						{
-							"match": "\\\\\"\"\"",
-							"name": "constant.character.escape.java"
-						},
-						{
-							"match": "\\\\.",
+							"match": "(\\\\\"\"\")(?!\")|(\\\\.)",
 							"name": "constant.character.escape.java"
 						}
 					]


language-support/java/java.tmLanguage.json
```diff
@@ -1596,11 +1596,7 @@
 					"name": "string.quoted.triple.java",
 					"patterns": [
 						{
-							"match": "\\\\\"\"\"",
-							"name": "constant.character.escape.java"
-						},
-						{
-							"match": "\\\\.",
+							"match": "(\\\\\"\"\")(?!\")|(\\\\.)",
```
rgrunber, line 1599: Just wanted to confirm. The "OR" can now be done because it would only match when the first expression fails ?

rgrunber:
It's worth noting that `\""""""` doesn't even compile so no need to handle things like that.

PR #3419, opened 2023-12-07
Make the server status as a normal status bar item

### LINKED ISSUE #3416: Make the server status a normal status bar item
#### Problem

Currently the server status is put inside the language status. The main problem of that is, user cannot see the server status unless user opens a java relate file. This experience is not ideal since the server status is hidden when the user is, for example, viewing the setting page.

#### Solution
Information like language level and project type can still be hosted by the language status bar, but put the server status as a general status bar item.

jdneo:
This actually makes the implementation much clear.

I also removed the check for the advanced status bar item because according to the theia & VS Code api comparation page, theia already **fully** supports the createStatusBarItem API: https://[eclipse-theia.github.io/vscode-theia-comparator/status.html](https://eclipse-theia.github.io/vscode-theia-comparator/status.html)

Let me know if you have any input @rgrunber @fbricon 

fix #3416

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -129,8 +129,6 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 
 	cleanJavaWorkspaceStorage();
 
-	serverStatusBarProvider.initialize();
-
 	return requirements.resolveRequirements(context).catch(error => {
 		// show error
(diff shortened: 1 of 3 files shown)


rgrunber (approved):
Change looks fine to me. Feel free to merge when ready. I actually like seeing the status at all times (particularly for a Maven project that may take time to fully build) rather than just when a file is open.

Issue #3234, opened 2023-08-02
Change the scope of `java.import.gradle.home` to `machine-overridable`


cypher256:
Is it possible to change the scope of `java.import.gradle.home` to `machine-overridable`?
I would like to have remote `settings.json` updated when `vscode.workspace.getconfiguration()` is `update` by an extension in a remote environment such as WSL.

* Related Issue
https://github.com/redhat-developer/vscode-java/pull/2624
https://github.com/redhat-developer/vscode-java/issues/2001
https://github.com/redhat-developer/vscode-java/pull/2368

`java.import.gradle.java.home` is already `machine-overridable`.
```
        "java.import.gradle.home": {
          "type": "string",
          "default": null,
          "description": "Use Gradle from the specified local installation directory or GRADLE_HOME if the Gradle wrapper is missing or disabled and no 'java.import.gradle.version' is specified.",
          "scope": "window" // <<<
        },
        "java.import.gradle.java.home": {
          "type": "string",
          "default": null,
          "description": "The location to the JVM used to run the Gradle daemon.",
          "scope": "machine-overridable" // <<<
        },
```

PR #3430, opened 2023-12-14
Change scope of 'java.import.gradle.home' to machine-overridable

hopehadfield:
As requested in #3234. A [similar change](https://github.com/redhat-developer/vscode-java/pull/2624) was made for `java.import.gradle.java.home`, and I see no issues in this case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
         "java.home",
         "java.jdt.ls.vmargs",
         "java.configuration.runtimes",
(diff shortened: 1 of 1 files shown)


rgrunber (approved):
Seems fine to me.

rgrunber:
@hopehadfield I would add the setting to `restrictedConfiguration`. We don't validate the gradle home location on the language server side, only if the project uses a Gradle wrapper.

PR #3438, opened 2023-12-20
Fix the type hierarchy

### LINKED ISSUE #3429: Show Type Hierarchy stop working

##### Environment
Version: 1.86.0-insider
Commit: 72cb92fdc5d08101d65d59e4b1c046d397896dd2
Date: 2023-12-12T05:36:41.973Z
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code-Insiders/1.86.0-insider Chrome/118.0.5993.159 Electron/27.1.3 Safari/537.36

##### Steps To Reproduce
1. The first 1~2 query for the type hierarchy returns correct result.
2. After that, following type hierarchy actions cannot work properly.

https://github.com/redhat-developer/vscode-java/assets/6193897/6eb47161-7999-42d8-9c3d-f9d5fcd7f3c7

I can see this error when it stops working:
```
ERR TreeError [references-view.tree] Data tree node not found: [object Object]
```

jdneo:
fix #3429

While there is a side effect that the same node will keep the collapse state across the requests. (Previously, every request results to all collapsed nodes.)

Personally, I think it's acceptable.


src/typeHierarchy/model.ts
```diff
@@ -132,8 +132,7 @@ class TypeHierarchyTreeDataProvider implements vscode.TreeDataProvider<TypeHiera
 				vscode.Uri.parse(element.uri), <vscode.TextDocumentShowOptions>{ selection: element.selectionRange }
 			]
 		} : undefined;
-		// workaround: set a specific id to refresh the collapsible state for treeItems, see: https://github.com/microsoft/vscode/issues/114614#issuecomment-763428052
-		treeItem.id = `${element.data["element"]}${Math.floor(Math.random() * Number.MAX_SAFE_INTEGER)}`;
+		treeItem.id = element.data?.["element"];
```
alexr00, line 135: You might consider prepending some info about what the query was that resulted in this node showing. That way, a new query would result in a the node being "reset" but re-running the same query would preserve the collapse state.

jdneo: Good suggestion! Thank you!

jdneo:
Update:

The upstream issue has been fixed, so now the feature works again in insiders.

This PR can be a fix for stable vs code (considering the next stable vs code release would happen in early Feb.).

Meanwhile it's also a kind of enhancement thanks to @alexr00's suggestion. Now if requesting for type hierarchy on the same symbol, the collapse state will be preserved. Previously, it will be reset.

PR #3439, opened 2023-12-21
Reduce delegate command calls when classpath changes

jdneo:
This PR use the project uri from the classpath update event to reduce the delegate command invocation when checking the lombok dependency.

Previously, when a classpath update event comes, the extension will check the lombok dependency for all java projects. It was not using the project uri of the classpath update event. Assume a workspace contains 10 projects, it will generate 10*10=100 getClasspath command invocation.

I did a benchmark roughly for the project: https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/spring with following steps:

1. Disable auto build (too many modules causes too much time to build the workspace)
2. Set `java.trace.server` to message 
3. Open the folder.
4. Wait until no more traces appear in the output channel.

Without the change, it takes 7 mins and 30 secs, with 3295 requests sent to server.
With this change, it takes 5 mins and 40 secs, with 347 requests sent to server.

### DIFF
--- a/src/lombokSupport.ts
+++ b/src/lombokSupport.ts
@@ -4,7 +4,7 @@ import * as fse from "fs-extra";
 import * as path from "path";
 import * as semver from 'semver';
 import * as vscode from "vscode";
-import { ExtensionContext, window, commands } from "vscode";
+import { ExtensionContext, window, commands, Uri } from "vscode";
 import { Commands } from "./commands";
 import { apiManager } from "./apiManager";
 import { languageStatusBarProvider } from './runtimeStatusBarProvider';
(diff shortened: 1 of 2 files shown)


rgrunber:
I agree with the change itself, but I don't see how this would result in such a large improvement. When I modify a pom file of some multi-maven module project, I only see `checkLombokDependency` called once as part of the classpath update, which on subsequent calls will just save re-evaluating `java.project.getAll`, but that's just a single call.

jdneo:
After `java.project.getAll`, client will call `await apiManager.getApiInstance().getClasspaths(projectUri, {scope: 'test'});` for each project. But actually, there is only one project's classpath has changed.

rgrunber:
You're right. For some reason I thought you were only re-using a cached `projectUris` instead of re-calculating, but missed that you were actually passing in just a single element. so `getClassPaths` becomes a single call and `getAllJavaProjects` is avoided.

Issue #3424, opened 2023-12-08
can we make java.cleanup.actionsOnSave trigger manually with a keyboard shortcut?


moapu:
when i want to make a small change to a repo, java.cleanup.actionsOnSave automatically changes lot of the code i wish not to change.
however, I do want this feature available when i really do need it. 

following are the ones i have set right now:

```
  "java.cleanup.actionsOnSave": [
    "addOverride",
    "stringConcatToTextBlock",
    "addFinalModifier",
    "lambdaExpressionFromAnonymousClass",
    "switchExpression",
    "tryWithResource",
    "lambdaExpression"
  ],
```

##### Environment
- Operating System: Windows 10
- JDK version: 17
- Visual Studio Code version: 1.84.1
- Java extension version: v1.25.1

##### Steps To Reproduce
1. open a repo
2. change something slightly 
3. notice it will change lot of the code you are not working with

##### Additional Information
I simply want a way for the actions to be performed with a keyboard shortcut and not on save every time.

rgrunber:
I think this is pretty reasonable. Some code bases may have parts that should not be cleaned up on save, and I don't think VS Code has per-file/folder settings where you could just set ` "java.cleanup.actionsOnSave": []`, and even that isn't as convenient as a keybinding.

This could definitely be done. It's just a matter of whether VS Code provides an existing keybinding we could override, or finding a new one that doesn't clash with anything.

moapu:
if you search for `java.cleanup.actionsOnSave` in command pallette, nothing shows up. meaning there needs to be a command registered first?

rgrunber:
`java.cleanup.actions` will be the new setting to store the list of cleanups to apply. `java.saveActions.cleanup` will be the setting to enable cleanup actions on save. <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>alt</kbd>+<kbd>s</kbd> will be the default key binding to activate cleanups (can be configured through the `java.action.doCleanup` command ID)

PR #3451, opened 2024-01-04
Add keybinding & setting to perform cleanup actions manually. 

hopehadfield:
Implements request in #3424.

Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3010.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1212,9 +1212,9 @@
       "title": "Code Action",
       "order": 110,
       "properties": {
-        "java.cleanup.actionsOnSave": {
+        "java.cleanup.actions": {
           "type": "array",
-          "markdownDescription": "The list of clean ups to be run on the current document when it's saved. Clean ups can automatically fix code style or programming mistakes. Click [HERE](command:_java.learnMoreAboutCleanUps) to learn more about what each clean up does.",
+          "markdownDescription": "The list of clean ups to be run on the current document when it's saved or when the cleanup command is issued. Clean ups can automatically fix code style or programming mistakes. Click [HERE](command:_java.learnMoreAboutCleanUps) to learn more about what each clean up does.",
           "items": {
             "type": "string",
             "enum": [
@@ -1235,6 +1235,34 @@
           "default": [],
           "scope": "window",
           "order": 10
+        },
+        "java.cleanup.actionsOnSave": {
+          "type": "array",
+          "deprecationMessage": "Deprecated, please use 'java.cleanup.actions' instead.",
+          "items": {
+            "type": "string",
+            "enum": [
+              "qualifyMembers",
+              "qualifyStaticMembers",
+              "addOverride",
+              "addDeprecated",
+              "stringConcatToTextBlock",
+              "invertEquals",
+              "addFinalModifier",
+              "instanceofPatternMatch",
+              "lambdaExpressionFromAnonymousClass",
+              "lambdaExpression",
+              "switchExpression",
+              "tryWithResource"
+            ]
+          },
+          "default": [],
+          "scope": "window"
+        },
+        "java.saveActions.cleanup": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable cleanup actions on save."
         },
         "java.saveActions.organizeImports": {
           "type": "boolean",
@@ -1560,6 +1588,11 @@
         "command": "java.action.filesExplorerPasteAction",
         "title": "%java.action.filesExplorerPasteAction%",
         "category": "Java"
+      },
+      {
+        "command": "java.action.doCleanup",
+        "title": "%java.action.doCleanup%",
+        "category": "Java"
       }
     ],
     "keybindings": [
@@ -1583,6 +1616,11 @@
         "key": "ctrl+shift+v",
         "mac": "cmd+shift+v",
         "when": "explorerViewletFocus && config.editor.pasteAs.enabled"
+      },
+      {
+        "command": "java.action.doCleanup",
+        "key": "ctrl+shift+alt+s",
+        "when": "javaLSReady && editorLangId == java"
       }
     ],
     "menus": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1536,6 +1547,11 @@
         "key": "ctrl+shift+v",
         "mac": "cmd+shift+v",
         "when": "explorerViewletFocus && config.editor.pasteAs.enabled"
+      },
+      {
+        "command": "java.action.doCleanup",
+        "key": "ctrl+shift+alt+s",
+        "when": "javaLSReady && config.java.cleanup.manualKey && editorLangId == java"
```
rgrunber, line 1554: Is `config.java.cleanup.manualKey` needed ? I don't see any such setting to test.

src/sourceAction.ts
```diff
@@ -67,6 +87,17 @@ function registerOverrideMethodsCommand(languageClient: LanguageClient, context:
     }));
 }
 
+function registerManualCleanupCommand(languageClient: LanguageClient, context: ExtensionContext): void {
+    context.subscriptions.push(commands.registerCommand(Commands.MANUAL_CLEANUP, async () => {
+        const languageClient: LanguageClient | undefined = await getActiveLanguageClient();
+        const document = window.activeTextEditor !== undefined ? window.activeTextEditor.document : undefined;
+        if (document !== undefined){
```
rgrunber, line 94: This is just an extra check for safety right ? I would assume the document should always exist if you had  something like `editorTextFocus && editorLangId == java` in the when clause ? Is it worth adding `editorTextFocus` above ?

package.json
```diff
@@ -1189,6 +1190,11 @@
           "scope": "window",
           "order": 10
         },
+        "java.saveActions.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable key for manual cleanup actions typically done on save"
```
rgrunber, line 1196: I think the description is wrong here. Applying the cleanups through a keyboard should never be guarded by a setting. The ability to perform the cleanups on save is what's guarded now.

src/protocol.ts
```diff
@@ -233,6 +233,10 @@ export namespace OrganizeImportsRequest {
     export const type = new RequestType<CodeActionParams, WorkspaceEdit, void>('java/organizeImports');
 }
 
+export namespace ManualCleanupRequest {
+    export const type = new RequestType<TextDocumentIdentifier, WorkspaceEdit, void>('java/manualCleanup');
```
rgrunber, line 237: I think I would name it just `java/cleanup`.

package.json
hopehadfield: Leftover from when I had a setting that enabled/disabled the key (as is the description below), just realized I forgot to push the change. Will update.

src/sourceAction.ts
hopehadfield: I had this for safety prior to having `editorLangId == java` in the when clause. That condition alone is sufficient for ensuring the document exists. I'll remove the extra check.

rgrunber: Ok. I would maybe verify whether `editorTextFocus && editorLangId == java"` gives us anything extra. I see it used in a few spots. I think editorLangId is just about the editor that is visible, but editorTextFocus asserts the cursor is also in the editor. Maybe we would want the cleanup to only work when the cursor is actually in the editor.

src/sourceAction.ts
```diff
@@ -67,6 +87,14 @@ function registerOverrideMethodsCommand(languageClient: LanguageClient, context:
     }));
 }
 
+function registerCleanupCommand(languageClient: LanguageClient, context: ExtensionContext): void {
+    context.subscriptions.push(commands.registerCommand(Commands.MANUAL_CLEANUP, async () => {
+        const languageClient: LanguageClient | undefined = await getActiveLanguageClient();
+        const workspaceEdit = await languageClient.sendRequest(CleanupRequest.type, languageClient.code2ProtocolConverter.asTextDocumentIdentifier(window.activeTextEditor.document));
```
rgrunber, line 93: I would just add a comment here to indicate that `MANUAL_CLEANUP` is only active when `editorlangId == java`.

package.json
```diff
@@ -54,7 +54,8 @@
     "workspaceContains:.classpath",
     "workspaceContains:*/.classpath",
     "onCommand:_java.templateVariables",
-    "onCommand:_java.metadataFilesGeneration"
+    "onCommand:_java.metadataFilesGeneration",
+    "onCommand:java.action.filesExplorerPasteActionNoLS"
```
rgrunber, line 58: Was this from an experiment ?

rgrunber (approved):
Overall looks good. Just some minor things. I think this part is ready.

package.nls.json
```diff
@@ -26,5 +26,6 @@
 	"java.clean.sharedIndexes": "Clean Shared Indexes",
 	"java.server.restart": "Restart Java Language Server",
 	"java.edit.smartSemicolonDetection": "Java Smart Semicolon Detection",
-	"java.action.filesExplorerPasteAction": "Paste clipboard text into a file"
+	"java.action.filesExplorerPasteAction": "Paste clipboard text into a file",
+	"java.action.doCleanup": "Performs cleanup actions"
```
rgrunber, line 30: I noticed most of the labels for key bindings capitalize the start of each word so I would do the same here.

package.json
hopehadfield: Honestly not sure where that came from, I missed this. Must have been some old stash from when I was doing the paste keybinding? Got rid of it.

package.json
```diff
@@ -1236,6 +1236,34 @@
           "scope": "window",
           "order": 10
         },
+        "java.cleanup.actionsOnSave": {
+          "type": "array",
+          "markdownDeprecationDescription": "**Deprecated, please use 'java.clenaup.actions' instead.**",
```
rgrunber, line 1241: This doesn't exist. There's either `markdownDeprecationMessage` or `deprecationMessage`. You can check this by typing out the property in a settings file and there should be a warning showing the deprecation message. I would go with just `deprecationMessage` though like some other instances.

PR #3456, opened 2024-01-10
Add custom Lombok 1.18.31 snapshot into lombok folder.

### LINKED ISSUE #3454: (Lombok) annotations with arguments show an error in releases after 1.25.1
[(Lombok) annotations with arguments show an error.]

Type: Bug

##### Environment
- Operating System: Windows 11 / WSL: debian
- JDK version: OpenJDK 17
- Visual Studio Code version: 1.85.1
- Java extension version: 1.25.1+

##### Steps To Reproduce
1. Install latest pre-release
2. Use a lombok annotation with a argument such as @EqualsAndHashcode(callSuper = true)

##### Current Result
Annotations with arguments mess up the annotation reading of VSCode resulting lot's of erros in the editor.

##### Expected Result
Annotations are read as before.

##### Additional Informations
The bug appeared somewhere in the first half of last December. I fixed it today by installing the current stable 1.25.1 version, instead of a prerelease.

rgrunber:
- Fixes #3454
- Base version projectlombok/lombok@0089374 along with unmerged fix from projectlombok/lombok#3565

For testing purposes, make sure to clean out any older (non lombok-1.18.31.jar) versions from your `lombok/` folder. Only remaining issue is I feel this should be named something other than 1.18.31.


fbricon:
rename it 1.18.31-3565

rgrunber:
> rename it 1.18.31-3565

https://github.com/redhat-developer/vscode-java/blob/392a255188f4b72c6e9004ca7982e06c7dd9669e/src/lombokSupport.ts#L25

Would need to modify that line. Probably `/lombok-\d+.*\-\d+.jar$/; `. So it isn't part of any version calculation.

snjeza (approved):
It works fine.

PR #3463, opened 2024-01-15
Use `source.gradle-kotlin-dsl` instead of `source.kotlin` to avoid clashes with other Kotlin grammars

fwcd:
The grammar is currently only used for Gradle Kotlin DSL (`.gradle.kts`) scripts, however uses the same scope as regular Kotlin grammars (`source.kotlin`). This causes it to silently override the Kotlin syntax highlighting, even in `.kt` files, when another grammar is installed. See https://github.com/redhat-developer/vscode-java/pull/3377#issuecomment-1891124971 for further rationale.

This fixes the issue by using a new scope for the Kotlin grammar, namely `source.gradle-kotlin-dsl`.


fbricon:
Thanks @fwcd!

PR #3476, opened 2024-01-24
Introduce capability to check if workspace has exited improperly.

### LINKED ISSUE #2405: Exceptions activating vscode-java extension
Exceptions activating vscode-java extension. The spinning wheel just keeps going in the bottom-right corner.

##### Environment
- Operating System: OSX 12.3.1
- JDK version: embedded into vscode-java
- Visual Studio Code version: 1.66.1
- Java extension version: 1.4.0

##### Steps To Reproduce
1. Open a Java file
2. Java spinning wheel spinning in the bottom right corner. Exception in debug output:
```
rejected promise not handled within 1 second: Error: write EPIPE
/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:104
stack trace: Error: write EPIPE
	at WriteWrap.onWriteComplete [as oncomplete] (node:internal[/stream_base_commons:98:16]())
	at WriteWrap.callbackTrampoline (node:internal[/async_hooks:130:17]())
/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:104
rejected promise not handled within 1 second: Error [ERR_STREAM_DESTROYED]: Cannot call write after a stream was destroyed
/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:104
stack trace: Error: Cannot call write after a stream was destroyed
	at new NodeError (node:internal[/errors:371:5]())
	at _write (node:internal[/streams/writable:321:11]())
	at Socket.write (node:internal[/streams/writable:334:10]())
	at [/Users/aboyko/.vscode/extensions/redhat.java-1.4.0-darwin-x64/dist/extension.js:2:1014845]()
	at new Promise (<anonymous>)
	at c.write ([/Users/aboyko/.vscode/extensions/redhat.java-1.4.0-darwin-x64/dist/extension.js:2:1014763]())
	at _.doWrite ([/Users/aboyko/.vscode/extensions/redhat.java-1.4.0-darwin-x64/dist/extension.js:2:1004314]())
	at [/Users/aboyko/.vscode/extensions/redhat.java-1.4.0-darwin-x64/dist/extension.js:2:1004209]()
/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:104
```

Test Project: https://github.com/spring-guides/gs-consuming-rest

rgrunber:
- Fixes #2405 
- .snap files under the workspace metadata location indicate in improper workspace shutdown

### DIFF
--- a/USAGE_DATA.md
+++ b/USAGE_DATA.md
@@ -20,9 +20,9 @@ vscode-java has opt-in telemetry collection, provided by [vscode-redhat-telemetr
  * Information about the following settings. In the case of settings that store a well defined value (eg. path/url/string), we simply collect whether the setting has been set.
    * `java.settings.url`, `java.format.settings.url`, `java.quickfix.showAt`, `java.symbols.includeSourceMethodDeclarations`, `java.completion.guessMethodArguments`, `java.completion.postfix.enabled`, `java.cleanup.actionsOnSave`, `java.sharedIndexes.enabled`, `java.inlayHints.parameterNames.enabled`, `java.server.launchMode`, `java.autobuild.enabled`
  * The extension name and the choice made when a recommendation to install a 3rd party extension is proposed
- * The name of Java commands being manually executed, and any resulting errors
+ * The name of Java commands being executed, and any resulting errors
  * The number of results (eg. 20), whether an error occured (eg. false), and duration (in milliseconds) when code assist is activated
- 
+
 ## What's included in the general telemetry data
 
 Please see the
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,5 +1,6 @@
 'use strict';
 
+import { TelemetryService } from '@redhat-developer/vscode-redhat-telemetry/lib';
 import * as chokidar from 'chokidar';
 import * as fs from 'fs';
 import * as fse from 'fs-extra';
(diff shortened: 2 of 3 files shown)


testforstephen:
This fix is fine for me, but I have one concern. It would result in a fresh import and affect the project import speed. Before we determine to apply this fix, can we create another PR to monitor how often the .snap file appears in telemetry? I want to know how likely this situation is and how many users it affects. If it’s rare, we can clean the workspace automatically. If not, we might consider looking for a better solution. What do you think?

rgrunber:
Sounds fine. I can just modify this PR to also call `fireTraceEvent` so its accessible through the API, and temporarily avoid doing the cleanup. That way everything stays mostly the same and `java.cleanworkspace`'s `reason` key having a value of `automatic` will represent the occurrence through the telemetry API.

testforstephen (approved):
LGTM

rgrunber:
I don't think we can conclude that the workspace existing with unsaved changes is always harmful. As an example :

I took a simple project (can be invisible or even a small Maven one)

1. Open the project folder in VS Code for the first time, or alternatively clear the language server workspace

On the first import, if you watch the workspace folder that corresponds to the project you should see :

watch -n1 -d 'find . -name "*.snap"'
```
./redhat.java/jdt_ws/.metadata/.plugins/org.eclipse.core.resources/0.snap
...
...
```

3. Now attempt to kill the language server at least 5 times (to ensure it stops trying to start up again). You'll know you've succeeded when you see the server status show a :-1: (thumbs down) icon.
4. Now that there's no server controlling the workspace, shut off the VS Code client.
5. The `0.snap` file from above will not have disappeared.
6. Now start VS Code again on the same workspace.

You'll see

```
!ENTRY org.eclipse.core.resources 2 10035 2024-01-25 16:00:12.263
!MESSAGE The workspace exited with unsaved changes in the previous session; refreshing workspace to recover changes.
```
but as far as I can tell, the language server starts up and works as normal. So I don't think we can conclude that this error is always harmful :\ There could be a chance that something else happened prior to the workspace exiting with unsaved changes that resulted in an unrecoverable state.

testforstephen:
> but as far as I can tell, the language server starts up and works as normal. So I don't think we can conclude that this error is always harmful :\ There could be a chance that something else happened prior to the workspace exiting with unsaved changes that resulted in an unrecoverable state.

This aligns with my previous observation. Most of the time the unsaved changes can be restored well upon reopening vscode. One case that cannot be recovered is the error `"Caused by: org.eclipse.core.internal.dtree.ObjectNotFoundException: Tree element '/spring-petclinic/target/classes/templates/welcome.html' not found."`, which requires a clean cache as the workaround.

rgrunber:
Do you have steps to generate the above error consistently ? I might be able to find another way to catch if it's likely to happen through workspace storage files. Another way is to just read the logs of the current session sent from the language server : https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/d8643fce217162487a03cfadfe90e4dbacbe6a5a/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/LogHandler.java#L111 and handle at https://github.com/redhat-developer/vscode-java/blob/d3bcbaa3f5a3097dc21b5d94132d6858a0452a7c/src/standardLanguageClient.ts#L277-L281 .

rgrunber:
See https://github.com/redhat-developer/vscode-java/pull/3509 .

PR #3478, opened 2024-01-26
Revamp server status bar item.

jdneo:
This PR made following changes to the server status bar item.

1. Move to the left. This aligns VS Code's status bar [UX guideline](https://code.visualstudio.com/api/ux-guidelines/status-bar): Place primary (global) items on the left.

2. Display the progress more explicitly
screenshot-1706253148794

3. Use background color to highlight the status item when action is required
screenshot-1706253180936
screenshot-1706253204678

4. Make the status bar item as an entrance to access other useful commands.
screenshot-1706254021696

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -350,4 +350,18 @@ export namespace Commands {
      */
     export const RESOLVE_PASTED_TEXT = "java.project.resolveText";
 
+    /**
+     * The command when clicking the server status bar item.
+     */
+    export const OPEN_STATUS_SHORTCUT = "_java.openShortcuts";
+
+}
+
+/**
+ * Command titles used to render in the UI
+ */
+export namespace CommandTitle {
+    export const OPEN_JAVA_SETTINGS = "$(settings-gear) Open Java Settings";
+    export const OPEN_LOGS = "$(output) Open Logs...";
+    export const CLEAN_WORKSPACE_CACHE = "$(trash) Clean Workspace Cache";
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,12 +5,12 @@ import * as fs from 'fs';
 import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import { CodeActionContext, commands, ConfigurationTarget, Diagnostic, env, EventEmitter, ExtensionContext, extensions, IndentAction, InputBoxOptions, languages, RelativePattern, TextDocument, UIKind, Uri, ViewColumn, window, workspace, WorkspaceConfiguration } from 'vscode';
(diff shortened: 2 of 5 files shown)


rgrunber (commented):
Overall, looks good to me. The status is a lot more visible and access to opening logs / clearing workspace should hopefully fix some issues.

Only other consideration is whether users might be annoyed if they're doing something not related to Java and yet the status is a lot more prominent now. A user mentioned this but I don't think there's much we can do if the project itself is a Java project (particularly now that we have a global status bar).

jdneo:
> Only other consideration is whether users might be annoyed if they're doing something not related to Java and yet the status is a lot more prominent now.

We can let the user to hide the java server status bar when he does not want to see it for some workspaces. It's togglable.

image

testforstephen:
> 3. Use background color to highlight the status item when action is required
>    screenshot-1706253180936
>    screenshot-1706253204678

The yellow background color is a little inharmonious with the overall VS Code appearance. Other status icons do not have any extra background color, like the error diagnostics.
image

The icon and text are sufficient to draw attention. I would rather not set the background color when the status is warning.

jdneo:
> I would rather not set the background color when the status is warning.

What about the icons, if no background color change, do we need to change the icon to ⚠️ when warnings appear?

BTW, @rgrunber, I changed the icon from 👍 to ☕, in this PR, but I'm not sure if that `coffee` codicon is available in Theia as well. Do you know if it's supported in Theia?

testforstephen:
> What about the icons, if no background color change, do we need to change the icon to ⚠️ when warnings appear?

If no proper icon for java warning, we can still coffee icon since the text expresses its meaning as well.

rgrunber:
I think it should be fine to use the coffee codicon. What about using text to represent status vs. codicons ("Ready", "Warning", "Error", :+1: , :warning: , :x:  ).

Does " :coffee:  Java: :+1: " look weird ? Benefit would be saving a bit of space at the bottom. " :coffee: : :+1: " seems it could be confusing.

jdneo:
> Does " ☕ Java: 👍 " look weird ? Benefit would be saving a bit of space at the bottom. " ☕ : 👍 " seems it could be confusing.

I prefer `☕ Java: Ready`, etc...  😄

testforstephen:
> I prefer `☕ Java: Ready`, etc... 😄

+1, icon+text is better than two icons.

testforstephen (approved):
LGTM.

To me, the status shortcut list is a good way to expose features.

One follow-up request is to enable 3rd party extensions to contribute to the status shortcut list. For instance, vscode-java-pack offers many webviews to set up Java extension. If these options could appear in the status shortcut quickpick, users would find them more easily.

rgrunber:
Just be aware we may encounter similar kinds of problems that we did trying to merge https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2821 . Opening more things to the progress reporting means if they don't end their progress reporting the bar will "spin" forever.

PR #3490, opened 2024-02-07
Jump to specific position when clicking on Javadoc link

fbricon:
That was bothering me while doing javadoc stuff ;-)

https://github.com/redhat-developer/vscode-java/assets/148698/e6a57f3a-a7cf-4a0d-9438-38479f17f3ea

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,7 +5,7 @@ import * as fs from 'fs';
 import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import { CodeActionContext, commands, ConfigurationTarget, Diagnostic, env, EventEmitter, ExtensionContext, extensions, IndentAction, InputBoxOptions, languages, QuickPickItemKind, RelativePattern, TextDocument, UIKind, Uri, ViewColumn, window, workspace, WorkspaceConfiguration } from 'vscode';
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -399,7 +399,16 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
...
-				await window.showTextDocument(parsedUri);
+				const editor = await window.showTextDocument(parsedUri);
+				// Reveal the document at the specified line, if possible (e.g. jumping to a specific javadoc method).
+				if (editor && parsedUri.scheme === 'jdt' && parsedUri.fragment) {
+					const line = parseInt(parsedUri.fragment);
+					if (isNaN(line) || line < 1 || line > editor.document.lineCount) {
+						return;
+					}
+					const range = editor.document.lineAt(line -1).range;
+					editor.revealRange(range, TextEditorRevealType.AtTop);
```
rgrunber, line 410: Any reason you went for `AtTop` instead of `InCenterIfOutsideViewport` ? That seems to match the behaviour of go-to-definition on methods.

rgrunber (commented):
Works well for me. Only thing is the reveal strategy.

src/extension.ts
fbricon: it's the only thing I tried and I found it worked well, so I rolled with it.

PR #3494, opened 2024-02-12
Dynamically remove the Equinox launcher fragment for macOS x86_64.

### LINKED ISSUE #3484: Language Support for Java couldn't create connection to server since 1.27
Hello,

     Since yesterday, my Language Support for Java was updated to 1.27 and I get the message:
Language Support for Java (Syntax Server) client: couldn't create connection to server

##### Environment
- MacOS 10.15
- JDK version: 14.0.1
- Visual Studio Code version: 1.86
- Java extension version: 1.27

##### Steps To Reproduce
1. Open a directory containing Java code with no build tool
2. Pop-up that says the Java server cannot be started

##### Current Result
Java not working

rgrunber:
- Fixes #3484 
- `os.release()` seems like the right thing to use according to https://en.wikipedia.org/wiki/Uname#Examples (and not `version()` .
- `process.platform`, `process.arch` are said to be equivalent to `os.platform()`, `os.arch()`.

### DIFF
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -1,5 +1,6 @@
 
 import * as fs from 'fs';
+import * as fse from 'fs-extra';
 import * as glob from 'glob';
 import * as net from 'net';
 import * as os from 'os';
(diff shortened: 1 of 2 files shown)


src/javaServerStarter.ts
```diff
@@ -316,3 +317,14 @@ export function parseVMargs(params: any[], vmargsLine: string) {
 		}
 	});
 }
+
+export function removeEquinoxFragmentOnDarwinX64(context: ExtensionContext) {
+	// https://github.com/redhat-developer/vscode-java/issues/3484
+	const extensionPath = context.extensionPath;
+	new glob.Glob(`${extensionPath}/server/plugins/org.eclipse.equinox.launcher.cocoa.macosx.x86_64*.jar`, (_err, matches) => {
```
datho7561, line 324: The way that `Glob` is used here, the function returns before the callback is executed. This means that some or all of the rest of `activate` will be called before the callback is invoked, meaning that the language server might be started before equinox is removed.

The best way to avoid this is to make this function async, wrap the call to `Glob` in a Promise constructor, and return the promise. For the promise constructor, you will need to call `reject` at the beginning of the callback if `_err` is not undefined and `resolve` function at the end of the callback.

Then, in `activate()`, you will need to `await` this function invocation.

datho7561 (changes requested):

src/javaServerStarter.ts
rgrunber: Went with `globSync` & `removeSync`. I noticed the logging isn't really done so perhaps there's a bug to be fixed there in the future.

PR #3495, opened 2024-02-13
Apply selected in Generate ToString()

### LINKED ISSUE #2639: Generate toString() source action should support inclusion of methods
It would be helpful to add an option to generate a toString with getters instead of attribute names because it would allow the toString to work automatically in subclasses where the getters might be overridden. Currently, the dialog only allows you to select which attributes to include in the toString. It seems like it would be relatively simple to add a checkbox (or maybe a separate dialog box) that allows the user to select whether to use attribute names or getters.

snjeza:
Fixes #2639 
Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3055


kalebers (approved):

rgrunber (approved):
Looks fine to merge after the language server change makes its way in.

PR #3498, opened 2024-02-15
Re-release 1.28.0 as 1.28.1 due to partial publishing failure.

rgrunber:
- redhat.java 1.28.0 (generic) failed to publish due to request timeout


rgrunber:
In the future, we should just try using the vsce/ovsx option `--skip-duplicate`.

PR #3537, opened 2024-03-20
Simplify the server status item click action

jdneo:
This PR simplifies the actions when click the server status bar item. The item for the server status is removed. Instead, when clicking on the server status, it will directly open the build job terminal, and show the shortcut dropdown at the meantime.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -376,7 +376,10 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 					command: Commands.CLEAN_WORKSPACE
 				});
 
-				const choice = await window.showQuickPick(items);
+				const choice = await window.showQuickPick(items, {
+					ignoreFocusOut: true,
+					placeHolder: "Press 'ESC' to close."
+				});
 				if (!choice) {
 					return;
 				}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -358,21 +358,13 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 			// Register commands here to make it available even when the language client fails
 			context.subscriptions.push(commands.registerCommand(Commands.OPEN_STATUS_SHORTCUT, async (status: string) => {
 				const items: ShortcutQuickPickItem[] = [];
-				let statusCommand: string;
 				if (status === ServerStatusKind.error || status === ServerStatusKind.warning) {
-					statusCommand = "workbench.panel.markers.view.focus";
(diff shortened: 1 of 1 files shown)


rgrunber:
I'm seeing some odd behaviour with this change. The server status terminal view does get focus upon clicking, but the quick-pick also immediately disappears.

status-bar-flicker

It seems to me like the act of showing the server status closes the already opened quick-pick since the `executeCommand` runs asynchronously. I tried adding an `await` but still doesn't work. Occasionally it does stay open, so it's definitely timing related.

jdneo:
Ah, good catch!

I tried first adding the `preserveFocus` parameter when show the status terminal, but that did not work.

Then I went to the current solution that adding `ignoreFocusOut` flag when displaying the dropdown list. It did work, but a side effect is that now user needs to press esc to close it.

src/extension.ts
```diff
@@ -384,7 +376,10 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 					command: Commands.CLEAN_WORKSPACE
 				});
 
-				const choice = await window.showQuickPick(items);
+				const choice = await window.showQuickPick(items, {
+					ignoreFocusOut: true,
+					placeHolder: "Press 'ESC' to close."
```
rgrunber, line 381: This is true in the default case, but it's just controlled by `workbench.action.closeQuickOpen`. So technically someone could change that option (for eg. make it <kbd>shift+esc</kbd>) and <kbd>esc</kbd> would n longer work.

Does VS Code have API to get the main keyboard shortcut currently represented by some command ?

rgrunber (approved):
Change looks fine and works well. I found one minor thing so if it can be addressed, feel free to do so. If not, I'm also fine to merge.

src/extension.ts
jdneo: Found this answer: https://github.com/microsoft/vscode-discussions/discussions/780

Looks like there is no command but we can inspect the json file to check if user sets other keybindings.

I tend to keep the current simple implementation, but at least, we know there is a solution if we have to address it in the future.

rgrunber:
I can definitely confirm (at least on Linux) that the first activation of the server status bar (mouse click), followed by usage of the arrow keys to navigate the quick-pick works. However, upon activating it a second time (mouse click), the arrow keys have no effect, as the quick-pick does not appear to have focus.

jdneo:
@rgrunber Should be fixed by https://github.com/redhat-developer/vscode-java/pull/3548

PR #3548, opened 2024-03-28
Fix the focus issue of the java shortcuts

jdneo:
fix https://github.com/redhat-developer/vscode-java/pull/3537#issuecomment-2022748711

For some reason, I guess I missed something that I thought `preserveFocus` did not work by mistake.

This time, when I try it, it's working. So we can also get rid of the keybinding issue.

@rgrunber Sorry for the confusion I made.


rgrunber:
Nice! Works perfectly for me :rocket:

PR #3552, opened 2024-03-28
Regex update and sealed removed.

rgrunber:
- Make `.` optional in the regex. Also, remove `sealed` class tests.

@mavaddat I've created your PR here. I removed the `testAddPermitsToDirectSuperClass` as well. I wrongly (only) said remove `sealed` but should have also specified `permits` also. It's the keyword that sealed classes use to restrict their usage.


mavaddat:
Just verifying: It looks like you did everything yourself, right @rgrunber? It doesn't look like those tests are enabled anymore, correct?

PR #3558, opened 2024-04-03
Enable javac-based compilation

fbricon:
Requires [building eclipse.jdt.ls](https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3123) against [dom-with-javac branch of the JDT fork](https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/tree/dom-with-javac)
Requires running with Java 23 at the moment. 

First build the server jars with: 
> npm run build-server

then generate the vsix with:

> npx @vscode/vsce package

Enable with `"java.jdt.ls.javac.enabled":"on"` in vscode settings

You can enable the more experimental DOM-based completion with `"java.completion.engine": "dom"`:

Screenshot 2024-07-10 at 13 35 03

### DIFF
--- a/package.json
+++ b/package.json
@@ -409,6 +409,17 @@
             "scope": "window",
             "order": 90
           },
+          "java.jdt.ls.javac.enabled": {
+            "type": "string",
+            "enum": [
+              "on",
+              "off"
+            ],
+            "default": "off",
+            "markdownDescription": "[Experimental] Specify whether to enable Javac-based compilation in the language server. Requires running this extension with Java 23",
+            "scope": "window",
+            "order": 95
+          },
           "java.trace.server": {
             "type": "string",
             "enum": [
@@ -1008,6 +1019,21 @@
             "description": "Enable/disable code completion support",
             "scope": "window",
             "order": 10
+          },
+          "java.completion.engine": {
+            "type": "string",
+            "default": "ecj",
+            "description": "[Experimental] Select code completion engine",
+            "scope": "window",
+            "enum": [
+              "ecj",
+              "dom"
+            ],
+            "markdownEnumDescriptions": [
+              "Use ECJ-based code completion engine (default)",
+              "Use (highly experimental) JDT DOM-based code completion engine (requires `java.jdt.ls.javac.enabled` to be `on`)"
+            ],
+            "order": 1000
           },
           "java.completion.postfix.enabled": {
             "type": "boolean",
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -104,7 +104,40 @@ function prepareParams(requirements: RequirementsData, workspacePath, context: E
 				// See https://github.com/redhat-developer/vscode-java/issues/2264
 				// It requires the internal API sun.nio.fs.WindowsFileAttributes.isDirectoryLink() to check if a Windows directory is symlink.
 				'--add-opens',
-				'java.base/sun.nio.fs=ALL-UNNAMED');
+				'java.base/sun.nio.fs=ALL-UNNAMED'
+				);
+
+	const javacEnabled = 'on' === getJavaConfiguration().get('jdt.ls.javac.enabled');
+	if (javacEnabled) {
+		// Javac flags
+		params.push(
+		'--add-opens',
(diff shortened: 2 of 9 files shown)


fbricon:
Daily vscode-javac builds are available from https://github.com/fbricon/vscode-java/releases/tag/javac-prototype

src/requirements.ts
```diff
@@ -10,7 +10,7 @@ import { logger } from './log';
 import { checkJavaPreferences } from './settings';
 import { listJdks, sortJdksBySource, sortJdksByVersion } from './jdkUtils';
 
-const REQUIRED_JDK_VERSION = 17;
+const REQUIRED_JDK_VERSION = 23;
```
rgrunber, line 13: As discussed, this should probably depend on `java.jdt.ls.javac.enabled`, since for ECJ, 17 is just fine.

src/javaServerStarter.ts
```diff
@@ -104,7 +104,40 @@ function prepareParams(requirements: RequirementsData, workspacePath, context: E
...
+		'--add-opens',
+		'jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED',
+		'--add-opens',
+		'jdk.javadoc/jdk.javadoc.internal.doclets.formats.html.taglets.snippet=ALL-UNNAMED --add-opens jdk.javadoc/jdk.javadoc.internal.doclets.formats.html.taglets=ALL-UNNAMED',
+		'-DICompilationUnitResolver=org.eclipse.jdt.core.dom.JavacCompilationUnitResolver',
+		'-DCompilationUnit.DOM_BASED_OPERATIONS=true',
+		'-DAbstractImageBuilder.compilerFactory=org.eclipse.jdt.internal.javac.JavacCompilerFactory'
+		);
+
+		if('dom' === getJavaConfiguration().get('java.completion.engine')){
```
rgrunber, line 137: `getJavaConfiguration()` already returns everything under `java`, so this evaluates to `java.java.completion.engine`, which isn't set, so completion is never truly on. Just needs to be `'completion.engine'`

src/settings.ts
```diff
@@ -137,7 +137,9 @@ function hasJavaConfigChanged(oldConfig: WorkspaceConfiguration, newConfig: Work
 		|| hasConfigKeyChanged('server.launchMode', oldConfig, newConfig)
 		|| hasConfigKeyChanged('sharedIndexes.location', oldConfig, newConfig)
 		|| hasConfigKeyChanged('transport', oldConfig, newConfig)
-		|| hasConfigKeyChanged('diagnostic.filter', oldConfig, newConfig);
+		|| hasConfigKeyChanged('diagnostic.filter', oldConfig, newConfig)
+		|| hasConfigKeyChanged('jdt.ls.javac.enabled', oldConfig, newConfig)
+		|| hasConfigKeyChanged('java.completion.engine', oldConfig, newConfig);
```
testforstephen, line 142: -> `completion.engine`

src/javaServerStarter.ts
testforstephen: @rgrunber Can you take over the PR, resolve the issue, and handle the merge?

src/settings.ts
fbricon: fixed

src/javaServerStarter.ts
fbricon: fixed

package.json
```diff
@@ -409,6 +409,17 @@
             "scope": "window",
             "order": 90
           },
+          "java.jdt.ls.javac.enabled": {
+            "type": "string",
```
rgrunber, line 413: Are we ok with this being `boolean` ? Are we future-proofing for some kind of "auto"behaviour here ?

fbricon: indeed I was thinking about auto

snjeza:
@viswatejan Could you attach a project example reproducing the error?

viswatejan:
@snjeza , I could not reproduce the same issue with a sample project but I see some other issues related to the autocompletion with the [eclipse lombok agent](https://github.com/projectlombok/lombok/blob/master/src/eclipseAgent/lombok/eclipse/agent/PatchDiagnostics.java).

Delomboking and removing the Lombok dependency in my project loaded the project fine in the Java Project Explorer.

Issue #3568, opened 2024-04-09
Java string paste not auto escape


AYiXi:
after update, paste auto escape funtion is not working

Before Update:
String s = "{\\"a\\": 1}";

After Update:
String s = "{"a": 1}";

// my settings.json
{
    "editor.pasteAs.enabled": true,
    "editor.pasteAs.showPasteSelector": "afterPaste",
}

rgrunber:
I noticed that the language server is sending the correct, escaped content, but the client just isn't applying it.

https://github.com/redhat-developer/vscode-java/blob/e578b506363f50d4798e1d2c02b76727dedcb314/src/pasteEventHandler.ts#L64

https://github.com/microsoft/vscode/commit/091987d9431154a3624ba20bc1d6de65d71d6472#diff-b553efc31b13377c3b4765704b0e0a681d7489d1cb819194f327a589def055c8L40-R85

Looks like the API was updated for 1.88.0.

rgrunber:
This should be fixed in the next pre-release. Thank-you for filing this!

PR #3570, opened 2024-04-09
Adjust the document paste handling provider.

rgrunber:
- provideDocumentPasteEdits API was updated in VS Code 1.88.0
- Fixes https://github.com/redhat-developer/vscode-java/issues/3568

CC'ing @testforstephen for awareness. I would guess we're fine to not support the older API for those staying on previous versions.


snjeza (approved):
LGTM

testforstephen:
what happens if the user uses copy & paste in an old VS Code version (e.g.1.87) against this PR change?

rgrunber:
From trying out the change on an older VS Code (eg. 1.87), it simply pastes the text as usual with no modification. AFAICT the paste handling request isn't even sent to the language server.

testforstephen:
I see. As long as it doesn't break the default system copy/paste behavior in an older vscode, it's fine to merge this PR.

PR #3587, opened 2024-04-23
Change default LS transport from 'stdio' to 'pipe'.

rgrunber:
`pipe` has been the default for pre-releases / Code insiders instances since 1.24.0 (October 2023). Is it time we made it the default for everyone ?


testforstephen (approved):
LGTM

PR #3646, opened 2024-05-06
Improve paste edit support across different VS Code versions.

### LINKED ISSUE #3631: The issue of being unable to use copy and paste occurs in version 1.30.0.
While programming Java with vscode, after upgrading the "Language Support for Java(TM) by Red Hat" to version 1.30.0, the copy and paste functions do not work. I downgraded to 1.29.0 just in case, and it works properly. Could this be a bug?

##### Environment
- Operating System: Windows 10 Pro / 22H2 / 19045.4291
- JDK version: zulu-17
- Visual Studio Code version: 1.81.1
- Java extension version: 1.30.0

##### Steps To Reproduce
1. Copy text with ctrl + c
2. Paste text with ctrl + v
3. No response to paste
4. When downgrading the Java extension version to 1.29.0, it works properly.

##### Additional Informations
When downgrading the Java extension version to 1.29.0, it works properly.

rgrunber:
- Use VS Code version to determine whether a DocumentPasteEdit vs. DocumentPasteEdit[] is returned
- Fixes #3631

### DIFF
--- a/src/pasteEventHandler.ts
+++ b/src/pasteEventHandler.ts
@@ -1,8 +1,9 @@
-import { CancellationToken, commands, DataTransfer, DocumentPasteEdit as VDocumentPasteEdit, DocumentPasteEditProvider, DocumentPasteProviderMetadata, ExtensionContext, languages, Range, TextDocument, window, DocumentPasteEditContext, ProviderResult, DocumentPasteEdit } from "vscode";
+import { CancellationToken, commands, DataTransfer, DocumentPasteEdit as VDocumentPasteEdit, DocumentPasteEditProvider, DocumentPasteProviderMetadata, ExtensionContext, languages, Range, TextDocument, window, DocumentPasteEditContext, ProviderResult, DocumentPasteEdit, version } from "vscode";
 import { FormattingOptions, Location, WorkspaceEdit as PWorkspaceEdit } from "vscode-languageclient";
 import { LanguageClient } from "vscode-languageclient/node";
 import { Commands } from "./commands";
 import { JAVA_SELECTOR } from "./standardLanguageClient";
+import * as semver from 'semver';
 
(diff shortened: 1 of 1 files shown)


src/pasteEventHandler.ts
```diff
@@ -62,12 +63,13 @@ class PasteEditProvider implements DocumentPasteEditProvider {
 		}
 	}
 
-	async provideDocumentPasteEdits?(document: TextDocument, ranges: readonly Range[], dataTransfer: DataTransfer, context: DocumentPasteEditContext, token: CancellationToken): Promise<DocumentPasteEdit[]> {
+	async provideDocumentPasteEdits?(document: TextDocument, ranges: readonly Range[], dataTransfer: DataTransfer, context: DocumentPasteEditContext, token: CancellationToken): Promise<any> {
```
rgrunber, line 66: @datho7561 , specifically is this ok ? The return value for `DocumentPasteEditProvider` is listed as `ProviderResult<T[]>` but I'd like to be able to support both `DocumentPasteEdit` (older) & `DocumentPasteEdit[]` (newer).

rgrunber: The crash occurs at https://github.com/microsoft/vscode/blob/6445d93c81ebe42c4cbd7a60712e0b17d9463e97/src/vs/workbench/api/common/extHostLanguageFeatures.ts#L560 (on VS Code 1.81.0) because `insertText` is never checked for existence, so it treats the returned array as if it were a single element. I don't see how we would avoid crashing without checking the version and if we do that, we may as well support the full feature by just returning the single element.

datho7561: This makes sense to me. The other way around this would be to bump the minimum supported version of VS Code, but that might not be good since then the most recent supported release for those versions is faulty.

Issue #3561, opened 2024-04-03
lombok @Singular cause Lombok annotation handler class lombok.eclipse.handlers.HandleBuilder failure


lamplification:
Lombok @Singular cause Lombok annotation handler class lombok.eclipse.handlers.HandleBuilder failed

##### Environment
- Operating System: Ubuntu x86_64 x86_64 x86_64 GNU/Linux
- JDK version: 17
- Visual Studio Code version: 1.87.2
- Java extension version: Language Support  for Java(TM) RedHat 1.29.0

##### Steps To Reproduce
1. create a demo java project
2. create a simple POJO with use of @Singular and @Builder
3. see Problem tab - Lombok annotation handler class lombok.eclipse.handlers.HandleBuilder failed - See error log.
![image](https://github.com/redhat-developer/vscode-java/assets/4444408/65158d30-2d7c-4717-b9a9-5154088208f4)

[demo.zip](https://github.com/redhat-developer/vscode-java/files/14858792/demo.zip)

##### Current Result
Lombok annotation handler class lombok.eclipse.handlers.HandleBuilder failed - See error log.

##### Expected Result
No issue

##### Additional Informations
Comment @Singular in the Domain.java, everything work as expected.  It is working in Intellij.

snjeza:
This is an upstream lombok issue caused by https://github.com/eclipse-jdt/eclipse.jdt.core/commit/cb6774f3ba36fb8dafc3a4834eb22799072db6ef

lamplification:
Thank you @snjeza , I rolled back to v1.27.0 for Language Support for Java(TM) by Red Hat. That version is working.

sashirestela:
> Thank you @snjeza , I rolled back to v1.27.0 for Language Support for Java(TM) by Red Hat. That version is working.

Rolling back to v1.28.1 is working too.

sashirestela:
@snjeza Please, could you tell us if you have any estimated time for a fixing here?

snjeza:
@sashirestela This is a lombok issue. You could ask at https://github.com/projectlombok/lombok/issues/3648

ekeric13:
how do you rollback a vscode extension version?

azriel-healthpoint:
@ekeric13 Click on the drop down next to Uninstall:

<img width="623" alt="image" src="https://github.com/redhat-developer/vscode-java/assets/138735539/9d23c198-b685-49e0-b90a-153e591f27d0">

Then this list appears:

[...]

PR #3674, opened 2024-06-11
Adopt the Lombok 1.18.33 edge release.

marco-belonio:
https://github.com/redhat-developer/vscode-java/issues/3561

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -127,14 +127,18 @@ gulp.task('download_lombok', async function (done) {
 	}
 
 	await new Promise(function (resolve, reject) {
-		const lombokVersion = '1.18.32';
+		const lombokVersion = '1.18.33';
 		// The latest lombok version can be found on the website https://projectlombok.org/downloads
-		const lombokUrl = `https://projectlombok.org/downloads/lombok-${lombokVersion}.jar`;
+		const lombokUrl = `https://projectlombok.org/lombok-edge.jar`;
 		download(lombokUrl)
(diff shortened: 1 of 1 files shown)


blizzy78:
But edge != latest version?

sleberknight:
From the [Downloads](https://projectlombok.org/download) page, it looks like you can use the generic link:

https://projectlombok.org/downloads/lombok.jar

I am _assuming_ (but don't know for sure) that this link will always get the latest stable version. As of right now, that is 1.18.32 - the link text is _Download 1.18.32_

So maybe just need to use this link and then will always get the most recent stable version?

rgrunber:
Thanks for getting this started. I remember when the original issue was filed but didn't see any progress for a long while.

Could you make the change look like the gulpfile at https://github.com/redhat-developer/vscode-java/commit/b59c43a2f7e5c2963084d8f7c6f55ade50e8cd02#diff-25789e3ba4c2adf4a68996260eb693a441b4a834c38b76167a120f0b51b969f7 ? This won't require changing any runtime code. Instead of `1.18.31` (the unreleased version that "edge" was targetting then), call it `1.18.33`.

If the edge release contains a fix, we can definitely adopt.

marco-belonio:
@sleberknight The problem is, that the 1.18.32 does not work to resolve the problem :S

sleberknight:
> @sleberknight The problem is, that the 1.18.32 does not work to resolve the problem :S

Oops. Sorry! Clearly I missed that important part.

PR #3680, opened 2024-06-21
Check the pipe name when transport kind is default

jdneo:
`generateRandomPipeName()` is not called in default switch section.


testforstephen (approved):
Nice catch.

PR #3702, opened 2024-07-11
Fix class links in completion documentation

### LINKED ISSUE #3697: Links to classes do not appear in completion documentation
##### Environment
- Operating System: Fedora 40
- JDK version: The one bundled with the extension
- Visual Studio Code version:
```
1.91.0
ea1445cc7016315d0f5728f8e8b12a45dc0a7286
x64
```
- Java extension version: v1.33.2024062908 (pre-release)

##### Steps To Reproduce
1. Make a file `MyDocumentedClass.java`:
```java
package org.acme;

/**
 * This is some documentation, isn't it so cool?
 * 
 * @see System
 */
public class MyDocumentedClass {
    
    public void printSomeMessage() {
        System.out.println("Hello, World!");
    }

    public void myCoolMethod() {
        MyDocumented
    }
}
```
2. Hover over `MyDocumentedClass`. Notice how `System` is a link, and when you click it, it opens the `System` class.
3. Invoke completion after `MyDocumented`. Notice how the same documentation appears, but 

##### Current Result

Documentation links do not work in completion documentation.

##### Expected Result

Documentation links in completion documentation work the same as in hover documentation.

##### Additional Informations

See https://github.com/redhat-developer/vscode-microprofile/pull/246 for a fix I did in vscode-microprofile.

hopehadfield:
Fixes #3697

Before:
 
link-doesn't-work

After:

link-works

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -6,7 +6,7 @@ import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
 import * as semver from 'semver';
-import { CodeActionContext, commands, ConfigurationTarget, Diagnostic, env, EventEmitter, ExtensionContext, extensions, IndentAction, InputBoxOptions, languages, QuickPickItemKind, RelativePattern, TextDocument, TextEditorRevealType, UIKind, Uri, ViewColumn, window, workspace, WorkspaceConfiguration } from 'vscode';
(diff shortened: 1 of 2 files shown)


snjeza (approved):
It works fine for me.

rgrunber (approved):
Works well for me.

I just noticed this is likely possible using https://code.visualstudio.com/api/references/vscode-api#CompletionItemProvider & https://code.visualstudio.com/api/references/vscode-api#CompletionItemProvider.resolveCompletionItem . I guess at this point if we're using middleware, we might as well stick to that. On the other hand, we went with `languages.registerHoverProvider` to do roughly the same as we're doing here.

PR #3724, opened 2024-07-26
Clean old autogenerated globalStorage folders (#2597)

### LINKED ISSUE #2597: Clear out-of-date files in global storage
The extension will store some caches in global storage with each version. Each version will occupy around 15~20MB, which the total size will be quite large when users switched to the pre-release channel.

image

We can consider detect the out-of-date version and remove it.

logonoff:
- Fixes #2597

Delete all folders within `.../Code/User/globalStorage/redhat.java/` that are not named the current version as defined by package.json upon activating the extension

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -14,7 +14,7 @@ import { ClientErrorHandler } from './clientErrorHandler';
 import { Commands, CommandTitle } from './commands';
 import { ClientStatus, ExtensionAPI, TraceEvent } from './extension.api';
 import * as fileEventHandler from './fileEventHandler';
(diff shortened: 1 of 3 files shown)


logonoff:
Here it is working on Windows

https://github.com/user-attachments/assets/71b5352a-e42e-4127-86ca-024f0c2c9ac0

rgrunber (approved):
LGTM. Thanks for handling this. This also opens up the possibility to add this to `cleanJavaWorkspaceStorage()`. Although we'd probably have to rename that command since it's growing to clean things beyond the workspace storage.

Also, as a note, our `deleteDirectory` doesn't really handle read-only files, but I haven't really seen that be an issue.

PR #3728, opened 2024-07-31
Changelog for 1.33.0.

rgrunber:
- Adjust project compliance to reflect drop of pre-1.8 support in JDT

Due to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2551 (in the I-builds of late July).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## 1.33.0 (August 1st, 2024)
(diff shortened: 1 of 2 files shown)


testforstephen (approved):
LGTM.

Just one question about drop support for pre-1.8. Should we mention in the changelog that this version no longer supports Java versions (< 1.8) earlier than 1.8?

rgrunber:
> LGTM.
> 
> Just one question about drop support for pre-1.8. Should we mention in the changelog that this version no longer supports Java versions (< 1.8) earlier than 1.8?

Sounds reasonable. I'll add an entry mentioning in both JDT-LS & vscode-java

Issue #3686, opened 2024-06-26
Support linked correction proposals


hopehadfield:
Add support for linked correction proposals, as is done in Eclipse

![linked-proposal](https://github.com/redhat-developer/vscode-java/assets/115827695/482c0d79-0806-4a4e-88ae-7e607d47075c)

Currently, the same quickfix in vscode-java only gives the first option from the list.

![no-linked-proposal](https://github.com/redhat-developer/vscode-java/assets/115827695/740a8d00-170d-4ab3-84d2-eb66ae997150)

rgrunber:
The LSP Spec now supports [`SnippetTextEdit`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocumentEdit), so this is possible for code actions. You may have some difficulty sending it on the language server side with the existing lsp4j library as [it lacks the necessary classes](https://github.com/eclipse-lsp4j/lsp4j/issues/810), but should be possible to work around that. The language client we use (vscode-languageserver-node) also [seems to now support it](https://github.com/microsoft/vscode-languageserver-node/issues/1414).

fbricon:
as. usual, make sure the client supports it before sending SnippetTextEdits

rgrunber:
An update from @hopehadfield :

![dropdown](https://github.com/user-attachments/assets/5660ddc7-e35d-429a-996b-3a990029b2cb)

rgrunber:
We may have to live with the side effects of https://github.com/microsoft/vscode/issues/63129 since the server-side snippets get an additional indentation.

PR #3730, opened 2024-07-31
Add support for SnippetTextEdit in code actions

hopehadfield:
Fixes https://github.com/redhat-developer/vscode-java/issues/3686

dropdown

Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3229

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -317,43 +317,44 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 								}
 								const docChanges = result.edit !== undefined ? result.edit.documentChanges : undefined;
 								if (docChanges !== undefined) {
-									for (const editType of docChanges) {
-										if ("textDocument" in editType) {
-											for (const edit of editType.edits) {
+									for (const docChange of docChanges) {
+										if ("textDocument" in docChange) {
+											for (const edit of docChange.edits) {
 												if ("snippet" in edit) {
-													documentUris.push(editType.textDocument.uri);
-													snippetEdits.push(new SnippetTextEdit(asRange((edit as any).range), new SnippetString((edit as any).snippet.value)));
+													documentUris.push(docChange.textDocument.uri);
+													snippetEdits.push(new SnippetTextEdit(client.protocol2CodeConverter.asRange((edit as any).range), new SnippetString((edit as any).snippet.value)));
 												}
 											}
 										}
 									}
-								}
-								const codeAction = await client.protocol2CodeConverter.asCodeAction(result, token);
-								const docEdits = codeAction.edit !== undefined? codeAction.edit.entries() : [];
-								const newWorkspaceEdit = new WorkspaceEdit();
-								for (const doc of docEdits) {
-									const uri = doc[0];
-									if (documentUris.includes(uri.toString())) {
-										const editList = [];
-										for (const edit of doc[1]) {
-											let isSnippet = false;
-											snippetEdits.forEach((snippet, index) => {
-												if (edit.range.isEqual(snippet.range) && documentUris[index] === uri.toString()) {
-													editList.push(snippet);
-													isSnippet = true;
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -303,7 +304,62 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
...
+						const documentUris = [];
+						const snippetEdits = [];
+						const resolveCodeAction = async (item, token) => {
+							return client.sendRequest(CodeActionResolveRequest.type, client.code2ProtocolConverter.asCodeActionSync(item), token).then(async (result) => {
+								if (token.isCancellationRequested) {
+									return item;
+								}
+								const docChanges = result.edit !== undefined ? result.edit.documentChanges : undefined;
+								if (docChanges !== undefined) {
+									for (const editType of docChanges) {
```
rgrunber, line 320: I would rename `editType` to `docChange` just to make it easier to understand.

src/extension.ts
```diff
@@ -303,7 +304,62 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
...
+													snippetEdits.push(new SnippetTextEdit(asRange((edit as any).range), new SnippetString((edit as any).snippet.value)));
+												}
+											}
+										}
+									}
+								}
+								const codeAction = await client.protocol2CodeConverter.asCodeAction(result, token);
+								const docEdits = codeAction.edit !== undefined? codeAction.edit.entries() : [];
+								const newWorkspaceEdit = new WorkspaceEdit();
+								for (const doc of docEdits) {
```
rgrunber, line 334: I would rename `doc` to `docEdit`. Hovering over `docEdits`, it's pretty clear each element of the array is its own array  of the `[Uri, TextEdit[]]` so that part is fine.

src/extension.ts
```diff
@@ -1187,4 +1243,8 @@ function registerRestartJavaLanguageServerCommand(context: ExtensionContext) {
 	}));
 }
 
+function asRange(value) {
```
rgrunber, line 1246: The function is at https://github.com/microsoft/vscode-languageserver-node/blob/release/server/8.2.0-next.3/client/src/common/protocolConverter.ts#L394-L396 right ? Can't you call the library with `client.protocol2CodeConverter.asRange(..)` ?

I think when I mentioned needing to copy stuff from the vscode-languageserver-node library, I meant maybe any logic that is new and meant to deal with converting lsp snippets to vscode snippets. However converting lsp ranges to vscode ranges is always supported.

rgrunber (commented):
Just some small fixes.

src/extension.ts
```diff
@@ -303,7 +304,62 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
...
+											for (const edit of editType.edits) {
+												if ("snippet" in edit) {
+													documentUris.push(editType.textDocument.uri);
+													snippetEdits.push(new SnippetTextEdit(asRange((edit as any).range), new SnippetString((edit as any).snippet.value)));
+												}
+											}
+										}
+									}
+								}
+								const codeAction = await client.protocol2CodeConverter.asCodeAction(result, token);
```
rgrunber, line 331: If `docChanges` is `undefined` (some other code action returns not related to snippets), then we end up here, but there's really no point because this code is not going to change anything.

Seems to me it's a bit safer to move the logic below into the `if (docChanges !== undefined) {` block. Let me know what you think.

hopehadfield: I agree. Added this check just to prevent the other code actions that didn't have document changes from breaking. Moving the rest of the logic avoids running some unnecessary code at a very minimum, and definitely safer.

rgrunber (approved):
The client-side code here looks good, and is ready to be merged once the LS side is in.

rgrunber:
Random fact learned : `CodeAction.set(Uri, null)` will clear the edits for a particular `Uri` but a non-null 2nd argument will merely append. See https://github.com/microsoft/vscode/issues/197315

PR #3731, opened 2024-08-01
Recognize custom file extensions

snjeza:
Fixes https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3222
Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3230

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -208,7 +208,7 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 					{ scheme: 'untitled', language: 'java' }
 				],
 				synchronize: {
-					configurationSection: ['java', 'editor.insertSpaces', 'editor.tabSize'],
+					configurationSection: ['java', 'editor.insertSpaces', 'editor.tabSize', "files.associations"],
 				},
 				initializationOptions: {
 					bundles: collectJavaExtensions(extensions.all),
@@ -1061,7 +1061,7 @@ async function getTriggerFiles(): Promise<string[]> {
 function getJavaFilePathOfTextDocument(document: TextDocument): string | undefined {
 	if (document) {
 		const resource = document.uri;
-		if (resource.scheme === 'file' && resource.fsPath.endsWith('.java')) {
+		if (resource.scheme === 'file' && document.languageId === "java") {
 			return path.normalize(resource.fsPath);
 		}
 	}
--- a/src/fileEventHandler.ts
+++ b/src/fileEventHandler.ts
@@ -174,7 +174,25 @@ function getWillRenameHandler(client: LanguageClient) {
 }
 
 function isJavaFile(uri: Uri): boolean {
-    return uri.fsPath && uri.fsPath.endsWith(".java");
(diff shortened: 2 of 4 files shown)


src/extension.ts
```diff
@@ -1061,7 +1061,7 @@ async function getTriggerFiles(): Promise<string[]> {
 function getJavaFilePathOfTextDocument(document: TextDocument): string | undefined {
 	if (document) {
 		const resource = document.uri;
-		if (resource.scheme === 'file' && resource.fsPath.endsWith('.java')) {
+		if (resource.scheme === 'file' && document.languageId === "java") {
```
rgrunber, line 1064: If you look inside https://github.com/redhat-developer/vscode-java/blob/ab293222bb78dc130923949d057e38fe83366e5d/src/fileEventHandler.ts#L176-L178 , we still have some logic there regarding `endsWith('.java')`. I think that logic handles things like renaming, and auto-inserting snippets into newly created files, which currently aren't handled. Those should be updated as well if possible.

As a small note this may have implications for the language server as well. For example, currently this doesn't quite work (likely due to hard-coding) :

**.vscode/settings.json**
```
{
  "java.saveActions.cleanup": true
  "java.cleanup.actions": [ "renameFileToType" ]
}
```

**Test.maxj**
```
package org.example;
public class Test {
}
```

Renaming the "Test" type declaration to "Testing" renames "Test.maxj" to "Testing.java".

src/extension.ts
```diff
@@ -208,7 +208,7 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 					{ scheme: 'untitled', language: 'java' }
 				],
 				synchronize: {
-					configurationSection: ['java', 'editor.insertSpaces', 'editor.tabSize'],
+					configurationSection: ['java', 'editor.insertSpaces', 'editor.tabSize', "files.associations"],
```
rgrunber, line 211: We only support a small subset of the glob pattern syntax (*. or file name), but because we don't own the setting, there's no way to really document this. Is it worth validating+warning when a user targets 'java' with a pattern we don't support ?

snjeza: Yes, it is. I will add a warning.

src/extension.ts
snjeza: It has been fixed.

src/extension.ts
snjeza: Fixed.

snjeza:
@rgrunber I have updated the PR. Could you, please, try now?

src/extension.ts
rgrunber: The class snippet should have been auto-inserted, which is handled by the client-side `isJavaFile` method.

https://github.com/user-attachments/assets/9075c2c6-5b6c-47a5-912e-04d0a24dfb0d

snjeza: Fixed.

rgrunber (approved):
Change looks fine to me. Just should be merged after the JDT-LS part is approved.

PR #3748, opened 2024-08-24
Fix #2649: Add support for choosing search scope

### LINKED ISSUE #2649: Support for predefined workingsets for search operations
Add support to switch workingset such as "Main Source", "Test Source" for search operations such as 
- Call Hierarchy
- Type Hierarchy
- Reference Search
- Implementation Search
etc.

A simple implementation would be to introduce a server setting which will define the default workingset, by default this will be empty or include all main and test sources.

Then from the UI we could provide a picker to switch between Main, Test and All which are predefine values. We will show the current workingset in the status bar as well and clicking on that will trigger picker as well.

On the server side this setting will be used in the all Java search queries to define the scope of the search.

##### Environment
- Operating System:
- JDK version: 17
- Visual Studio Code version: 1.70.2
- Java extension version: 1.10.0

gayanper:
Provide two scopes to search from for search operations such as
- reference search
- call hierarchy search

Fixes https://github.com/redhat-developer/vscode-java/issues/2649


gayanper:
Adds the following status bar item to toggle and see the current scope in use

image

Clicking on the status bar item or triggering from command pallet will provide the following scope picker list

image

The command in the command pallet
image

package.json
```diff
@@ -1498,6 +1498,19 @@
             "default": "ask",
             "markdownDescription": "Specifies whether to reload the sources of the open class files when their source jar files are changed.",
             "scope": "window"
+          },
+          "java.search.scope": {
```
rgrunber, line 1502: It makes more sense to place the setting under https://github.com/redhat-developer/vscode-java/blob/1d7163f2222ef6dea01c347ce9a432a7c8a13de3/package.json#L1367-L1370 . We started grouping our settings by categories a little while ago.

rgrunber (commented):
Just some initial comments. The main one being I'd prefer to place the command under the server status bar with the other shortcuts rather than having its own status bar item at the bottom.

We need to be careful not to have too many.

package.json
```diff
@@ -1498,6 +1498,19 @@
...
+            "type": "string",
+            "enum": [
+              "all",
+              "main"
```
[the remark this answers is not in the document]
gayanper: Will runtime gives a wrong message considering the runtime scope maven and gradle dependencies ?

gayanper:
Other the name 'main' and 'test' for scopes, I applies all other suggestions since I also agree with those suggestion considering how we want to features to expressed in extension.

package.json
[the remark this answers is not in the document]
rgrunber: Ok, then let's keep it as is for now.

src/extension.ts
```diff
@@ -491,6 +491,16 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 				}
 			});
 
+			context.subscriptions.push(commands.registerCommand(Commands.CHANGE_JAVA_SEARCH_SCOPE, async () => {
+				const selection = await window.showQuickPick(["all", "main"], {
```
rgrunber, line 495: Only minor annoyance here is the quick-pick doesn't indicate what the current value is. However there is an overloaded version of `showQuickPick` where the list of input elements can be a list of things that extend `QuickPickItem`. One of its fields is `picked` which should be able to indicate whether it is selected.

rgrunber (commented):
Overall, I think this is ready to be merged once we figure out the LS situation.

src/extension.ts
gayanper: I tried this, but seems like it doesn't work unless you set the selection to be multi select. For single selection I couldn't find a way to set the initial selection into the selection nor get checkboxes to work.

package.json
```diff
@@ -1499,6 +1504,19 @@
...
+            ],
+            "enumDescriptions": [
+              "Search on all classpath entries including reference libraries and projects.",
+              "All classpath entries excluding test classpath entries."
+            ],
+            "default": "all",
+            "markdownDescription": "Specifies the scope which must be used for search operation like \n - Find Reference\n - Call hierarchy\n"
```
rgrunber, line 1519: Now also Workspace Symbols .

rgrunber (commented):
Under the README.md, can you add a section listing the new setting for the upcoming 1.35.0 ? See https://github.com/redhat-developer/vscode-java/commit/bf05afb36fb2ba9a5f6b271e07b805ac09ab57b6#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R252 for an example.

gayanper:
I will add during today @rgrunber

gayanper:
> Under the README.md, can you add a section (at the bottom of settings) listing the new setting for the upcoming 1.35.0 ? See [bf05afb#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R252](https://github.com/redhat-developer/vscode-java/commit/bf05afb36fb2ba9a5f6b271e07b805ac09ab57b6#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R252) for an example. You can reuse the `markdownDescription` from your setting.

@rgrunber Done

PR #3803, opened 2024-10-07
Automatically add /// markdown prefix on next line

### LINKED ISSUE #3801: Automatically add /// on new line when editing Markdown javadoc
When editing Markdown javadoc, it should behave like in Eclipse and automatically add /// on new lines
Oct-07-2024 10-28-44

fbricon:
Fixes #3801 
Oct-07-2024 11-00-03

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -803,6 +803,11 @@ function enableJavadocSymbols() {
 				// e.g.  *-----*/|
 				beforeText: /^(\t|(\ \ ))*\ \*[^/]*\*\/\s*$/,
 				action: { indentAction: IndentAction.None, removeText: 1 }
+			},
+			{
+				// e.g. /// ...| (Markdown javadoc)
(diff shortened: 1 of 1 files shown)


mickaelistria:
Isn't the java/languageConfiguration.json file used for such cases? Having it in this file could allow to spread value wider as the file can be reused in other IDEs/editors (while the extension.ts cannot)

rgrunber (approved):
Apparently prior to this change, the language server's `textDocument/onTypeFormatting` would cause this :

https://github.com/user-attachments/assets/02e2e6df-6a47-4423-af98-2c19c2d1a0c3

No longer happen with this change, but no idea why it would try to do that in the first place.

rgrunber:
It only occurs when there's no content immediately above/below (only newlines), and typing <kbd>///</kbd> followed by <kbd>enter</kbd>. It doesn't happen in lightweight mode which also confirmed it's the standard server.

PR #3804, opened 2024-10-07
Remove Dependency Analytics extension from being recommended.

rgrunber:
- redhat.fabric8-analytics


src/extension.ts
```diff
@@ -149,12 +147,8 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 	}
 	initializeLogFile(clientLogFile);
 
-	const telemetryService: Promise<TelemetryService> = Telemetry.startTelemetry(context);
```
fbricon, line 152: so nothing starts the telemetry then?

rgrunber: Noticed this as well :neutral_face: , fixed a little while ago https://github.com/redhat-developer/vscode-java/commit/50ed9ee39f9bcd8ffafbdd86569e2d6f9d699a31 .

PR #3809, opened 2024-10-10
Use Node 20 in release-job only.

rgrunber:
- ovsx 0.10.0 requires Node >= 20

```
$ nvm use 18
$ npm install -g ovsx
$ ovsx --version
ovsx requires at least NodeJS version 20. Check your installed version with `node --version`.
```


datho7561 (approved):
Looks good

`ovsx` is also installed as part of the packaging step, where nodejs 18 is used, but `ovsx` is not used during that step, so it shouldn't cause a build error.

PR #3812, opened 2024-10-13
Create devcontainer.json

Rjonesout:


rgrunber:
This looks exactly the same as https://github.com/redhat-developer/vscode-java/pull/2992. Does anything else need to be added to make sure this works ? (eg. See https://github.com/redhat-developer/vscode-java/pull/2992#issuecomment-1470196693 )

PR #3845, opened 2024-11-06
Improve code actions

snjeza:
Fixes https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3321
Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3322


rgrunber (approved):
This part looks fine to me. I was hesitant about informing about the error, because we don't know when it might happen, and it could be annoying, but it's better than having the UI mysteriously not come up.

PR #3846, opened 2024-11-07
Create devcontainer.json

Ramzifayyaz3161:


rgrunber:
Is there a particular use-case for this or is this some kind of auto-generated contribution ? Seems same as https://github.com/redhat-developer/vscode-java/pull/3812 .

fbricon:
WIthout feedback from OP, I consider this like spam. closing.

PR #3856, opened 2024-11-13
Add 'removeUnusedLambdaParameters' to the cleanup settings.

rgrunber:
- Based on https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3319

Calling it lambda parameters seems a little odd, given it also supports unused pattern variables but I think we can live with this for now. Calling it "unusedLocalVariable" based on the naming in JDT seems a bit too confusing given that it doesn't rename all types of unused local variables.


iwangbowen:
Great feature

PR #3877, opened 2024-11-26
Lombok agent doesn't need to be configured if javac is enabled

snjeza:


src/lombokSupport.ts
```diff
@@ -91,6 +91,11 @@ export function addLombokParam(context: ExtensionContext, params: string[]) {
 	for (let i = deleteIndex.length - 1; i >= 0; i--) {
 		params.splice(deleteIndex[i], 1);
 	}
+	// https://github.com/redhat-developer/vscode-java/issues/3875
+	// Lombok agent doesn't need to be configured if java.jdt.ls.javac.enabled is on
+	if ("on" === vscode.workspace.getConfiguration().get("java.jdt.ls.javac.enabled")) {
```
testforstephen, line 96: The flag `"java.jdt.ls.javac.enabled"` only controls whether to delegate compilation and DOM AST to javac, which does not need an additional lombok agent. However, if code completion continues to rely on the ECJ-based engine, the Lombok agent will still be required.

Therefore, the lombok agent is not needed only when both `"java.jdt.ls.javac.enabled"` and `"java.completion.engine": "dom"` settings are enabled.

rgrunber: What about [`SourceIndexer.DOM_BASED_INDEXER`](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/287ae94f04ad850483a375a3e7ee90a409b9852b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java#L100) ? Lombok is affect by source references right ?

snjeza: > Therefore, the lombok agent is not needed only when both "java.jdt.ls.javac.enabled" and "java.completion.engine": "dom" settings are enabled.
> What about [SourceIndexer.DOM_BASED_INDEXER](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/287ae94f04ad850483a375a3e7ee90a409b9852b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java#L100) ? Lombok is affect by source references right ?

Fixed.

rgrunber: The issue is we don't even set the dom based indexer at https://github.com/redhat-developer/vscode-java/blob/2f57115288bba92124a96dc399e2af04bbdc0827/src/javaServerStarter.ts#L143-L145 so I meant we need a separate setting for it as well.

Issue #3900, opened 2024-12-19
bug - javac can not navigate to method [pre-release version]


mamilic:
When trying to navigate to method, or just trying to open methods docs nothing happens, however in logs I can see this error.

<details><summary>Details</summary>
<p>

```
[Error - 9:27:19 PM] Request textDocument/codeAction failed.
  Message: Internal error.
  Code: -32603 
java.util.concurrent.CompletionException: java.lang.IllegalStateException: java.lang.NoSuchFieldError: Class com.sun.tools.javac.parser.Tokens$Comment$CommentStyle does not have member field 'com.sun.tools.javac.parser.Tokens$Comment$CommentStyle JAVADOC'
	at java.base/java.util.concurrent.CompletableFuture.wrapInCompletionException(CompletableFuture.java:323)
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:359)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:364)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:693)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:527)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1458)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:2034)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:189)
Caused by: java.lang.IllegalStateException: java.lang.NoSuchFieldError: Class com.sun.tools.javac.parser.Tokens$Comment$CommentStyle does not have member field 'com.sun.tools.javac.parser.Tokens$Comment$CommentStyle JAVADOC'
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.analyze(JavacTaskImpl.java:383)
	at org.eclipse.jdt.core.dom.JavacCompilationUnitResolver.parse(JavacCompilationUnitResolver.java:722)
	at org.eclipse.jdt.core.dom.JavacCompilationUnitResolver.toCompilationUnit(JavacCompilationUnitResolver.java:545)
[...]

Issue #3897, opened 2024-12-19
Compile bug with javac


XuPengfei-1020:
I get this error after I set javac as compiler.
And compile can not be finished correctly.

{
  message: '[Error - 3:02:24 PM] Dec 19, 2024, 3:02:24 PM Failed to initialize the custom compiler factory - org.eclipse.jdt.internal.javac.JavacCompilerFactory\n' +
    'org.eclipse.jdt.internal.javac.JavacCompilerFactory cannot be found by org.eclipse.jdt.core_3.40.0.z20241127-1618\n' +
    'java.lang.ClassNotFoundException: org.eclipse.jdt.internal.javac.JavacCompilerFactory cannot be found by org.eclipse.jdt.core_3.40.0.z20241127-1618\n' +
    '\tat org.eclipse.osgi.internal.loader.BundleLoader.generateException(BundleLoader.java:562)\n' +
    '\tat org.eclipse.osgi.internal.loader.BundleLoader.findClass0(BundleLoader.java:557)\n' +
    '\tat org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:434)\n' +
    '\tat org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:174)\n' +
    '\tat java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:528)\n' +
    '\tat java.base/java.lang.Class.forName0(Native Method)\n' +
    '\tat java.base/java.lang.Class.forName(Class.java:462)\n' +
    '\tat java.base/java.lang.Class.forName(Class.java:453)\n' +
    '\tat org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.newCompiler(AbstractImageBuilder.java:612)\n' +
    '\tat org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.<init>(AbstractImageBuilder.java:134)\n' +
    '\tat org.eclipse.jdt.internal.core.builder.BatchImageBuilder.<init>(BatchImageBuilder.java:86)\n' +
    '\tat org.eclipse.jdt.internal.core.builder.JavaBuilder.buildAll(JavaBuilder.java:313)\n' +
    '\tat org.eclipse.jdt.internal.core.builder.JavaBuilder.build(JavaBuilder.java:220)\n' +
    '\tat org.eclipse.core.internal.events.BuildManager$2.run(BuildManager.java:1077)\n' +
[...]

PR #3902, opened 2024-12-20
Open requested system packages for javac

snjeza:
The related issues/commits:
- https://github.com/redhat-developer/vscode-java/issues/3900
- https://github.com/redhat-developer/vscode-java/issues/3897
- https://github.com/redhat-developer/vscode-java/issues/3885
- https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/commit/bf3c8ce7abbad257be2bbb7cea58a5f1e51137e7


src/javaServerStarter.ts
```diff
@@ -135,6 +135,16 @@ function prepareParams(requirements: RequirementsData, workspacePath, context: E
...
+		'--add-opens',
+		'jdk.compiler/com.sun.tools.javac.jvm=ALL-UNNAMED',
+		'--add-opens',
+		'jdk.zipfs/jdk.nio.zipfs=ALL-UNNAMED',
+		'--add-opens',
+		'java.compiler/javax.tools=ALL-UNNAMED',
+		'--add-opens',
+		'java.base/java.nio.channels=ALL-UNNAMED',
+		'--add-opens',
+		'java.base/sun.nio.ch=ALL-UNNAMED',
```
mickaelistria, line 147: another one is freshly necessary: `--add-opens java.compiler/javax.tools=ALL-UNNAMED`. Note that the exact list is available in the org.eclipse.jdt.core.javac/META-INF/p2.inf` file and should be part of the .ini file generated with the JDT-LS product.
Ideally, instead of hardcoding/duplicating those here, they should be read from one of the upstream files.

mickaelistria:
Good to be merged ASAP IMO

rgrunber:
@snjeza , if you have time, I would also update https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/1d41ddee3b461fcd8b30eadb3d7dd6abf56814ed/pom.xml#L343 .

snjeza:
>  if you have time, I would also update https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/1d41ddee3b461fcd8b30eadb3d7dd6abf56814ed/pom.xml#L343 .

@rgrunber Please check https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3351

PR #4030, opened 2025-04-23
Update required Java for javac to 24

### LINKED ISSUE #4029: Javac-based compilation requires Java SDK 24 (mismatch with documentation)
When I enable the Javac-based compilation feature, the extension no longer starts up, stating that it requires Java 24 or later. The feature says it only needs 23 or newer which is what I'm running. Is it 23 or 24?

Image

Image

Image

snjeza:
Fixes #4029

### DIFF
--- a/package.json
+++ b/package.json
@@ -416,7 +416,7 @@
               "off"
             ],
             "default": "off",
-            "markdownDescription": "[Experimental] Specify whether to enable Javac-based compilation in the language server. Requires running this extension with Java 23",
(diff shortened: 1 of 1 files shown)


rgrunber (approved):
We just need to include these in https://github.com/redhat-developer/vscode-java/pull/3983 , https://github.com/redhat-developer/vscode-java/pull/3987/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 or update the PR when the bot proposes it.

rgrunber:
Just FYI, Eclipse is experiencing some major service disruptions and given that the core incubator builds reside within Jenkins, we likely will fail to build this for a little while.

PR #4159, opened 2025-09-16
Remove node_modules prefix from gulp commands

tsmaeder:
Use `gulp <command>` instead of `gulp ./node_modules.bin/gulp`. This allows to build on Windows and likely is more robust, anyway.


rgrunber (approved):
Seems to follow from https://docs.npmjs.com/cli/v10/using-npm/scripts#path .

PR #4172, opened 2025-09-23
Update release.yml Java 21

mozhuanzuojing:
set java version = 25


rgrunber:
Based on a comment, I realized that we really should be able to use JDK 21 for compilation still. Let me verify this.

PR #4181, opened 2025-09-28
Enable formatting java in jupyter notebook

snjeza:
Fixes #4163 
Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3548
cc @rgrunber @fbricon @maxandersen @mickaelistria

PR #4187, opened 2025-09-30
fix: enable workspace/executeCommand in the Syntax language client

fbricon:
Follow up on https://github.com/redhat-developer/vscode-java/pull/4181#issuecomment-3350927276

### DIFF
--- a/src/syntaxLanguageClient.ts
+++ b/src/syntaxLanguageClient.ts
@@ -8,13 +8,14 @@ import { ClientErrorHandler } from "./clientErrorHandler";
 import { ClientStatus } from "./extension.api";
 import { logger } from "./log";
 import { OutputInfoCollector } from "./outputInfoCollector";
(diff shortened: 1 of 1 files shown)


rgrunber:
Do you know what the command was ? I see we pass the same https://github.com/redhat-developer/vscode-java/blob/813f9e9f42ead472215085eed21798197e2bace5/src/extension.ts#L235 to the syntax & standard servers, so I guess it should be supported.

The only ones I see called server-side are for bundle reloading (which I don't think is valid for syntax server), and `java.action.organizeImports.chooseImports`.

fbricon:
It's caused by the new notebook support in https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3548/files#diff-1b0c46dbd4816a8643a902345c0f91f26497e55525a9642c2b8a2ca06423304bR101

PR #4188, opened 2025-10-01
fix: forward _all_ LSP telemetry events

fbricon:
Currently only "java.workspace.initialized" and "java.ls.error" events are sent. 

Either we keep whitelisting events, but then new stuff like potentially https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3544 need to add new events to the whitelist, or we trust all JDT.LS events to be forwarded to our telemetry backend.

### DIFF
--- a/src/standardLanguageClient.ts
+++ b/src/standardLanguageClient.ts
@@ -360,6 +360,7 @@ export class StandardLanguageClient {
 					return Telemetry.sendTelemetry(Telemetry.LS_ERROR, e.properties);
 				}
 			}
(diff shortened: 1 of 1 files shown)


fbricon:
After thinking about it, I think the downside of this PR is it would include potential crap a rogue jdt.ls extension could send.

fbricon:
OK, after discussing with @rgrunber, we decided it's safer for vscode-java to explicitly control what goes in our telemetry pipeline.

Issue #4114, opened 2025-07-14
Opening a .class file on a remote window doesn't show decompiled source


apimentasoto:
The custom editor for java class files is not used when opening .class files on a remote window.

I use the remote development extension for WSL.

The glob used as selector for the editor won't match the scheme of remote files.

```
"selector": [
          {
            "filenamePattern": "file:/**/*.class"
          }
]
```

Got it working adding a pattern which matches remote files:
```
"selector": [
                    {
                        "filenamePattern": "file:/**/*.class"
                    },
                    {
                        "filenamePattern": "vscode-remote:/**/*.class"
                    }
]
```
I'm new to remote development so I have no idea if I should be seeing this scheme in the first place. In any case would be nice to have .class files open the decompiled source by default - currently there is no error message or any indication that something's missing - just the 'cannot open binary file' page.

nkilders:
I was facing the same problem, but don't completely understand your solution. What's the full name of the setting you use? Here is an alternative solution which works for me. Both patterns seem to be fine.

```json5
{
	"workbench.editorAssociations": {
		"*.class": "decompiled.javaClass",
		// or
		"vscode-remote:/**/*.class": "decompiled.javaClass",
	},
	...
}
```

Edit: the `*.class` pattern causes problems when using `Ctrl`+click to jump into class files. It does not focus the correct line (e.g. of the method you clicked), but always the beginning of the file. The other pattern works without this problem.

apimentasoto:
Oh I was refering to the selectors configured by the extension when the editor is contributed. Changing those would make the editor open class files on remote windows out of the box.
[Here](https://github.com/redhat-developer/vscode-java/blob/c0a956530660f16576dcc150fa80966d27d96564/package.json#L65)

I didn't even think about using user settings to fix this but it makes a lot of sense if the package.json of the extension is not updated.

nkilders:
Ahh, now I get it. Yes, your solution would be much better, as it would work without any configuration effort on the user's side. I have created a PR which adds the new filename pattern to the `package.json` file. Let's see what happens 😊

PR #4189, opened 2025-10-01
Fix decompiled.javaClass editor in WSL

nkilders:
This PR fixes https://github.com/redhat-developer/vscode-java/issues/4114

## Problem

When using the extension in Windows Subsystem for Linux (WSL), the `decompiled.javaClass` editor does not work properly. Opening a `.class` file results in an error text, "The file is not displayed in the text editor because it is either binary or uses an unsupported text encoding."

<details>
<summary>Screenshot</summary>

Opening .class file without fix

</details>

Clicking "Open Anyway" opens the raw bytecode rather than the decompiled sources:

<details>
<summary>Screenshot</summary>

'Open Anyway' before fix

</details>

When right-clicking a `.class` file and selecting "Open With...", the `decompiled.javaClass` editor is not listed:

<details>
<summary>Screenshot</summary>

'Open With...' dialog without fix

</details>

## Solution

By adding a new filename pattern to the `package.json` file, the issue is gone. Opening a `.class` file directly shows the decompiled code, and the "Open With..." dialog also lists the `decompiled.javaClass` editor.

<details>
<summary>Screenshots</summary>

Opening .class file with fix

'Open With...' dialog with fix

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,9 @@
         "selector": [
           {
             "filenamePattern": "file:/**/*.class"
+          },
+          {
+            "filenamePattern": "vscode-remote:/**/*.class"
           }
         ]
       }


rgrunber:
Seems pretty reasonable to me, especially when it just works :stuck_out_tongue_winking_eye:  I'm guessing all the URIs are `vscode-remote://` when dealing with WSL. @fbricon let me know if you see any issues with this.

fbricon:
thanks @nkilders!

PR #4198, opened 2025-10-06
Update flags for jdt-javac

datho7561:
Update the flags to match https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/blob/dom-with-javac/org.eclipse.jdt.core.javac/META-INF/p2.inf (I've reordered some flags to match that file).

This opens a bunch of jdk internals that the jdt-javac work uses, and also sets the flags to enable the javac-based search.


datho7561:
I need to double check that I've got all the correct flags still before this is ready for review. Hopefully I'll get that done first thing tomorrow

mickaelistria:
Is completion flag `-DICompletionEngineProvider=org.eclipse.jdt.core.dom.DOMCompletionEngineProvider` intentionally skipped?

datho7561:
Yes, there's a separate setting to enable it: https://github.com/redhat-developer/vscode-java/blob/main/src/javaServerStarter.ts#L165

Issue #4226, opened 2025-10-31
Java JDK auto-detection problem


LeducNic:
If I have JDK 25 installed on my system, but the java.jdt.ls.java.home is not set, then the tooling JRE is set to the embedded JDK 21 even if it doesn't meet the requirement (JDK 25) and I have a newer version on my system.  Then, I propose the following change to the requirements.ts, so that the toolingJRE is reset if it doesn't meet the requirement forcing the method to find another tooling JRE.  The second modification is there to keep the javaHome set by the user even if a newer version of the JDK is found for the toolingJRE.  So, I can set javaHome to JDK 17 and the tooling will use a JDK (embedded or elsewhere on my system) that meet the requirement.

I am not setup to do a commit and a PR, so I include the proposed corrections as a diff below.

```
*** ORIGINAL requirements.ts"	
--- NEW requirements.ts
***************
*** 42,47 ****
--- 42,51 ----
          let javaHome = javaPreferences.javaHome;
          let javaVersion: number = 0;
          const REQUIRED_JDK_VERSION = ('on' === getJavaConfiguration().get('jdt.ls.javac.enabled'))?25:21;
+         if (toolingJreVersion < REQUIRED_JDK_VERSION) { // embedded tooling JRE doesn't meet requirement
+             toolingJre = null;
+             toolingJreVersion = 0;
+         }
          if (javaHome) {
              const source = `${preferenceName} variable defined in ${env.appName} settings`;
              javaHome = expandHomeDir(javaHome);
***************
*** 79,88 ****
              const validJdks = javaRuntimes.filter(r => r.version.major >= REQUIRED_JDK_VERSION);
              if (validJdks.length > 0) {
                  sortJdksBySource(validJdks);
!                 javaHome = validJdks[0].homedir;
!                 javaVersion = validJdks[0].version.major;
!                 toolingJre = javaHome;
!                 toolingJreVersion = javaVersion;
              }
          } else { // pick a default project JDK/JRE
              /**
--- 83,94 ----
              const validJdks = javaRuntimes.filter(r => r.version.major >= REQUIRED_JDK_VERSION);
              if (validJdks.length > 0) {
                  sortJdksBySource(validJdks);
!                 toolingJre = validJdks[0].homedir;
!                 toolingJreVersion = validJdks[0].version.major;
!                 if (!javaHome) { // keep javaHome if set even if it is an older JDK version
!                     javaHome = toolingJre;
!                     javaVersion = toolingJreVersion;
[...]

PR #4235, opened 2025-11-06
Fix: auto-detection of JDK (#4226)

LeducNic:
Use the JDK found with jdk-utils if the embedded JDK does not meet the requirement

Fixes https://github.com/redhat-developer/vscode-java/issues/4226


datho7561:
(going to let the CI jobs run before merging just in case they catch something)

Issue #2765, opened 2022-10-28
TypeError: Cannot read properties of null (reading '0')


CsCherrYY:
when opening spring-petclinic project with lombok dependency in the classpath, we can get this error in the console output:
```
rejected promise not handled within 1 second: TypeError: Cannot read properties of null (reading '0')
c:\Users\chenshi\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:94
stack trace: TypeError: Cannot read properties of null (reading '0')
	at E (c:\Users\chenshi\Documents\GitHub\vscode-java\dist\extension.js:2:727247)
	at Object.getLombokVersion (c:\Users\chenshi\Documents\GitHub\vscode-java\dist\extension.js:2:728652)
	at Object.initializeLombokStatusBar (c:\Users\chenshi\Documents\GitHub\vscode-java\dist\extension.js:2:767391)
	at Object.<anonymous> (c:\Users\chenshi\Documents\GitHub\vscode-java\dist\extension.js:2:729598)
	at Generator.next (<anonymous>)
	at o (c:\Users\chenshi\Documents\GitHub\vscode-java\dist\extension.js:2:725923)
```

looks like the null check is needed in function `lombokPath2Version()`.

PR #4247, opened 2025-11-10
fix: parsing lombok version error

fbricon:
Fixes https://github.com/redhat-developer/vscode-java/issues/2765 if the lombok jar doesn't match the expectd pattern.

Screenshot 2025-11-10 at 18 42 42
Screenshot 2025-11-10 at 18 43 01

I tested by renaming the embedded jar to lombok.jar

### DIFF
--- a/src/lombokSupport.ts
+++ b/src/lombokSupport.ts
@@ -67,22 +67,30 @@ function getExtensionLombokPath(context: ExtensionContext): string {
 	return path.join(context.asAbsolutePath("lombok"), files[0]);
 }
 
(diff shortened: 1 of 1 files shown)


datho7561:
I think I don't know enough about lombok to review this well. I renamed the bundled lombok(s) to `lombok.jar` but don't get any errors when launching vscode-java and opening a project that uses lombok.

Fred, what's the dialog that you have open there? I can't seem to get it to appear.

fbricon:
You need to click on `{} Java` at the bottom when a Java file is open, then click `Configure Lombok`

PR #4249, opened 2025-11-12
add JSpecify annotations for null analysis

martinlippert:
Due to JSpecify emerging as the new standard for null analysis annotations, here is the corresponding config for the default settings. The corresponding PR for the server-side changes to detect JSpecify will follow shortly.


martinlippert:
Thanks so much @fbricon for merging this so quickly. Very much appreciated.

PR #4269, opened 2025-11-24
Remove title for package.json schema contribution

### LINKED ISSUE #4268: Do not define title in package.json schema as it is shown in tooltip for all package.json files
See https://github.com/redhat-developer/vscode-xml/issues/1117

datho7561:
This title will compete with the title of the base schema in cases like highlight, and the title is nonsensical for package.json for things other than VS Code extensions.

Fixes #4268


fbricon:
Thanks @datho7561!

PR #4304, opened 2026-01-11
Support AspectJ in Gradle projects

snjeza:
See https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3670
Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3671

cc @fbricon @rgrunber @tsmaeder


fbricon:
So when toggling the setting, nothing happens, on restart nothing happens, you really need to trigger a workspace clean for the errors to disappear. We could automatically do it when the setting value changes instead.

snjeza:
> So when toggling the setting, nothing happens, on restart nothing happens, you really need to trigger a workspace clean for the errors to disappear. We could automatically do it when the setting value changes instead.

@fbricon I have created https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3675

fbricon:
Thanks @snjeza!

Issue #4305, opened 2026-01-13
Bug: Broken Paste logic


CookedMelon:
Using this extension causes issues with the copy function in VS Code.

**Step to Reproduce:**

Create a file like `test.java`, the content is:

```
class test {
    private String function () {
        String a = "555\n555";
        a = "ggg";
        return a;
    }
}
```

**Scenario A (Selection Paste):** Use the mouse to select the \n characters (inside the double quotes), copy them, and attempt to paste them.

**Scenario B (Line Copy/Paste):** Place the cursor anywhere inside the double quotes (on the line String a = "\n";) without selecting any text. Use the VS Code shortcut Ctrl + C then Ctrl + V (intended to duplicate the entire line).

**Expected Behavior:**

**Scenario A**

Before Paste

<img width="386" height="196" alt="Image" src="https://github.com/user-attachments/assets/a64f7c9e-8541-438b-b56b-ffc9bb7eed85" />

After Paste

<img width="428" height="205" alt="Image" src="https://github.com/user-attachments/assets/9e98330a-838e-4dfc-8add-169e29ecf38f" />

**Scenario B**

Before Paste

<img width="444" height="242" alt="Image" src="https://github.com/user-attachments/assets/075ddb52-a39c-40df-9af7-796d91ab873b" />

After Paste

<img width="477" height="266" alt="Image" src="https://github.com/user-attachments/assets/763cfb45-43c4-46f4-b216-4026bd7e8679" />

**Actual Behavior:**

**Scenario A**

<img width="486" height="284" alt="Image" src="https://github.com/user-attachments/assets/8f46a593-31ec-4fa6-ab64-5a966e653886" />

**Scenario B**

If I place the cursor inside the double quotes

<img width="506" height="259" alt="Image" src="https://github.com/user-attachments/assets/b89b226b-8acd-42a4-94bf-8834b950c7a5" />

I press ctrl c & ctrl v

<img width="561" height="271" alt="Image" src="https://github.com/user-attachments/assets/48848934-5315-4d1d-a26e-059ac7061b14" />

fbricon:
The paste logic mostly does what it's supposed to do: escape the content before pasting it into a Java String (see https://github.com/redhat-developer/vscode-java/issues/1249). So it will literally escape all new lines from the clipboard to insert into a compilable java String.
What you're expecting is raw pasting, you can choose by either Ctrl+Shift+P, or selecting "Insert Plain Text"

[...]

PR #4306, opened 2026-01-13
Fix paste handling for single line copies

CookedMelon:
A description of this issue can be found at this link https://github.com/redhat-developer/vscode-java/issues/4305


CookedMelon:
Thanks for the feedback, @fbricon. I have updated the logic and remove the redundant quote-checking block as suggested.

fbricon:
thanks @CookedMelon

PR #4313, opened 2026-01-21
Support Kotlin,Groovy in Gradle projects

### LINKED ISSUE #4312: Support Kotlin,Groovy in Gradle projects

snjeza:
Fixes #4312 
Requires https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3682

cc @fbricon @rgrunber @tsmaeder


fbricon:
It should be advertised as "partial" support, since we don't provide completion, or pretty much any LSP smartness besides making sure the other java classes compile.

We should probably enable those by default.

snjeza:
> It should be advertised as "partial" support, since we don't provide completion, or pretty much any LSP smartness besides making sure the other java classes compile.

I have marked the properties as Experimental.

> We should probably enable those by default.

Done.

fbricon:
@snjeza thanks!

PR #4330, opened 2026-02-03
build: avoid calling webpack for each platform build

fbricon:
That should speed up the release process, as webpack will be called once instead of 7 times (6 platforms + 1 universal)


fbricon:
build is 2-3 minutes faster

Issue #4320, opened 2026-01-29
Create file or method from quick fix doesn't work.


yuyully:
##### Environment
- Operating System: Windows
- JDK version: 21.0.8
- Visual Studio Code version: 1.108.2
- Java extension version: 1.52.0

##### Steps To Reproduce
1. Open a java project > New a java file
2. Set "editor.pasteAs.enabled": false
3. Copy `File f = new File("demo.txt");` into the java file > Create file from quick fix > It doesn't work
4. Copy `Foo f = new Foo();		f.call();` into the java file > Create method from quick fix > It doesn't work

##### Current Result

<img width="858" height="410" alt="Image" src="https://github.com/user-attachments/assets/e2ad1f11-5fab-4812-8ca2-67d13ad82ff2" />

<img width="558" height="331" alt="Image" src="https://github.com/user-attachments/assets/2d1ef81a-0c88-4eb4-8b8a-ee1a5337fcf8" />

fbricon:
I can reproduce. I have a hunch this might have been caused by https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3650. Need to investigate further.

fbricon:
This is what codeAction/resolve returns in 1.52, doesn't contain any text edits:
```
[Trace - 10:06:38 AM] Received response 'codeAction/resolve - (575)' in 1ms.
Result: {
    "title": "Create class 'File'",
    "kind": "quickfix",
    "diagnostics": [
        {
            "range": {
                "start": {
                    "line": 5,
                    "character": 2
                },
                "end": {
                    "line": 5,
                    "character": 6
                }
            },
            "severity": 1,
            "code": "16777218",
            "source": "Java",
            "message": "File cannot be resolved to a type",
            "data": {
                "arguments": [
                    "File"
                ]
            }
        }
    ],
    "command": {
        "title": "refresh Diagnostics",
        "command": "java.project.refreshDiagnostics",
        "arguments": [
            "file:///Users/fbricon/Dev/souk/quickfix-bug/src/main/java/Apple.java",
            "thisFile",
            false,
            true
        ]
    }
}
```

This is what 1.51.0 returns: 

```
[Trace - 10:11:39 AM] Received response 'codeAction/resolve - (35)' in 18ms.
Result: {
    "title": "Create class 'File'",
    "kind": "quickfix",
    "diagnostics": [
        {
            "range": {
                "start": {
                    "line": 5,
                    "character": 2
                },
[...]

PR #4347, opened 2026-02-24
Changelog for 1.53.0

fbricon:


datho7561:
~~Maybe I'm missing something but I don't see https://github.com/redhat-developer/vscode-java/issues/4320 in the changelog~~ yep that was in last release my bad

PR #4367, opened 2026-03-19
Add JDK 26

snjeza:
cc @fbricon @rgrunber @datho7561


fbricon:
Thanks @snjeza!

fbricon:
I merged too quickly, we need to update the checks for the javac plugin, which now needs to run with 26  when enabled

PR #4368, opened 2026-03-19
Check JDK for javac

snjeza:
cc @fbricon

### DIFF
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -2,13 +2,13 @@
 
 import * as expandHomeDir from 'expand-home-dir';
 import * as fse from 'fs-extra';
-import { getRuntime, getSources, JAVAC_FILENAME, JAVA_FILENAME } from 'jdk-utils';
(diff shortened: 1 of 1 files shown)


fbricon:
so it turns out we don't actually need to enforce a jdk 26 requirement for the javac plugin, it's only required if the user needs to compile against java 26. Which will most likely not be the case for a majority of users. I think we should keep this PR on the backburner for now.

src/requirements.ts
```diff
@@ -41,7 +41,7 @@ export async function resolveRequirements(context: ExtensionContext): Promise<Re
         const preferenceName = javaPreferences.preference;
         let javaHome = javaPreferences.javaHome;
         let javaVersion: number = 0;
-        const REQUIRED_JDK_VERSION = ('on' === getJavaConfiguration().get('jdt.ls.javac.enabled'))?25:21;
```
jglick, line 44: By the way https://github.com/redhat-developer/vscode-java/wiki/Javac%E2%80%90based-compilation-support currently says 24 is required; should that say 25?

Issue #4182, opened 2025-09-29
Java version support is quite misleading


mapio:
Here you say that this extension "Supports code from Java 1.8 to Java 25". But since it is based on https://github.com/eclipse-jdtls/eclipse.jdt.ls if you read the details there, it says: "Supports compiling projects from Java 1.8 through 24".

So it is not clear if using Java 25 is safe now. The "code" is supported, but not "compiling" it? 

When using gradle configured for Java 25 (with the code obtained from `gradle init`) I can compile and execute tests, but the "Problems" panel is full of 
```
[{
"resource": ".../bugrep/app/src/main/java/org/example/App.java",
"owner": "_generated_diagnostic_collection_name_#5",
"code": "4195414",
"severity": 4,
"message": "Compiling for Java version '25' is not supported yet. Using '24' instead",
"source": "Java",
"startLineNumber": 1,
"startColumn": 1,
"endLineNumber": 1,
"endColumn": 2,
"origin": "extHost1"
}]
```

This is quite confusing.

rgrunber:
Yes, the 2 phrases are meant to imply the same thing. Though we should probably change "code" to "compiling" to make it clear we aren't restricting our support to just runtime. For Gradle, are you using version >= 9.1.0. For using certain versions of Java, you need to be using a version of Gradle that supports that. See https://docs.gradle.org/current/userguide/compatibility.html .

mapio:
I am using Gradle 9.1.0 and the code compiles. But unless I switch to the insider version and the pre-released extensions, in the problems panel every class reports `Compiling for Java version '25' is not supported yet. Using '24' instead`.

I don't know the reason, but it is for sure very confusing for a new user (I'm teaching to a second year undergraduate CS class and when students see an alarm and an error are usually blocked).

mapio:
To make it more clear, this is a fresh start with

`code . --profile "standard" --user-data-dir /tmp/BUGREP/profile-standard`

<img width="1186" height="613" alt="Image" src="https://github.com/user-attachments/assets/9a51f721-ba98-4fc6-9385-2b8194d5090b" />

where you can see the problems reported, and the following is obtained with the insiders version

`code-insiders . --profile "insiders" --user-data-dir /tmp/BUGREP/profile-insiders`

<img width="1207" height="627" alt="Image" src="https://github.com/user-attachments/assets/e6fd844e-a8af-45d5-8bd0-8c5e63501125" />

[...]

Issue #4018, opened 2025-04-04
Broken with OpenJDK 24 (gradle 9.x snapshots)


theoparis:
See https://github.com/gradle/gradle/pull/32597
```
[error] FAILURE: Build failed with an exception.

* What went wrong:
Resolution of the configuration ':brigadier:annotationProcessor' was attempted without an exclusive lock. This is unsafe and not allowed.
```

The project builds fine from the command line (with JAVA_HOME set to Java 24), so I assume it is an issue with this extension, or maybe eclipse jdtls.

PyvesB:
Facing a similar issue with Gradle 9.0.0 and Java 24, though in my case it's when resolving the `:grpc-utils:annotationProcessor` configuration.

einarjohnson:
did you ever figure this one out? having the same issue

PyvesB:
Worth noting that this is not specific to Java 24 nor to Gradle snapshots. The issue remains on Java 25 and 26, as well as stable Gradle 9 releases.

I managed to work around the issue by adding the following Gradle init file with Claude's help (`~/.gradle/init.d/ide-import-compat.init.gradle.kts`):
```
import org.gradle.api.internal.project.ProjectInternal

// Gradle 9.x requires an exclusive project lock to resolve any configuration.
// When the Gradle Build Server fetches the IDE model via a BuildAction, it resolves
// all resolvable configurations (annotationProcessor, checkstyle, compileClasspath, …)
// without holding individual sub-project locks, causing:
//   "Resolution of the configuration '...' was attempted without an exclusive lock."
//
// Fix: pre-resolve ALL resolvable configurations in gradle.projectsEvaluated, which fires
// after ALL afterEvaluate callbacks complete. We acquire each project's lock explicitly via
// applyToMutableState so hasMutableState() is true during resolution.
// DefaultConfiguration.resolveGraphIfRequired() then short-circuits on the cached result
// before the hasMutableState() check when the GBS model builder runs.
if (gradle.startParameter.taskNames.isEmpty()) {
    gradle.projectsEvaluated {
        rootProject.allprojects {
            val projectOwner = (this as ProjectInternal).owner
            configurations
                .filter { it.isCanBeResolved }
                .forEach { config ->
                    projectOwner.applyToMutableState { _ ->
                        try {
                            config.resolve()
                        } catch (_: Exception) {
                            // Ignore — best-effort pre-resolution; IDE import takes priority.
                        }
                    }
                }
        }
    }
}
```

PR #4383, opened 2026-04-01
Update bundled JRE to Java 25

### LINKED ISSUE #4353: update included JRE to 25
It would be awesome if the included JRE could be updated to a recent Java 25 runtime.

Background: the Spring Tools try to re-use the JDK that comes with `vscode-java` to avoid including its own JRE. Since we would like to start using specific JDK 25 features, I am wondering when `vscode-java` will update its internal JRE to 25... :-)

We could - of course - include and ship our own JRE, but due to the fact that we rely on `vscode-java` to be installed side by side anyway, we try to avoid putting another JRE install on the users machine. 😀

amritamishra01:
## Summary

Updates the bundled JustJ runtime from Java 21 to Java 25.

### Changes

* Updated default bundled JRE version in `scripts/jre.mjs`
* Updated release workflow packaging step to download Java 25 runtime
* Updated CLI usage example in `scripts/index.mjs`

## Validation

* Verified `LATEST_JRE` now points to 25
* Verified release workflow packages VSIX with `--javaVersion 25`
* Verified CLI usage example reflects Java 25

Fixes #4353

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -140,7 +140,7 @@ jobs:
           # Package for each platform (only downloads JDK and creates VSIX, no recompilation)
           platforms=("win32-x64" "win32-arm64" "linux-x64" "linux-arm64" "darwin-x64" "darwin-arm64")
           for platform in ${platforms[@]}; do
-            npm run download-jre -- --target ${platform} --javaVersion 21
+            npm run download-jre -- --target ${platform} --javaVersion 25
             SKIP_WEBPACK=true vsce package ${{ env.publishPreReleaseFlag }} --target ${platform} -o java-${platform}-${{ env.EXT_VERSION }}-${GITHUB_RUN_NUMBER}.vsix
           done
 
--- a/scripts/index.mjs
+++ b/scripts/index.mjs
@@ -113,7 +113,7 @@ Commands:
     repo-fix                            Fix package-lock.json registry references
 
 Examples:
-  node scripts/index.js download-jre --target darwin-x64 --javaVersion 21
+  node scripts/index.js download-jre --target darwin-x64 --javaVersion 25
   node scripts/index.js build-or-download
   node scripts/index.js watch-server
             `);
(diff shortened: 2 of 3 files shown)


wenytang-ms:
I gave this a spin on a real JDK 25 (Temurin 25.0.3) — launched the bundled server and imported actual projects over LSP. Mostly looks great, but I hit one thing worth calling out before release.

**Stuff that works fine on 25:**
- Server boots and the LSP handshake is clean, no class-version/OSGi errors.
- Maven imports/builds with 0 errors, both ECJ and `javac.enabled`. All the `jdk.compiler/*` add-opens are still valid on 25.
- Lombok 1.18.39 agent loads fine and generates accessors correctly.

**The catch — Gradle:**
Opening a Gradle 8.5 project blows up with `Unsupported class file major version 69` → *"Can't use Java 25.0.3 and Gradle 8.5"*. The embedded JRE is what Buildship uses to run Gradle unless you set `java.import.gradle.java.home`, so this hits the default setup. Gradle''s [matrix](https://docs.gradle.org/current/userguide/compatibility.html) says running on JDK 25 needs Gradle >= 9.1.0, whereas 21 was happy with 8.5 — so this bumps the floor a lot and breaks most Gradle 8.x projects. Basically the same thing as #4018.

So: totally fine to merge, the runtime''s ready on 25. But two things would be nice — a heads-up in the release notes that Gradle < 9.1.0 folks need to upgrade Gradle or point `java.import.gradle.java.home` at a JDK <= 21, and maybe sort out the "Java 25" wording from #4182 while we''re at it.

amritamishra01:
Hi @wenytang-ms , Thanks for taking the time to test this and for the detailed feedback! The Gradle compatibility note is a good catch. Since this is due to Gradle's current Java 25 support requirements rather than the runtime update itself, I agree that it would be helpful to mention it in the release notes or documentation.

XhstormR:
ping

Issue #4219, opened 2025-10-29
Very long Javadoc is truncated on hover


fbricon:
When hovering over "Pattern", jdt.ls returns a 122324-character hover, but the returned javadoc is truncated at 10000 characters by VS Code.

<img width="768" height="278" alt="Image" src="https://github.com/user-attachments/assets/0454b951-4283-47bd-9e32-56ae6e43d091" />

[pattern_hover_response.txt](https://github.com/user-attachments/files/23208677/pattern.txt).

In the past ([2020](https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/1481#issuecomment-643826234)), VS Code could render the Pattern Javadoc without any issues:

![](https://user-images.githubusercontent.com/148698/84604809-feb8b000-ae98-11ea-8d5b-a160357e97da.png)

1- we need to figure out if there's a way for vscode to cope with large hovers
2- the rendered markdown shows very large amounts of whitespace characters in use somehow, there might be some optimization  possible there.

datho7561:
This was fixed by eclipse-jdtls/eclipse.jdt.ls#3786

PR #4385, opened 2026-04-03
Reduce excessive whitespace in large Javadoc hover documentation

amritamishra01:
## Summary

This PR reduces excessive whitespace in large Javadoc hover and completion documentation.

Long Javadoc content such as `java.util.regex.Pattern` preserved repeated spaces, tabs, and excessive blank lines, which inflated the rendered markdown size and caused large hover popups to waste vertical space.

## Changes

* collapse repeated spaces/tabs into a single space
* collapse 3+ consecutive newlines into standard markdown paragraph spacing
* trim leading/trailing whitespace
* preserve existing `jdt://` link rewriting behavior

## Validation

Tested with `java.util.regex.Pattern` hover, especially the **Quotation**, **Boundary matchers**, and **Special constructs** sections that previously showed large whitespace gaps.

The hover now renders more compactly and allows significantly more regex constructs to be visible in the same viewport.

 fixes# #4219

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -125,8 +125,8 @@ export function fixJdtLinksInDocumentation(oldDocumentation: MarkdownString): Ma
 		return `${group1}${uri})`;
 	});
 
-	// Normalize excessive whitespace in large Javadoc content to help 
-	// large hovers render more completely within VS Code's limits.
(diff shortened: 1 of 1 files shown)


datho7561:
I think this should probably be addressed on the server side, so that all language servers can benefit from it.

PR #4412, opened 2026-05-11
fix(language-support): support generic records

its-miroma:
Notice line 14:

| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/language-support/java/java.tmLanguage.json
+++ b/language-support/java/java.tmLanguage.json
@@ -1448,16 +1448,14 @@
 							"name": "entity.name.type.record.java"
 						}
 					},
-					"end": "\\)",
-					"endCaptures": {
-						"0": {
-							"name": "punctuation.definition.parameters.end.bracket.round.java"
-						}
-					},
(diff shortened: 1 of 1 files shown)


wenytang-ms:
LGTM, better to add a unit test case

its-miroma:
Hello @wenytang-ms, I just got back to this!

I don't see any tests for the tmLanguage in the repo yet, do you know where I can find them?

datho7561 (approved):
Looks good; improves the syntax highlighting for this case before the semantic tokens kick in. There's no need for unit tests; I don't think there's a good way to do that and we haven't been from what I understand.

Thanks, Miroma!

its-miroma:
Thank you for merging!

> I don't think there's a good way to do that

There are definitely ways to add tests for a tmLanguage, which might be interesting to explore:
<https://github.com/PanAeon/vscode-tmgrammar-test>

For example, by using the `vscode-tmgrammar-snap` command, one can create snapshots of the grammar scopes for certain snippets of Java source code, and track them in the repo.

PR #4428, opened 2026-06-15
Add Copy Fully Qualified Name command

### LINKED ISSUE #374: Add feature: copy fully qualified name
From the user @mojo2012 logged on the debugger side: 

I often need to copy a classes fully qualified name, eg. to put into a spring.xml file.

Best would be to have it directly in the context menu when right-clicking on a class name.

MeherSru:


src/extension.ts
```diff
@@ -558,6 +558,33 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
...
+
+				const fullyQualifiedName = await commands.executeCommand<string>(
+					Commands.EXECUTE_WORKSPACE_COMMAND,
+					Commands.GET_FULLY_QUALIFIED_NAME,
+					JSON.stringify(params)
+				);
+
+				if (fullyQualifiedName) {
+					await env.clipboard.writeText(fullyQualifiedName);
+					window.showInformationMessage(`Copied: ${fullyQualifiedName}`);
```
datho7561: I think we should remove this information message before merging. We want to avoid overwhelming the user with information through the notifications.

test/standard-mode-suite/extension.test.ts
```diff
@@ -128,6 +128,9 @@ suite('Java Language Extension - Standard', () => {
 				Commands.FILESEXPLORER_ONPASTE,
 				Commands.RESOLVE_PASTED_TEXT,
 				Commands.CHANGE_JAVA_SEARCH_SCOPE,
+				Commands.COPY_FULLY_QUALIFIED_NAME,
+				Commands.GET_FULLY_QUALIFIED_NAME,
+				"java.getTroubleshootingInfo",
```
datho7561: What is `"java.getTroubleshootingInfo"`? What does it do? It seems unrelated to this PR.

datho7561 (changes requested):

test/standard-mode-suite/extension.test.ts
MeherSru: I initially pushed the change without adding `java.getTroubleshootingInfo`. When I ran the tests, the command registration test failed because `java.getTroubleshootingInfo` appeared in the registered command list but was missing from the expected list.

I then noticed that the command already exists as `Commands.GET_TROUBLESHOOTING_INFO`, so I updated the test to use the existing constant instead of a raw string for cleaner maintenance. Since the command wasn't introduced by this PR, this was just a small cleanup to keep the test consistent with the rest of the command definitions.

datho7561:
Fixes https://github.com/redhat-developer/vscode-java/issues/374

datho7561 (approved):
Looks good. Thanks, Meher!

PR #4436, opened 2026-06-25
Fix #3874: preserve escaped commas in snippet literals

### LINKED ISSUE #3874: Create local variable quick fix is broken for types with multiple generic arguments.
I had something like

```
x = new HashMap<Integer, Integer>();
```

This gives an error `x cannot be resolved to a variable`. This has a fix `Create local vartiable 'x'`. 

It used to become (not sure when it changed, but fairly recently)

```
HashMap<Integer, Integer> x = new HashMap<Integer, Integer>();
```

But now it becomes

```
HashMap<Integer x = new HashMap<Integer, Integer>();
```

With a drop down containing `HashMap<Integer` and `Integer>`.

MeherSru:
Fixes https://github.com/redhat-developer/vscode-java/issues/3874

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1305,6 +1305,6 @@ function registerRestartJavaLanguageServerCommand(context: ExtensionContext) {
 
 function escapeSnippetLiterals(value: string): string {
     return value
-        .replace(/\\/g, '\\\\')       // Escape backslashes
+        .replace(/\\(?!,)/g, '\\\\')       // Escape backslashes, but preserve escaped commas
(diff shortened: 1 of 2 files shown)


wenytang-ms (approved):
LGTM, it's better to add a unit test case.

MeherSru:
> LGTM, it's better to add a unit test case.

Thanks for the suggestion! I've added a unit test and pushed the changes.

wenytang-ms (approved):

datho7561 (approved):
Looks good and fixes the issue. Thanks, Meher!

PR #4447, opened 2026-06-30
fix: fall back to stdio when pipe startup fails

### LINKED ISSUE #4433: When using VSCode in portable mode with the tmp directory enabled, extensions fail to activate properly and get stuck in an activation state.
**Describe the bug**
On macOS, download vscode and enable portable mode. At the same time, create a temporary tmp file directory in the data directory, then open the vscode and Java project and install the Java plugin. The plugin will always be stuck in active mode, with no error messages or actions.

**To Reproduce**
1. Download the VSCode directory installed in portable mode, and create the tmp directory in the data directory of that mode,
2. Launch VSCode normally, install the Java plugin, and open the Java project. The plugin will show 'Activated' in the status bar
3. But nothing happened, and the project settings kept loading and couldn't be opened.
4. The final inspection results showed  Sock Unix domain path too long

** log ** 
!SESSION 2026-06-18 13:27:33.609 ----------------------------------------------- 
 eclipse.buildId=unknown 
 java.version=21.0.11 
 java.vendor=Eclipse Adoptium 
 BootLoader constants: OS=macosx, ARCH=aarch64, WS=cocoa, NL=zh_CN_#Hans 
 Framework arguments:  --pipe=/Users/renjinwei/work/software/JAVA/VScode/code-portable-data/tmp/lsp-f071153a9be153d587c38afa2047.sock 
 Command-line arguments:  -data /Users/renjinwei/work/software/JAVA/VScode/code-portable-data/user-data/User/workspaceStorage/f312aede2c1bb8af530fef927489ae6f/redhat.java/jdt_ws --pipe=/Users/renjinwei/work/software/JAVA/VScode/code-portable-data/tmp/lsp-f071153a9be153d587c38afa2047.sock 
 
 !ENTRY org.eclipse.jdt.ls.core 1 0 2026-06-18 13:27:34.543 
 !MESSAGE class org.eclipse.jdt.ls.core.internal.JavaLanguageServerPlugin is started 
 
 !ENTRY org.eclipse.jdt.ls.core 1 0 2026-06-18 13:27:34.564 
 !MESSAGE Started org.eclipse.buildship.core 15ms 
 
 !ENTRY org.eclipse.jdt.ls.core 1 0 2026-06-18 13:27:34.564 
 !MESSAGE Started org.eclipse.m2e.core 0ms 
 
 !ENTRY org.eclipse.jdt.ls.core 4 0 2026-06-18 13:27:34.569 
 !MESSAGE Unix domain path too long 
 !STACK 0 
 java.net.SocketException: Unix domain path too long 

No need,

**Expected behavior**
I hope it works properly. If the path is too long, you can use a shorter /tmp path or at least give an error message.

**Screenshots**

**Environment**
- Operating System: macos@26
- JDK version: jdk@25
- Visual Studio Code version: code@1.125.1
- Java extension version: 1.55.2026061909

**Additional Information**
none

chagong:
Fixes [#4433](https://github.com/redhat-developer/vscode-java/issues/4433).

This adds a guarded startup fallback for the standard language server: when the initial `TransportKind.pipe` executable does not connect within the 30-second pipe startup timeout, the extension recreates the client with `TransportKind.stdio` and starts again. This keeps `vscode-java` resilient while the underlying `vscode-jsonrpc` path-budget fix in [microsoft/vscode-languageserver-node#1804](https://github.com/microsoft/vscode-languageserver-node/pull/1804) is pending/released.

Implementation notes:
- The fallback is limited to executable server options using `TransportKind.pipe`.
- The fallback only handles pipe startup timeout; non-timeout startup failures still propagate normally.
- The fallback does not attempt to terminate the server process from this layer.
- File rename handlers resolve the current language client lazily so they use the replacement client after fallback.
- Added fake-client tests for pipe rejection, pipe timeout, and stdio failure behavior.

Validation:
- `npm run compile` passed, with the existing webpack warning from `vscode-languageserver-types`.
- `npm run eslint` passed.
- `git diff --check` passed.
- `npm test` ran the new `Standard Language Client Test`; all 3 new fallback tests passed. The full suite still fails on the pre-existing/unrelated command registration assertion where `java.getFullyQualifiedName` is missing from the actual registered commands in this local run.

### DIFF
--- a/src/fileEventHandler.ts
+++ b/src/fileEventHandler.ts
@@ -13,6 +13,7 @@ import * as stringInterpolate from 'fmtr';
 import { apiManager } from './apiManager';
 
 let serverReady: boolean = false;
+type LanguageClientProvider = LanguageClient | (() => LanguageClient);
 
 const BRACE_POSITION_KEY = "org.eclipse.jdt.core.formatter.brace_position_for_type_declaration";
 const END_OF_LINE = "end_of_line";
@@ -24,7 +25,7 @@ export function setServerStatus(ready: boolean) {
     serverReady = ready;
 }
 
-export function registerFileEventHandlers(client: LanguageClient, context: ExtensionContext) {
+export function registerFileEventHandlers(client: LanguageClientProvider, context: ExtensionContext) {
     if (workspace.onDidCreateFiles) {// Theia doesn't support workspace.onDidCreateFiles yet
         context.subscriptions.push(workspace.onDidCreateFiles(handleNewJavaFiles));
     }
@@ -188,7 +189,11 @@ async function handleNewJavaFiles(e: FileCreateEvent) {
     }, 100);
 }
 
-function getWillRenameHandler(client: LanguageClient) {
+function getLanguageClient(client: LanguageClientProvider): LanguageClient {
+    return typeof client === 'function' ? client() : client;
+}
+
+function getWillRenameHandler(client: LanguageClientProvider) {
     return function handleWillRenameFiles(e: FileWillRenameEvent): void {
         if (!serverReady) {
             return;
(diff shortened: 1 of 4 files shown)


datho7561:
@chagong It seems the fix has been merged in vscode-languageserver-node. Has it been released, and if so, should we prioritize upgrading to work around it? I have a work in progress PR to upgrade the language server client to the latest version.

datho7561 (approved):
I was able to reproduce the bug and can confirm that this patch fixes it. The code looks reasonable, too. Given this is a workaround for a bug that's fixed in a recent language client release, we may want to revert it after we've updated the language client.

Thanks, Changyong!

Issue #4417, opened 2026-05-24
Change Method Signature: parameter Type field doesn't visibly render typed text (contenteditable text node escapes shadow DOM slot)


benanavd:
### Environment

- OS: Windows 10/11 x64
- VS Code version: not captured
- Java extension version: 1.55.2026052308 (redhat.java)
- JDK: not captured

### Steps to reproduce

1. Trigger **Change Method Signature** on any method.
2. Click the **Type** cell of a parameter and type a new type.

### Result

The typed text does not appear in the **Type** field - the cell looks blank or unchanged.

Inspecting the DOM shows the text is captured, but as a direct text node on the `<vscode-data-grid-cell contenteditable="true">` host, sitting outside the shadow root, which contains only:

```html
<template shadowrootmode="open"><slot></slot></template>
```

The text is therefore present in the DOM but not visibly rendered. Example captured DOM:

```html
<vscode-data-grid-cell ... contenteditable="true">.</template>Stringdjdajljdlrfajedfla</vscode-data-grid-cell>
```

where `String` is the original type and the rest is typed input.

### Expected

Typed text should be visible in the **Type** cell while editing.

### Note

`vscode-data-grid-cell` is from the deprecated Webview UI Toolkit; the `contenteditable` text node is not being placed into a slotted/rendered position.

benanavd:
Adding the environment details that were missing from the original report, plus a tightened reproduction:

**Environment**
- OS: Windows 11 x64 (build 26200)
- VS Code version: 1.122.1 (user setup), commit 8761a5560cfd65fdd19ce7e2bd18dab5c0a4d84e, Chromium 142.0.7444.265, Electron 39.8.8
- Java extension version: 1.55.2026052308 (redhat.java)
- JDK: 21 (JavaSE-21)

**Steps to reproduce (any project, any method)**
1. Place the cursor on a method name, trigger the refactor, and choose **Change Method Signature**.
2. In the dialog, click the **Type** cell of an existing parameter to edit it.
3. Type any text (e.g. a new type name).

**Result**
- The typed text never appears in the cell — it looks blank/unchanged while editing.
- On clicking OK, the typed text is committed but **appended** to the original type rather than replacing it.

**Root cause (confirmed via DOM inspection)**
The edited text is captured as a bare text node placed directly on the `contenteditable` `<vscode-data-grid-cell>` host, *outside* its shadow root. The shadow root template is only:
```
<template shadowrootmode="open"><slot></slot></template>
[...]

PR #4448, opened 2026-06-30
Fix Change Method Signature editable cells not rendering typed text (#4417)

wenytang-ms:
### Problem
On **Change Method Signature**, editing a parameter **Type** (or Name / Default value, or an exception Type) shows nothing while typing, and on **OK** the typed text is *appended* to the original value instead of replacing it (e.g. `String` + `int` -> `Stringint`).

Root cause: these cells set `contentEditable` directly on the deprecated `@vscode/webview-ui-toolkit` `vscode-data-grid-cell`. That component's shadow root is only `<template shadowrootmode=""open""><slot></slot></template>`, so the text node typed onto the host sits outside any slot and is never rendered. On commit the value was read via `outerText`, which concatenates the original type with the stray text node — explaining both the invisible-while-typing and append-on-commit behavior. Reproduces with `--disable-gpu` (not a compositing issue).

### Fix
Render a real `<input>` as slotted content of the cell while editing, and read its `.value` on commit instead of `outerText`:
- Typed text is now visibly rendered.
- Commit uses correct replace semantics.
- `focus()`/`select()` moved into the `setState` callback (the input mounts after re-render); entering edit selects the existing text for quick replacement.
- Removed the now-obsolete `contentEditable` DOM-reset on cancel — component state is the single source of truth.

This is a minimal, dependency-free fix scoped to the editable cells. Fully migrating the dialog off the archived toolkit (it is the only webview still using it) can be tracked separately.

### Testing
- `npx tsc --noEmit` — passes
- `eslint` on the changed file — clean
- webpack `changeSignature` bundle — builds successfully
- Manual VS Code UI verification of the dialog still recommended by a reviewer.

### DIFF
--- a/src/webview/changeSignature/App.css
+++ b/src/webview/changeSignature/App.css
@@ -110,6 +110,19 @@ main {
 	background-color: var(--vscode-input-background);
 }
 
+.parameter-input {
+	box-sizing: border-box;
+	width: 100%;
+	height: 100%;
+	padding: 0;
+	border: 0;
+	outline: none;
+	background-color: transparent;
+	color: var(--vscode-input-foreground);
+	font-family: inherit;
+	font-size: inherit;
+}
+
 .parameter-cell-edit-button {
 	padding: 0;
 	border: 0;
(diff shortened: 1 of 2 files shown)


wenytang-ms:
current behavior

https://github.com/user-attachments/assets/6437aaf7-5fb6-4008-a972-1b478dfe21b4


fixed

https://github.com/user-attachments/assets/4d4a535c-89da-4106-bdc1-627c412d5004

PR #4450, opened 2026-06-30
Fix #4167: use java.projects context for explorer actions

### LINKED ISSUE #4167: Use Array Context Keys for More Precise Action Enablement
We could manage array valued contexts for various interesting resource sets, like "java projects" or "source folders" and use those to more precisely contribute actions to the explorer context menu. The code would look like: 

```
setContext('java.projects', ['path/To/Project1', '/path/To/Project']);
```

and then in our context actions:

```
when: resourcePath in java.projects
```

MeherSru:
This change introduces the java.projects array context and uses it to make the Add/Remove Source Path Explorer actions more precise. These actions are now only shown for Java project folders instead of every folder in the Explorer.

Fixes #4167

### DIFF
--- a/package.json
+++ b/package.json
@@ -1921,12 +1921,12 @@
           "group": "1_javaactions"
         },
         {
-          "when": "explorerResourceIsFolder&&javaLSReady",
+          "when": "explorerResourceIsFolder&&javaLSReady&&resourceDirname in java.projects",
           "command": "java.project.addToSourcePath.command",
           "group": "1_javaactions@1"
         },
         {
-          "when": "explorerResourceIsFolder&&javaLSReady",
+          "when": "explorerResourceIsFolder&&javaLSReady&&resourceDirname in java.projects",
           "command": "java.project.removeFromSourcePath.command",
           "group": "1_javaactions@2"
         }
--- a/package.json
+++ b/package.json
@@ -1921,12 +1921,12 @@
           "group": "1_javaactions"
         },
         {
-          "when": "explorerResourceIsFolder&&javaLSReady",
+          "when": "explorerResourceIsFolder&&javaLSReady&&resourcePath in java.projects",
           "command": "java.project.addToSourcePath.command",
           "group": "1_javaactions@1"
         },
         {
-          "when": "explorerResourceIsFolder&&javaLSReady",
+          "when": "explorerResourceIsFolder&&javaLSReady&&resourcePath in java.projects",
           "command": "java.project.removeFromSourcePath.command",
           "group": "1_javaactions@2"
         }
--- a/package.json
+++ b/package.json
@@ -1921,12 +1921,12 @@
           "group": "1_javaactions"
         },
         {
-          "when": "explorerResourceIsFolder&&javaLSReady&&resourcePath in java.projects",
+          "when": "explorerResourceIsFolder&&javaLSReady",
           "command": "java.project.addToSourcePath.command",
           "group": "1_javaactions@1"
         },
         {
-          "when": "explorerResourceIsFolder&&javaLSReady&&resourcePath in java.projects",
+          "when": "explorerResourceIsFolder&&javaLSReady",
           "command": "java.project.removeFromSourcePath.command",
           "group": "1_javaactions@2"
         }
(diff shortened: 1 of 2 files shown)


MeherSru:
This change introduces the java.projects array context and uses it to make the Add/Remove Source Path Explorer actions more precise. These actions are now only shown for Java project folders instead of every folder in the Explorer.

package.json
```diff
@@ -1914,12 +1914,12 @@
           "group": "1_javaactions"
         },
         {
-          "when": "explorerResourceIsFolder&&javaLSReady",
+          "when": "explorerResourceIsFolder&&javaLSReady&&resourcePath in java.projects",
```
wenytang-ms: `java.projects` only holds project *root* paths, and `in` is an exact-membership check (no prefix/descendant matching). So this only matches when you right-click the project root itself.

But Add/Remove Source Path are mostly used on **subfolders** (e.g. marking `src`/`test` as a source root) — those will now stop showing the action entirely, which is the opposite of what we want. Same applies to the `removeFromSourcePath` `when` just below.

For *remove*, what you probably want is a separate `java.sourceFolders` context (the actual source paths) and gate on `resourcePath in java.sourceFolders`. *Add* is trickier, since "folder is inside a project" can't really be expressed with exact membership.

src/standardLanguageClient.ts
```diff
@@ -152,6 +152,13 @@ export class StandardLanguageClient {
 					// Disable the client-side snippet provider since LS is ready.
 					snippetCompletionProvider.dispose();
 					registerDocumentValidationListener(context, this.languageClient);
+					try {
+						const projectUris = await getAllJavaProjects();
+						const projectPaths = projectUris.map((uriString) => Uri.parse(uriString).fsPath.replace(/[\\/]$/, ''));
+						await commands.executeCommand('setContext', 'java.projects', projectPaths);
```
wenytang-ms: This is only set once, on `ServiceReady`. Projects change over time (import/delete, classpath updates) and this never refreshes — so newly imported projects won't get the action, and deleted ones keep it until a reload.

Worth extracting a small helper and also calling it from `onDidProjectsImport` / `onDidProjectsDelete` (both already used in `runtimeStatusBarProvider.ts`).

src/standardLanguageClient.ts
```diff
@@ -152,6 +152,13 @@ export class StandardLanguageClient {
 					// Disable the client-side snippet provider since LS is ready.
 					snippetCompletionProvider.dispose();
 					registerDocumentValidationListener(context, this.languageClient);
+					try {
+						const projectUris = await getAllJavaProjects();
```
wenytang-ms: `await`ing this means `javaLSReady` (set a few lines down) now waits on a full LS round-trip — and gets blocked completely if this call ever hangs. Since `javaLSReady` gates a lot of menus/keybindings, I'd set it first and then populate `java.projects` fire-and-forget (it already has its own try/catch).

wenytang-ms (commented):
Thanks for tackling #4167! The array-context-key approach is the right tool here, but I think it's wired to the wrong set for these two actions. Left a few inline notes.
please also consider adding PR's description.

MeherSru:
@wenytang-ms Thanks for the detailed feedback! I've updated the PR to address the comments. I reverted the Explorer menu changes, moved the java.projects update out of the javaLSReady path, and added a helper to refresh the context when projects are imported or deleted.

datho7561 (changes requested):
The context is set up properly, but it seems like it's not actually used to adjust access to context menu items. The `when` clauses for adding and removing source folders will need to be adjusted. See https://code.visualstudio.com/api/references/when-clause-contexts . Take a look at package.json line 1917, where the when clause is configured

datho7561:
Ahh just noticed that changes to the package.json were removed after @wenytang-ms 's review. I think there's a way to configure it properly with how you've set it up

datho7561 (approved):
I think this is a good idea. The criteria for when it's shown isn't perfect (for example, this prevents you from adding nested folders to the source path), but I think it's still helpful. (In Eclipse projects that don't use Maven or Gradle, the source folders are usually located at the top level of the project, eg. `./src/`)

I'm going to figure out why the tests are failing before merging this.

datho7561:
The test failure is an existing, consistent test failure that I've documented here: https://github.com/redhat-developer/vscode-java/issues/4458.

I think we're good to merge, unless you want to take another look @wenytang-ms ?

PR #4459, opened 2026-07-09
Wait for project import before getting project settings

MeherSru:


datho7561:
I think that https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3844 fixed the test failure

PR #4462, opened 2026-07-13
Fix race condition when executing Java workspace commands during standard server startup

### LINKED ISSUE #2481: TypeError: Cannot read properties of undefined (reading 'sendRequest')
Below error randomly occurs:

```
[2022-05-30 10:00:24.949] [exthost] [error] TypeError: Cannot read properties of undefined (reading 'sendRequest')
	at c:\Users\admin\.vscode\extensions\redhat.java-1.6.6\dist\extension.js:2:698983
	at o._executeContributedCommand (c:\Users\admin\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:83:31333)
	at o._doExecuteCommand (c:\Users\admin\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:83:30153)
	at o._doExecuteCommand (c:\Users\admin\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:83:30868)
	at processTicksAndRejections (node:internal/process/task_queues:96:5) java.execute.workspaceCommand {"value":"redhat.java","_lower":"redhat.java"}
```

There are lots of places that `client.sendRequest` is called. NPE somewhere I guess.

shin19991207:
Fixes https://github.com/redhat-developer/vscode-java/issues/2481

There is a race condition in extension start-up where `java.execute.workspaceCommand` can be invoked while the standard Java language server is still starting. During that startup window, `standardClient.getClient()` can be `undefined`, but the command handler calls `standardClient.getClient().sendRequest(...)`. That can throw:
```
TypeError: Cannot read properties of undefined (reading 'sendRequest')
```

This is especially visible when standard server startup is delayed or skipped by startup decisions such as build tool conflict handling (i.e. open a project with both `build.gradle` and `pom.xml` existed).

This PR prevents `java.execute.workspaceCommand` from using the Java language client before it is initialized by:
- making `java.execute.workspaceCommand` wait for an active standard server startup before sending the workspace command request
- returning with a warning if the standard client is still unavailable after startup is skipped/cancelled, for example when the user dismisses the build tool conflict prompt
- reordering `postExtensionStartInit(...)` so the initial standard-server startup decision happens before commands that depend on the standard client are registered

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -52,6 +52,7 @@ const jdtEventEmitter = new EventEmitter<Uri>();
 const extensionName = 'Language Support for Java';
 let storagePath: string;
 let clientLogFile: string;
+let standardServerStart: Promise<LanguageClient | undefined> | undefined;
 
 const excutable=  new Deferred<Executable>();
 
@@ -444,6 +445,17 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 	return apiManager.getApiInstance();
 }
 
+async function getStandardLanguageClient(): Promise<LanguageClient | undefined> {
+	try {
+		if (standardServerStart) {
+			return await standardServerStart;
+		}
+	} catch (error) {
+		logger.error(`Failed to initialize the Java language client: ${getMessage(error)}`);
+	}
+	return standardClient.getClient();
+}
+
 async function postExtensionStartInit(
 	context: ExtensionContext,
 	requirements: requirements.RequirementsData,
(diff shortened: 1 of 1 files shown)


datho7561 (approved):
Looks good and fixes the problem. Thanks, Morgan!

PR #4463, opened 2026-07-14
Add settings for the TODO comment in generated method/catch stubs

### LINKED ISSUE #2963: Setting to modify/remove TODO Auto-generated method stub comments
When we click **Add unimplemented methods**, it includes a comment: `// TODO Auto-generated method stub`.

It's a small thing, but it's getting a little annoying to remove that comment in every instance, especially with large interfaces. ^-^'

There's already the `throw new UnsupportedOperationException` so I can decide for myself from there if it needs implemented or not, but including the comment just adds an unnecessary step when writing new implementations.

Could an option please be introduced to override/remove the comment?

Example:
```java
public class Test implements Map {

    @Override
    public int size() {
        // TODO Auto-generated method stub
        throw new UnsupportedOperationException("Unimplemented method 'size'");
    }

    @Override
    public boolean isEmpty() {
        // TODO Auto-generated method stub
        throw new UnsupportedOperationException("Unimplemented method 'isEmpty'");
    }

    // …
} 
```

##### Environment
- Operating System: Debian 11
- JDK version: 11
- Visual Studio Code version: 1.75.1

MeherSru:
Fixes #2963.

Adds java.templates.methodBody, java.templates.methodBodySuper, and java.templates.catchBody so people can customize or remove the
"// TODO Auto-generated ..." comment jdt.ls inserts into generated
method stubs and catch blocks. Server-side change is here:
https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3847

Defaults are set to match what's generated today (comment included), so this shouldn't change anything for anyone who doesn't touch the setting. Added the new settings to the README and the
predefined-variables doc too.

Tested end to end  and went through all three actions
with and without the settings set.


datho7561 (approved):
Looks good. Thanks, Meher!

datho7561:
I updated the wiki page to reflect the changes in the template variables markdown file

Issue #4469, opened 2026-07-22
Update to eslint 10


datho7561:
**Is your feature request related to a problem? Please describe.**
It'd be nice to update eslint to the latest version to get all the latest features and bug fixes and to allow us to use a newer version of typescript (the current setup only supports typescript up to 5.2.0).

This requires migrating the eslint config to the new flat config style: https://eslint.org/docs/latest/use/configure/migration-guide

The TypeScript plugin for eslint also has a different method of setting it up, requiring a different npm module to be installed (see https://typescript-eslint.io/getting-started/ for instructions)

**Describe the solution you'd like**
Migrate to eslint 10

**Describe alternatives you've considered**
Keep current version, but this limits what version of TypeScript we can use

**Additional context**
Upgrading TypeScript _might_ be required to upgrade to the latest version of the languageclient (I'm not too sure at this point, I need to dig further into it)

datho7561:
@MeherSru let me know if you're interested in working on this

MeherSru:
@datho7561 yes, I'd like to work on this!

datho7561:
Okay!

datho7561:
Fixed by #4470

PR #4470, opened 2026-07-24
Update ESLint to v10 

MeherSru:
This updates ESLint from v8 to v10, using the new flat config format, as requested in #4469. Also swaps to the new unified typescript-eslint package, and fixes a couple of other outdated tools (eslint-webpack-plugin, webpack-cli) that broke when trying to use the new ESLint.


eslint.config.mjs
```diff
@@ -0,0 +1,123 @@
+import { defineConfig, globalIgnores } from "eslint/config";
+import tseslint from "typescript-eslint";
+import globals from "globals";
+import stylistic from "@stylistic/eslint-plugin";
+
+export default defineConfig([globalIgnores([
+    "**/out",
```
datho7561: From what I understand `**/out` matches any folder under the current directory called `out`. To match the existing `.eslintignore` rule, I think you want just `out` (and same goes for the other entries in this array).

eslint.config.mjs
```diff
@@ -0,0 +1,123 @@
...
+    "**/.settings",
+    "**/.classpath",
+    "**/.project",
+    "test/resources/projects/**/.vscode",
+    "test/resources/projects/maven/salut/testGradle",
+    "**/test-temp",
+    "**/vscode*.d.ts",
+	"eslint.config.mjs",
+	"**/scripts/**",
+	"webpack.config.js",
```
datho7561: These files were linted before, and I think we should continue to lint these files. By adding them here, it prevents eslint from checking them.

I think the key is that you need to specify which files should be used in the next section, eg. add `files: ["src/**"],` before `languageOptions`.

...and then configure typescript parsing for the next two sections. eg. add:

```
    languageOptions: {
        globals: {
            ...globals.node,
        },

        parser: tseslint.parser,
        ecmaVersion: "latest",
        sourceType: "commonjs",
    }
```

after `files: ["**/*.js"],` and `files: ["**/*.test.ts"],`

datho7561 (changes requested):
Looks like a good start but needs some small changes to match what we had before.

It took me a while to review, because I forgot that vscode-java didn't adopt the default configuration options for eslint, and I thought you had missed that step in the migration. I think we should eventually adopt the default eslint options, but not in this PR. I'll open an separate issue for that.

datho7561 (approved):
Looks good. Thanks, Meher!

PR #4476, opened 2026-07-29
Update language client to 10.1.0

datho7561:
- Update to `vscode-langaugeclient@^10.1.0`
- Adjust to changes to language server logging API
  - VS Code now provides client-side log filtering based on a configured log level. This interacts in a slightly annoying way with our own server side log filtering, but ultimately both are helpful to have, since disabling logging on the client still means the server is sending trace level messages, which may be annoying.

~~Requires https://github.com/redhat-developer/vscode-java/pull/4487~~ merged

Required to get `workspace/textDocumentContent` working, see https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3855

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -284,13 +284,6 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 					});
 				}
 			},
-			resolveCompletionItem: async (item, token, next): Promise<CompletionItem> => {
-				const completionItem = await next(item, token);
-				if (completionItem?.documentation instanceof MarkdownString) {
-					completionItem.documentation = fixJdtLinksInDocumentation(completionItem.documentation);
-				}
-				return completionItem;
-			},
 			// https://github.com/redhat-developer/vscode-java/issues/2130
 			// include all diagnostics for the current line in the CodeActionContext params for the performance reason
 			provideCodeActions: async (document, range, context, token, next) => {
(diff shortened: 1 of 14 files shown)


src/extension.ts
```diff
@@ -286,9 +286,9 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 			},
 			resolveCompletionItem: async (item, token, next): Promise<CompletionItem> => {
 				const completionItem = await next(item, token);
-				if (completionItem?.documentation instanceof MarkdownString) {
-					completionItem.documentation = fixJdtLinksInDocumentation(completionItem.documentation);
-				}
+				// if (completionItem?.documentation instanceof MarkdownString) {
```
datho7561: Fix this

datho7561:
@MeherSru if you have time, do you mind testing this? Specifically:

1. Make sure that the regular vscode-java features work (errors, go to definition, hover, etc.)
2. Check specifically that the `jdt://` use case works. eg. links in hover Javadoc that point to classes in the JDK work properly (this is what you implemented for https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3855)

MeherSru:
regular features (errors, hover, go to definition) work fine, and jdt:// hover links work correctly too, clicking a JDK class link in hover Javadoc opens the decompiled source properly.

datho7561:
Thanks for reviewing, Meher!

PR #4477, opened 2026-07-29
Project-only search scope

datho7561:
See https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3850, this is the client side changes needed to adopt that PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1644,11 +1644,13 @@
             "type": "string",
             "enum": [
               "all",
-              "main"
+              "main",
+              "projectOnly"
             ],
             "enumDescriptions": [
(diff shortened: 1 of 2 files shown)


MeherSru:
Tested end to end with a small sample project that has a type used both in project code and a library jar.

- projectOnly shows up fine in settings and the search scope picker
- Picking it actually saves correctly
- Ran Find References before/after switching, went from 4 results to 3, library one dropped off like it should
- Diff is straightforward, just wires up the new option on the client side. Setting description is clear and matches actual behavior.

Looks good to me 👍

MeherSru (approved):

datho7561:
Thanks for the review, Meher!

PR #4482, opened 2026-08-07
Implement confirmation for non-fatal "Move Instance Method" problems before applying edits

### LINKED ISSUE #4460: Move Instance Method changes field null-check semantics and introduces NullPointerException
## Description

`Move Instance Method` changes the behavior of the program when moving the instance method `process()` from class `A` to class `C`.

Before refactoring, `process()` checks whether the field `c` is null. Since `a.c` is null in the original program, the method safely returns `0`.

After moving the instance method `process()` to class `C`, VS Code Java changes the call site from:

```java
a.process()
```

to:

```java
a.c.process()
```

However, `a.c` is null, so the refactored program throws `NullPointerException` before the moved method is entered.

In addition, the original condition:

```java
if (c != null) return 2;
```

is rewritten as:

```java
if (this != null) return 2;
```

This is not semantically equivalent. The original condition checks whether the source object's field `c` is null, while `this != null` is effectively always true once the instance method is entered.

## Code before refactoring

```java
public class t {
        A a = new A();
        System.out.println(a.process());
    }
}

class A {
    C c;

    // Move Instance Method target method
    int process() {
        if (c != null) return 2;
        return 0;
    }
}

class B extends A { }

class C extends B { }
```

Before refactoring, the program prints:

```text
0
```

## Refactoring operation

Apply `Move` to method `A.process()` and move it to class `C`.

## Code after refactoring

```java
public class t {
        A a = new A();
        System.out.println(a.c.process());
    }
}

class A {
    C c;
}

class B extends A { }

class C extends B {
    int process() {
        if (this != null) return 2;
        return 0;
    }
}
```

After refactoring, running the program throws:

```text
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "C.process()" because "<local1>.c" is null
    at t.main(t.java:4)
```

shin19991207:
Fixes https://github.com/redhat-developer/vscode-java/issues/4460

Depends on and was tested with https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3862

This PR adds a confirmation dialog for non-fatal "Move Instance Method" refactoring problems reported by JDT LS.

In Eclipse IDE, `ERROR`-level refactoring problems are displayed in the **Found problems** dialog. The user can review the problems and choose whether to continue or cancel the refactoring.

Eclipse IDE Found problems dialog

This PR provides the same choice in VS Code:

VS Code refactoring confirmation dialog

When JDT LS reports non-fatal problems, VS Code displays all found problems and asks the user whether to continue. No edit is applied if the dialog is cancelled. If the user selects **Continue**, the move request is repeated with the confirmation token returned by JDT LS.

The confirmation token prevents the user from confirming a refactoring after its source or conditions have changed. To test this behavior: Select **Refactor...** > **Move...** > Wait for the problem dialog to appear > Modify and save the source file in another editor > Select **Continue** in the dialog > The stale confirmation is rejected:

VS Code stale refactoring confirmation error

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -255,6 +255,7 @@ export async function activate(context: ExtensionContext): Promise<ExtensionAPI>
 				advancedExtractRefactoringSupport: true,
 				inferSelectionSupport: ["extractMethod", "extractVariable", "extractField"],
 				moveRefactoringSupport: true,
+				moveRefactoringConfirmationSupport: true,
 				clientHoverProvider: true,
 				clientDocumentSymbolProvider: true,
 				gradleChecksumWrapperPromptSupport: true,
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -350,9 +350,11 @@ export interface RenamePosition {
 }
 
 export interface RefactorWorkspaceEdit {
-    edit: WorkspaceEdit;
+    edit?: WorkspaceEdit;
     command?: Command;
     errorMessage?: string;
+    canContinue?: boolean;
+    confirmationToken?: string;
 }
 
 export interface GetRefactorEditParams {
@@ -411,6 +413,7 @@ export interface MoveParams {
     params: CodeActionParams;
     destination?: any;
     updateReferences?: boolean;
+    confirmationToken?: string;
 }
 
 export interface MoveDestinationsResponse {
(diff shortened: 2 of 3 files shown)


datho7561 (approved):
Works well and seems very helpful. Thanks, Morgan!

Issue #3870, opened 2024-11-22
enhancement - option to set class/enum/interface/record templates


mamilic:
Whenever new class is created it is populated with template, I was wondering if there is already some existing setting to configure this.
Also it would be really handy to be able to set template for such use cases.

datho7561:
Closed by https://github.com/redhat-developer/vscode-java/pull/4484

PR #4484, opened 2026-08-13
Add setting to make new classes package-private instead of public

MeherSru:
This adds a new setting: java.templates.preferPackagePrivateVisibility
When turned on, new Java files (class/interface/enum/record) won't
be created with "public" anymore. They'll be package-private instead.


datho7561:
~~This PR seems to work well without the server side changes in jdtls. Are the server side changes needed? If so, what do they do?~~ oops figured it out

datho7561:
Fixes https://github.com/redhat-developer/vscode-java/issues/3870

datho7561 (approved):
Looks good and works well. Thanks, Meher!

PR #4493, opened 2026-08-31
Add a setting to specify the Maven project cache size

datho7561:
This new setting, `java.jdt.ls.mavenProjectCacheSize`, configures a command line option for jdtls that modifies m2e's project cache size. Increasing the project cache size can significantly decrease the loading time of Maven projects that have many subprojects, such as monorepos. However, doing this also uses more memory and risks causing an out of memory exception. Setting the project cache to a larger number can be used as a stop-gap solution for loading larger Maven projects while we investigate better methods of properly addressing the fact that Maven projects with many sub projects load very slowly.


shin19991207 (approved):
LGTM. Thanks, David!

lch1986216:
能不能解决大型项目，cpu占用高，自动编译卡主，跳转卡死的问题，像idea，eclipse一样流畅

PR #4508, opened 2026-09-20
Fix class file editors not restored after restart

### LINKED ISSUE #4406: Opened class files in referenced libraries are not restored on restart
**Description**
When `workbench.editor.restoreEditors` is enabled, opened project files are restored across VSCode restarts, but this does not work with opened class files from a referenced library.

This is fairly annoying when a project significantly involves some source-unavailable library.

**To Reproduce**
1. Ensure `workbench.editor.restoreEditors` is enabled
2. Create a Java project (e.g. using Gradle)
3. Add a JAR dependency
4. Open a class file from the dependency in the "Java Projects" view
5. Restart VSCode - The class file editor is not present

Having looked at trace-level logs for both this extension and the VSCode extension host, I do not believe they contain any useful information here (no clear consistent error message that seems related to this, and the class file itself is only mentioned once by this extension in the context of a large list of other classfiles that were not open prior to the restart). This issue has also persisted across multiple extension versions and devices, so I highly doubt it is a environment-related issue.

**Expected behavior**
The opened classfiles should remain open through restarts when `workbench.editor.restoreEditors` is enabled.

**Environment**
- Operating System: Windows 11
- JDK version: Oracle Java 24.0.2
- Visual Studio Code version: 1.117.0
- Java extension version: 1.54.0

sripal2013:
## Problem
With `workbench.editor.restoreEditors` enabled, class files opened from referenced libraries (JAR dependencies) are not restored after a VS Code restart, while regular files restore fine. (Issue #4406)

## Root cause
Class files from JARs open as `class://` virtual documents, resolved by the `class` scheme `TextDocumentContentProvider`. That provider was registered in `postExtensionStartInit()`, which `activate()` kicks off fire-and-forget after slow awaits (requirements resolution, modal import dialogs). VS Code restores editors early during window startup, so the restore pass ran before any `class` provider existed and those editors were silently dropped.

## Fix
Register the `class` content provider synchronously as the first step of `activate()`, before any asynchronous work. I extracted the registration into `registerClassContentProvider()` in `providerDispatcher.ts`; the remaining client providers stay lazy since they aren't on the restore path.

## Testing
- `tsc` compiles cleanly, ESLint clean on both changed files.
- I couldn't run the full headed test suite in this environment, so it's worth a manual check: open a class file from a JAR in the Java Projects view, restart VS Code, and confirm the tab comes back.

Fixes #4406


datho7561:
I couldn't get this to work. I opened String.java through Ctrl+Click, the closed and reopened the workspace in VS Code. A regular file is restored but the virtual document isn't.

datho7561:
Oh wait maybe that's because String is in the JDK, I'll try out something from a library

datho7561:
Nope, same thing, the virtual document for a library class doesn't seem to get restored.

sripal2013:
thanks for testing it. 

You're right, registering the provider early wasn't enough on its own. When VS Code restores the editor, activate() is still waiting on requirements resolution, so getActiveLanguageClient() throws (apiManager isn't initialized yet) and the restore gets dropped.

I just pushed an update, the content provider now waits for the language client to be available instead of failing, so the restored editor resolves once the server is up.

datho7561:
Did you actually test it this time to see if it works?

I tried it out and it still doesn't work.

Is it even possible to persist virtual documents? It looks like they're already gone by the time I reopen VS Code.

